PR #12, opened 2020-07-01
Improve errors

sebbacon:
Provide specific errors for common error cases.

Where specific exceptions are not used, we return a deliberately
uninformative message, to avoid leaking potentially sensitive

Also, add a specific id to every log message associated
with a job.

### DIFF
--- a/runner/__init__.py
+++ b/runner/__init__.py
@@ -185,7 +185,7 @@ def fetch_study_source(workdir, job):
 def report_result(future):
     job = future.job
     joblogger = get_job_logger(job)
-    id_message = f"Error id {job_id_from_job(job)}"
+    id_message = f"id {job_id_from_job(job)}"
     try:
         job = future.result(
             timeout=COHORT_EXTRACTOR_TIMEOUT
--- a/runner/__init__.py
+++ b/runner/__init__.py
@@ -214,7 +214,7 @@ def report_result(future):
             job["url"],
             json={
                 "status_code": error.status_code,
-                "status_message": f"{repr(error)} {id_message}",
+                "status_message": f"{error.safe_details()} {id_message}",
             },
             auth=get_auth(),
         )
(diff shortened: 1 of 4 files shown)


tests/test_runner.py
```diff
@@ -3,6 +3,10 @@
 from unittest.mock import patch
 import time
 
+from runner import make_volume_name
+from runner import run_cohort_extractor
+from runner.exceptions import BadDockerImageName
```
inglesp, line 8: `make_volume_name` and `BadDockerImage` are imported but not used.

runner/exceptions.py
```diff
@@ -0,0 +1,30 @@
+class OpenSafelyError(Exception):
```
inglesp, line 1: Could you add a note to this module to say that -1 and 99 are reserved?

runner/__init__.py
```diff
@@ -47,107 +75,164 @@ def validate_input_files(workdir):
...
+            f"--output-dir={output_path}",
+        ]
+
+        os.chdir(workdir)
+        joblogger.info("Running subdocker cmd `%s` in %s", cmd, workdir)
+        result = subprocess.run(cmd, capture_output=True, encoding="utf8")
+        if result.returncode == 0:
+            joblogger.info("cohort-extractor subdocker stdout: %s", result.stdout)
+        else:
+            raise CohortExtractorError(result.stderr)
```
inglesp, line 146: Since we include `result.stderr` in the JSON that we `PUT` later on, we need to be really careful here.  Is there a risk that `result.stderr` could contain something sensitive?

tests/test_runner.py
sebbacon: Bad commit-splitting on my part; they are used in the next commit

runner/__init__.py
sebbacon: Yes!

evansd: We could make `OpenSafelyError` accept two arguments, something like `message` and `sensitive_info` and then have the both appear in the logs, but only the former in the reported error.

sebbacon:
Yes, I thought about that suggestion (it came through after I'd written the approach here).

I was worried about repetition of the messages, but realised a flag could fix that (see latest commit)

PR #19, opened 2020-07-31
Project yaml support

sebbacon:
This PR supports running jobs securely on a level 2 server, via a job queue. See the README.md for an overview.


README.md
```diff
@@ -1,26 +1,177 @@
...
+    "workspace_id": 1,
+}
+```
+
+In practice, jobs are (will be) added to the queue via a web interface on the
+job server.
+
+## Consuming jobs
+
+A job runner is service installed on a machine that has access to a given
```
ghickman, line 41: a service*

README.md
```diff
@@ -1,26 +1,177 @@
...
+  accessed using the `PRIVATE_REPO_ACCESS_TOKEN` supplied in the environment.
+* Its `project.yaml` is parsed
+  * Individual `actions` are extracted
+  * Variables in each action are interpolated, and the `run` action is converted
+    to a `docker run` invocation
+  * A dependency graph is calculated for the requested action; for example, an
+    action might depend on three previous actions before it can be run
+  * Each action in the graph is checked to see if it needs to be run
+    * Actions that either: (a) already have output generated from a previous
+      run; (b) are currently running; (c) failed on their last run to do not
```
ghickman, line 62: s/on their last run to do not/on their last run do not/

inglesp:
This is a good piece of work with lots of moving parts.  I haven't done a nitty tactical review, but here's some feedback from reading through what look like the main parts of the code.

#### What gets run when?

I don't fully understand the flow control around deciding what gets run and when, and I'm not confident that it's free of race conditions.

There are a few reasons for this:

1. There's a pool of processes that are each deciding what they should be doing, and this just makes me nervous
1. It's not explicitly tested
1. Using the `DependencyRunning` exception is odd (but probably ok)
1. The recursive nature of `run_job_and_dependencies` is a little hard to follow (but probably ok)

I think it would worth considering separating out:

  1. working out what needs to run (a dispatcher?)
  1. running what needs to be run

The dispatcher should then run in a single process, avoiding many of the possible race condition situations.  It could have a function called `decide_what_needs_to_run` which is a pure function taking some state, and returning a list of jobs that can be run.  The dispatcher could then update the global state, and kick off a worker.

#### Passing around mutable dicts

We're passing around lots of dicts, and it's not always clear what they contain -- eg what's in something called `job_config`?  It might be worth creating classes for this kind of thing, and maybe using `attrs` for this.

The dicts are also mutated in a number of places, which makes me nervous.  As far as I can tell, this is mostly for setting sensible defaults.  This could be avoided if the dicts were objects.

####  Naming consistency

We're using more than one name for the same thing.  Eg is it an `action` or an `operation`?  `kwargs` or a `job_config`?  Should we define a vocabulary document it somewhere?

README.md
```diff
@@ -1,26 +1,177 @@
...
+data to prepare it for release (for which it is planned to add a
+`minimally_sensitive` category).
+
+Outputs are therefore persisted to filesystem paths according to the following
+environment variables:
+
+```sh
+# A location where cohort CSVs (one row per patient) should be
+# stored. This folder must exist.
+OPENSAFELY_HIGH_PRIVACY_STORAGE_BASE=/home/opensafely/high_security
```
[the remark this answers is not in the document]
sebbacon: I didn't understand your point here!

PR #38, opened 2020-10-07
Misc fixes

sebbacon:
A raggle-taggle assortment of fixes I've had to make to fulfil the obligation I've set myself to get snotty noses run by the end of the week.


evansd (approved):
I've made an inline note of the one obvious thing we should revisit.

Going to approve and merge as discussed as these changes are already live.

PR #42, opened 2020-10-09
Refactor Loading project.yaml

ghickman:
This does a pure refactor to split apart loading a `project.yaml` and validating it.

As part of opensafely/job-server#61 we want to get the actions defined in a `project.yaml`.  Downloading a file from GitHub via their API gives us the file contents  so rather than saving to disk just to read it into `load_and_validate_project` it makes sense to break up those two bits of functionality.

I'm expecting to add validation to the web flow later, to add an extra layer of checks on invalid `project.yaml` reaching job-runner.

### DIFF
--- a/jobrunner/project.py
+++ b/jobrunner/project.py
@@ -61,7 +61,7 @@ def variables_in_string(string_with_variables, variable_name_only=False):
         return [x[0] for x in matches]
 
 
-def load_and_validate_project(workdir):
+def validate_project(workdir):
     """Check that a dictionary of project actions is valid"""
     with open(os.path.join(workdir, "project.yaml"), "r") as f:
         project = yaml.safe_load(f)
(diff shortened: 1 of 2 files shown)


evansd (approved):
:+1: Nice

sebbacon (approved):

sebbacon:
N.B. this project uses semantic-release to handle release bumping: https://github.com/mathieudutour/github-tag-action/#bumping

So ideally at least one of the commits should include [an Angular-convention header](https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#-git-commit-guidelines), e.g. `refactor:` in this case.

That said it's not very important in this case as a refactor would not result in a version bump!

ghickman:
@sebbacon – `refactor` added!

PR #44, opened 2020-10-14
fix: use status_code to determine something is running

sebbacon:
Previously we relied on `started` having been set to `False` as one of
the conditions for remaining in a `pending` meta-state.

However, this flag is set and unset by the main loop and may be in an
inconsistent state. The `status_code`, on the other hand, will remain
that of `DependencyRunning` until the related job completes, leading to
a lower likelihood of inconsistent state.


evansd (approved):
This makes sense

PR #46, opened 2020-10-16
Avoid including token in repo URL

evansd:
Previously we tried to redact the token from anywhere it might leak to,
but this is fiddly and error prone. Git doesn't directly support
supplying the token in an environment variable, but it does allow us to
specify a custom "askpass" executable which can grab the token from the
environment for us.


evansd:
The two test failures seem to be common to all PRs at the moment, so I don't think my changes have broken anything new.

sebbacon (approved):

PR #48, opened 2020-10-16
Set temporary database environment variable

evansd:
This allows us to make use of the mechanism whereby results are written
to a temporary table before downloading so that if there's a problem
with the download it can be retried without re-running all the queries.

At the same time we change the docker invocation to pass the database
credentials as an environment variable

Passing it as an argument means it shows up to anyone running `docker ps`.
There's no real security difference here as the environment
variables are all visible using `docker inspect` if you try hard enough;
but it's at least a bit less in your face this way.


evansd:
NB. I'm _pretty sure_ this is all correct but it's quite hard for me to test at the moment.

Also, I've added the new environment variable to the `.env` file in production so it should all be ready to go. I didn't have permission to edit this, so I made a copy, edited that, and then moved it into place using Docker. Hopefully that doesn't cause any issues.

sebbacon (approved):

PR #53, opened 2020-10-21
fix: always report the detail of validation errors

sebbacon:


evansd (approved):
👍 this is the approach I took in the refactor I've been working on

Issue #41, opened 2020-10-08
I/O issues when copying/using large input files


sebbacon:
When running several actions in parallel on the production (Windows) server, things grind to a halt.

Just running `ls` inside a container is extremely slow.

This was observed when several docker containers had been started at once.  The hypothesis was that the step where files are copied from the relevant storage into the container was hitting some kind of IO limit. 

As an attempted fix, https://github.com/opensafely/job-runner/pull/38/commits/10bd9ffcd6794c0d37f8d8418d59e34b46c04bcd caused individual input files to be mounted read-only, rather than copied.

However, similar behaviour was observed following the change.

As ever, it's hard to understand the implementation of volume / bind mounting in Docker, but my sense is that a bind mount still results in a copy operation before the `RUN` command is executed, thus making the change above a bit pointless.

The documentation is unclear, but I believe we currently use bind mounts rather volumes; that is, I think `--volume /path:/dst` is a bind mount from the `/path` location on the host filesystem, but `--volume path:/dst` is a volume mount from the `path` volume on the host filesystem.

The Docker documentation says that "a volume’s contents exist outside the lifecycle of a given container", which implies that perhaps creating a volume and mounting that will involve less IO.

The process would be 

1. Before running a docker container, at the point input files are copied in, use some kind of hashed key to see if a volume has already been created for those input files
2. If not, create a volume and copy the input files in (as far as I can tell, this must be done by creating a docker container with a new volume mounted)
3. Then mount the volume

Before implementing this, it's worth creating a test case to check assumptions about IO.  This would involve:

```
fallocate -l 3G very_large_file.txt

for i in {1..5}; do
  docker run -d --rm --entrypoint bash --volume $(pwd)/very_large_file.txt:/tmp/very_large_file.txt:ro docker.opensafely.org/job-runner -c sleep 600s
done;

docker run --rm --entrypoint bash  -it --volume $(pwd)/very_large_file.txt:/tmp/very_large_file.txt:ro docker.opensafely.org/job-runner -c "ls -l /tmp/very_large_file.txt"


```

evansd:
Beware that if you're doing this in git-bash you need an extra leading slash on the host path (so `/$(pwd)` rather than `$(pwd)`) for it to work as expected.

sebbacon:
So the above test case does not create lots of IO. Next step: same thing but reading the mounted files in each docker

```

for i in {1..5}; do
[...]

PR #56, opened 2020-10-23
Volumes not binds

sebbacon:
Per #41, using bind mounts is extraordinarily slow on Windows.  Using Docker volumes is much faster on Windows and the same speed on Linux.

At the moment, we throw away volumes after we've used them. There's a possibility in the future of keeping volumes around based on cache keys to prevent re-making them for actions that consuming identical inputs.

The reason for implementing this right now is not for performance reasons, but so we can run synchronous (cohort-extractor-mediated) action pipelines within Github Actions; bind mounts break because GHA runs docker-in-docker with a shared pipe to a common docker daemon.


sebbacon:
> so we can run synchronous (cohort-extractor-mediated) action pipelines within Github Actions

...and predictably, I see the tests are failing in GHA where they don't locally... :crying_cat_face:

ghickman (approved):

evansd (approved):
I didn't initially understand why we needed the extra container along with the volume, but I can see it's because `docker cp` doesn't support creating intermediate directories.

It might be possible to use this [temp dir trick](https://github.com/moby/moby/issues/20920#issuecomment-345873422) to work around this, but that's not obviously simpler than creating the extra container.

PR #57, opened 2020-10-27
Use docker's built in PID 1 init process

bloodearnest:
Also enabled service restarts, which means the job-server become more of a
proper service in production.

PID 1 is special, and python doesn't always behave correctly as PID 1 wrt its
children. Docker includes tini, a PID 1 runner that behaves as expected and
runs your process and passes on signals and does process reaping properly.

You can use via `docker run --init` or via `init: true` in docker-compose.yaml.

Good explanation here: https://github.com/krallin/tini/issues/8

So, this enables init for job-runner, and also for jobs themselves. It may not
be related to the lock ups we've seen, but it very well could be, and it's
recommended best practise for docker daemons anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ jobs.
 
 Jobs belong to a `workspace`. This describes the git repo containing the
 OpenSAFELY-compliant project under execution; the git branch, and kind of
-database to use. The workspace also accts as a kind of namespace for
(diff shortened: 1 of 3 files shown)


evansd (approved):
Good catch

sebbacon (approved):

PR #62, opened 2020-11-03
fix: use posixpath where we plan to run in linux container

### LINKED ISSUE #60: Use PosixPath rather than Path for correct behaviour on Windows
For the [`os-demo-research/tree/cohort-study`](https://github.com/opensafely/os-demo-research/tree/45490f0ec96adfe97be8a296b4d2e1915d13ffe6) repo, I'm trying to test the [project.yaml](https://github.com/opensafely/os-demo-research/blob/45490f0ec96adfe97be8a296b4d2e1915d13ffe6/project.yaml) locally, as per instructions [here](https://docs.opensafely.org/en/latest/project_pipelines/)

When I try to run the first action using `cohortextractor run dummy generate_cohort_deaths expectations`, it gives the following error:

```
Invalid requested path \workspace\output\input_deaths.csv, not in /workspace
```

I suspect this is just a silly mistake somewhere but I can't get past it. Perhaps I'm specifying relative file paths wrong in the yaml but I've tried a few things with no success.

sebbacon:
Fixes #60

### DIFF
--- a/jobrunner/job.py
+++ b/jobrunner/job.py
@@ -7,7 +7,7 @@ import subprocess
 import tempfile
 import time
 from contextlib import contextmanager
-from pathlib import Path
+from pathlib import PosixPath
 
 from jobrunner.exceptions import (
     DependencyRunning,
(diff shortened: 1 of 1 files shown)


ghickman (approved):
`asd`?

PR #67, opened 2020-11-04
ci: fix volume cleanup for slow docker

bloodearnest:
When docker ref counts volumes asynchronously, so can some times take a short
while after a container is stopped before the volume can be deleted.

Also, renamed test exceptions in not be parsed as test cases, and fixed a dummy
filename to avoid potential clashes.

### DIFF
--- a/jobrunner/job.py
+++ b/jobrunner/job.py
@@ -125,8 +125,6 @@ def volume_from_filespec(input_file_spec):
     finally:
         cmd = ["docker", "stop", volume_container_name]
         subprocess.check_call(cmd)
-        cmd = ["docker", "wait", volume_container_name]
-        subprocess.check_call(cmd)
         # docker volumes are reference counted, apparently
         # non-deterministically, so wait to ensure that the docker daemon has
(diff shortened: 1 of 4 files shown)


jobrunner/job.py
```diff
@@ -125,6 +125,10 @@ def volume_from_filespec(input_file_spec):
     finally:
         cmd = ["docker", "stop", volume_container_name]
         subprocess.check_call(cmd)
+        # docker volumes are reference counted, apparently
+        # non-deterministically, so wait to ensure that the docker daemon has
+        # decremented the volume count, or we can not remove it.
+        time.sleep(0.1)
         cmd = ["docker", "volume", "rm", volume_name]
```
evansd, line 132: Apologies for zero-content comment here, but can you not just use `rm --force`?

evansd (approved):

jobrunner/job.py
bloodearnest: I tried, it doesn't work when the container is still in use!

evansd: :disappointed:

PR #71, opened 2020-11-10
feat: Version 2

evansd:
This is already running in production. 

It largely implements the protocol discussed in https://github.com/opensafely/job-server/issues/84

BREAKING CHANGE: Many things are different so this should be a semver major bump


jobrunner/sync.py
```diff
@@ -0,0 +1,116 @@
...
+def main():
+    while True:
+        sync()
+        time.sleep(config.POLL_INTERVAL)
+
+
+def sync():
+    results = api_get(
+        "job-requests", params={"active": "true", "backend": config.BACKEND}
+    )
```
sebbacon, line 29: Should there be some explicit query param setting for number of jobs to request in one batch? If the answer is "as many as possible, leave it to the server" then is that worth mentioning here?

jobrunner/sync.py
```diff
@@ -0,0 +1,116 @@
...
+    response.raise_for_status()
+    return response.json()
+
+
+def job_request_from_remote_format(job_request):
+    """
+    Convert a JobRequest as received from the job-server into our own internal
+    representation
+    """
+    return JobRequest(
```
sebbacon, line 63: There have been conversations about enforcing the required contract via an Interface Definition Language. This is possibly worth considering as something that could then be tested in both the server and the runner. 

There is similarly an argument (actually, stronger, I think) for using a type annotation for JobRequest. Though if you're going to bother I think it would then make sense to use annotations everywhere. It would give extra reassurance about the code.  But again, it's moot. Just wanted to throw it out there...

sebbacon (commented):
Nice and readable - will do full review later

jobrunner/create_or_update_jobs.py
```diff
@@ -0,0 +1,247 @@
...
+
+
+def create_jobs(job_request):
+    validate_job_request(job_request)
+    # In future I expect the job-server to only ever supply commits and so this
+    # branch resolution will be redundant
+    if not job_request.commit:
+        job_request.commit = get_sha_from_remote_ref(
+            job_request.repo_url, job_request.branch
+        )
```
ghickman, line 74: This should be redundant now that job-server provides a commit with each JobRequest??

ghickman (approved):

PR #73, opened 2020-11-18
fix: consistently join posix/windows mixed paths

sebbacon:


jobrunner/utils.py
```diff
@@ -50,7 +50,7 @@ def get_workdir():
 def safe_join(startdir, path):
     """Given a `startdir` and `path`, join them together, while protecting against directory traversal attacks that take us outside `startdir`
     """
-    requested_path = posixpath.normpath(posixpath.join(startdir, path))
+    requested_path = posixpath.normpath(posixpath.join(startdir, *path.split("\\")))
```
evansd, line 53: Slightly neater would be to split on `os.sep` here, but what you have obviously works on POSIX too so fine to leave as is.

evansd (approved):

ghickman (approved):
Did this fix a particular bug or is it a consistency fix?  Just curious!

jobrunner/utils.py
sebbacon: I opted for the non-`os.sep` route because I thought it made the reason we're doing it clearer. But on reflection I was wrong!

sebbacon:
@ghickman slightly magically hoping it'll fix this (without having a Windows machine at my desk to test it at the moment)

image

PR #85, opened 2020-11-24
Translate Cross System ID Field Name

ghickman:
This translates job-server's internal `Job.id` field _name_ to `runner_id` when talking to the API.  job-server maintains an internal primary key for Jobs, tracking the cross-system identifier in the runner_id field.

@evansd – we've used `identifier` for `JobRequest`'s cross-system ID, should we standard to that for both?  (happy to hold off on this PR to do that!)

### DIFF
--- a/jobrunner/sync.py
+++ b/jobrunner/sync.py
@@ -87,12 +87,11 @@ def job_to_remote_format(job):
     Convert our internal representation of a Job into whatever format the
     job-server expects
     """
-    return {
+    data = {
         key: value
         for (key, value) in job.asdict().items()
         if key
         in [
-            "id",
             "job_request_id",
             "action",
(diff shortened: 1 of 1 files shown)


jobrunner/sync.py
```diff
@@ -104,6 +104,11 @@ def job_to_remote_format(job):
         ]
     }
 
+    # convert cross-system identifier key
+    data["runner_id"] = job.id
```
evansd, line 108: I think it would be worth removing `id` from the list of keys above to avoid any confusion.

evansd (approved):
I think I prefer `identifier` but not too fussed

ghickman:
Switched `identifier` for `Job`s on both sides.

PR #86, opened 2020-11-24
Remove Active Filter

ghickman:
This removes the `?active=true` filter since job-server only returns active `JobRequest`s.

I've also configured the API URL as part of the integration test so it's stable during test and local configuration doesn't affect it.

### DIFF
--- a/jobrunner/sync.py
+++ b/jobrunner/sync.py
@@ -30,7 +30,7 @@ def sync():
         # We're deliberately not paginating here on the assumption that the set
         # of active jobs is always going to be small enough that we can fetch
         # them in a single request and we don't need the extra complexity
-        params={"active": "true", "backend": config.BACKEND},
+        params={"backend": config.BACKEND},
     )["results"]
     job_requests = [job_request_from_remote_format(i) for i in results]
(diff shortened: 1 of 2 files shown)


evansd:
This is very bikesheddy of me, but I have a mild preference for either keeping the `?active=true` parameter or changing the endpoint to something like `job-requests/active/`

Basically, I want something that reveals the intention that these are only active JobRequests were expecting. If we do keep the `?active=true` param, could we stick an assert or something in the view that just makes sure that you can never call it without setting that param?

Happy to stick this on the "tidy up later" pile in the interests of moving things forward.

evansd (approved):

ghickman:
> keeping the ?active=true parameter

Let's say we keep the parameter… what happens when you remove it?  We currently only serve active JobRequests because that's what the job-runner thinks in terms of.  While we very much could expand the API in the future, I think this is probably YAGNI/KISS for now?

evansd:
> Let's say we keep the parameter… what happens when you remove it?

You get a 500 as per

> could we stick an assert or something in the view that just makes sure that you can never call it without setting that param

But yeah, happy to leave for now.

PR #91, opened 2020-11-30
Refer to default branch with HEAD

### LINKED ISSUE #80: [v2] Local Testing
My branch, `job-server-compat`, has a failing test locally which is passing in CI.  `test_integration` gets both the wrong number of jobs and has the wrong statuses.

Here's the output from pytest:
```
        # Run sync to grab the JobRequest from the mocked job-server
        jobrunner.sync.sync()
        # Check that three pending jobs are created
        jobs = get_posted_jobs(requests_mock)
>       assert [job["status"] for job in jobs.values()] == [
            "pending",
            "pending",
            "pending",
            "pending",
        ]
E       AssertionError: assert ['failed'] == ['pending', '...g', 'pending']
E         At index 0 diff: 'failed' != 'pending'
E         Right contains 3 more items, first extra item: 'pending'
E         Full diff:
E         - ['pending', 'pending', 'pending', 'pending']
E         + ['failed']
```

This is mostly for me to dig back into later but if this is an obvious failure then please do point me in the right direction @evansd!

ghickman:
The default repo for a branch can be configured globally via
init.defaultBranch.  Rather than relying on a user's default branch
matching our expectations, this uses HEAD to pick whatever is in use
locally.

Fixes #80

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -37,7 +37,7 @@ def test_integration(tmp_work_dir, docker_cleanup, requests_mock, monkeypatch):
                     "workspace": {
                         "name": "testing",
                         "repo": str(repo_path),
-                        "branch": "master",
+                        "branch": "HEAD",
                         "db": "dummy",
                     },
                 }


evansd:
Tests are failing because we're in the middle of moving docker registries, but I've run locally and all good.

evansd (approved):

PR #99, opened 2020-12-03
fix: Temporary workaround for Docker auth issues in CI

evansd:
This is done so that existing Github Actions tests run as-is, even
though they're authenticating with the old Docker registry.

Should be removed as soon as we update the actions.


bloodearnest (approved):
LGTM

PR #100, opened 2020-12-03
Fix our sdist packaging

evansd:
We were missing the VERSION file so our wheel installed OK but the sdist didn't. I realised this when attempting to vendor the PyPI released package (which uses the sdist).

This also adds tests that the packages we build can be installed.


CarolineMorton (approved):
Looks fine to me

PR #106, opened 2020-12-07
feat: support external stata licence configuration

bloodearnest:
A stata licence can be configured by setting a STATA_LICENCE environment
variable, which will be needed for GHA and backend server usage.

For local runs, we add a best-effort convenience method - checking out a
private repo containing the opensafely license using the current user's git
credentials. If the user has access to the repo (either an opensafely member or
external collaborator), everything should just work.  If not, they'll get an
error message telling them to set the environment variable for their stata licence.

The goal of this is so we can make the stata docker image public, and avoid the
need for any alternative credentials for users - we just need GitHub.

Note: the stata-mp image does not yet require a license - that change will land
once this one has.

### DIFF
--- a/jobrunner/local_run.py
+++ b/jobrunner/local_run.py
@@ -179,17 +179,12 @@ def create_and_run_jobs(
         if is_stata and config.STATA_LICENSE is None:
             config.STATA_LICENSE = get_stata_license()
             if config.STATA_LICENSE is None:
-                print('You need to configure a stata licence to use stata '
-                      'actions in your project.\n'
-                      'Set the STATA_LICENCE environment variable to the '
-                      'contents of your stata.lic file.'
-                )
-                return False
+                # TODO switch this to failing when the stata image requires it
+                print('WARNING: no STATA_LICENSE found')
                 
         if not docker.image_exists_locally(image):
             print(f"Fetching missing docker image {image}")
             print(f"\nRunning: docker pull {image}")
-            import pdb; pdb.set_trace()
             try:
                 # We want to be chatty when running in the console so users can
                 # see progress and quiet in CI so we don't spam the logs with
--- a/jobrunner/local_run.py
+++ b/jobrunner/local_run.py
@@ -22,12 +22,14 @@ import argparse
 import json
 import os
 from pathlib import Path
+import platform
 import random
 import shlex
 import shutil
 import string
 import subprocess
 import sys
+import tempfile
 import textwrap
 
 from .run import main as run_main, JobError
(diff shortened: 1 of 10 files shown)


bloodearnest:
This is a WIP, not ready to merge yet.

TODO: possibly support both using git+ssh:// and https:// protocols to access the repo, as not sure what will work in Docker for Windows

TODO: figure out why the minimal stata project won't run.

jobrunner/local_run.py
```diff
@@ -171,30 +173,32 @@ def create_and_run_jobs(
...
+        is_stata = 'stata-mp' in image
+
+        if is_stata and config.STATA_LICENSE is None:
+            config.STATA_LICENSE = get_stata_license()
+            if config.STATA_LICENSE is None:
+                # TODO switch this to failing when the stata image requires it
+                print('WARNING: no STATA_LICENSE found')
+                
+        if not docker.image_exists_locally(image):
+            print(f"Fetching missing docker image {image}")
```
evansd, line 186: The most trivial of bikesheddy comments, but is this extra print necessary given the one below? Could maybe combine them as:
```py
print(f"\nFetching missing docker image: docker pull {image}")
```

jobrunner/local_run.py
```diff
@@ -335,30 +337,65 @@ def temporary_docker_auth_workaround(missing_docker_images):
...
+        return False
+
+    return True
+
+
+def get_stata_license(repo=config.STATA_LICENSE_REPO):
+    """Load a stata license from local cache or remote repo."""
+    cached = Path(f'{tempfile.gettempdir()}/opensafely-stata.lic')
+    repos = [repo, repo.replace('https://', 'git+ssh://git@')]
+    env = {'GIT_TERMINAL_PROMPT': '0'}
```
evansd, line 372: These look like leftover code from a previous iteration

jobrunner/local_run.py
```diff
@@ -335,30 +337,65 @@ def temporary_docker_auth_workaround(missing_docker_images):
...
+
+def get_stata_license(repo=config.STATA_LICENSE_REPO):
+    """Load a stata license from local cache or remote repo."""
+    cached = Path(f'{tempfile.gettempdir()}/opensafely-stata.lic')
+    repos = [repo, repo.replace('https://', 'git+ssh://git@')]
+    env = {'GIT_TERMINAL_PROMPT': '0'} 
+
+    def git_clone(repo_url, cwd):
+        cmd = ['git', 'clone', '--depth=1', repo_url, 'repo']
+        result = subprocess_run(cmd, cwd=cwd, env={'GIT_TERMINAL_PROMPT': '0'})
```
evansd, line 376: TIL `GIT_TERMINAL_PROMPT`, nice! Worth a comment here though?

evansd, line 376: Do we want to capture output here? Otherwise don't we end up with git output in the terminal?

tests/conftest.py
```diff
@@ -2,6 +2,8 @@
 
 import pytest
 
+from jobrunner.database import get_connection_from_file
```
evansd, line 5: Is the redundant?

evansd (approved):
Looks great. A little bit of trivial nitpicking from me, but nothing blocking.

jobrunner/local_run.py
bloodearnest: absolutely, will add

PR #121, opened 2020-12-15
Make connection caching threadsafe

evansd:
In the near future we plan to run the `sync` service in a thread
alongside the `run` service (for operational simplicity) and so we want
to make sure they don't attempt to share database connections.

We also make the in-memory database unique for each call to `local_run`
which simplifies testing.

### DIFF
--- a/jobrunner/local_run.py
+++ b/jobrunner/local_run.py
@@ -120,7 +120,10 @@ def create_and_run_jobs(
     docker.LABEL = docker_label
     config.LOCAL_RUN_MODE = True
     config.HIGH_PRIVACY_WORKSPACES_DIR = project_dir.parent
-    config.DATABASE_FILE = ":memory:"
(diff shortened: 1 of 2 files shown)


jobrunner/local_run.py
```diff
@@ -120,7 +120,10 @@ def create_and_run_jobs(
     docker.LABEL = docker_label
     config.LOCAL_RUN_MODE = True
     config.HIGH_PRIVACY_WORKSPACES_DIR = project_dir.parent
-    config.DATABASE_FILE = ":memory:"
+    # Append a random value so that multiple runs in the same process will each
+    # get their own unique in-memory database. This is only really relevant
+    # during testing.
+    config.DATABASE_FILE = f":memory:{random.randrange(sys.maxsize)}"
```
bloodearnest, line 126: Nice!

bloodearnest (approved):
Love the bonus feature.

PR #122, opened 2020-12-15
feat: add jobrunner.service for running as a single process.

bloodearnest:
This means manually parsing an envfile, and running the sync loop in a
thread.

Includes a hacky fix for the merged logging, simply prefixing sync logs with
'sync: '. I'd like to revisit logging at a later point and clean this up.


jobrunner/service.py
```diff
@@ -0,0 +1,55 @@
...
+    path = Path(os.environ.get('ENVPATH', '.env'))
+    if path.exists():
+        log.info(f'Loading environment variables from {path}')
+        env = parse_env(path.read_text())
+        if env:
+            os.environ.update(env)
+
+    try:
+        log.info("jobrunner.service started")
+        # daemon=True means this thread will be automatically join()ed when the
```
evansd, line 32: Ah yes, I always forget that this flag has the exact opposite meaning from the one I expect it to have

jobrunner/sync.py
```diff
@@ -19,13 +19,14 @@
...
 
 def main():
-    log.info(
-        f"Polling for JobRequests at: "
-        f"{config.JOB_SERVER_ENDPOINT.rstrip('/')}/job-requests/"
-    )
-    while True:
-        sync()
-        time.sleep(config.POLL_INTERVAL)
+    with set_log_context(prefix='sync'):
```
evansd, line 22: :+1: This looks like a reasonable re-use of the existing logging stuff to get what we want for now. I wonder if it would make sense to do this inside `service.py` though to localise the hack, as it were. It would mean wrapping `sync.main` in another function to use as the thread target, but would only be a couple of lines.

jobrunner/service.py
```diff
@@ -0,0 +1,55 @@
...
+        # process exits
+        thread = threading.Thread(target=sync.main, daemon=True)
+        thread.start()
+        run_main()
```
evansd, line 36: Don't you mean `run.main()`? Or am I missing something

evansd (changes requested):
Generally looks great, and pleasingly little code! But I think you've got a typo at the most crucial point!

jobrunner/sync.py
bloodearnest: Ok, that makes sense for now

jobrunner/service.py
bloodearnest: :facepalm: 

fixed and tests added.

evansd (approved):
:+1: Looks good

Might be worth chucking it through `black` and `flake8` as looks like you've got some odd formatting and unused imports, but nothing blocking.

sebbacon:
> Might be worth chucking it through black and flake8 as looks like you've got some odd formatting and unused imports, but nothing blocking

I see `pre-commit` is no longer in `requirements.txt`, which would have enforced this - is that deliberate?

evansd:
> I see pre-commit is no longer in requirements.txt, which would have enforced this - is that deliberate?

No, not deliberate. I was just clean-slating the requirements and missed it. (Though it would now belong in `requirements.dev.txt` rather than `requirements.txt`.)

bloodearnest:
> > I see pre-commit is no longer in requirements.txt, which would have enforced this - is that deliberate?
> 
> No, not deliberate. I was just clean-slating the requirements and missed it. (Though it would now belong in `requirements.dev.txt` rather than `requirements.txt`.)

Shall I add as part of this change?

evansd:
> Shall I add as part of this change?

I think we should do a separate "dev tooling" PR where we look at all that stuff.

PR #123, opened 2020-12-15
ci: add tooling to build dependencies repo

bloodearnest:
The idea is to be able to manage a repo we can directly use as dependencies on the server, as we only have access to github.com, not pypi.org. And we possibly don't even have pip.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ app.log
 *.egg-info/
 pip-wheel-metadata/*
 /workdir
+lib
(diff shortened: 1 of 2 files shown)


evansd (approved):
Nice!

We do have `pip` on the server (via `python -m pip`) but this approach avoids the extra install step we'd have if we shipped the wheels in.

bloodearnest:
> Nice!
> 
> We do have `pip` on the server (via `python -m pip`) but this approach avoids the extra install step we'd have if we shipped the wheels in.

Right. If we need to add more dependencies, especially binary ones, we can move to installing from wheels-in-box

PR #125, opened 2020-12-16
Usage Optimisations

ghickman:
Couple of small "optimisations", most for local dev when debugging API interactions, although the removal of empty job updates saves a bit of traffic in production.

### DIFF
--- a/jobrunner/sync.py
+++ b/jobrunner/sync.py
@@ -62,8 +62,16 @@ def api_request(method, path, *args, **kwargs):
     session.auth = (config.QUEUE_USER, config.QUEUE_PASS)
     response = session.request(method, url, *args, **kwargs)
 
-    if response.status_code == 400:
-        log.info("job-server returned 400: %s" % response.text)
+    log.debug(
+        "%s %s %s post_data=%s %s"
+        % (
+            method.upper(),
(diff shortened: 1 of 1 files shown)


jobrunner/sync.py
```diff
@@ -44,7 +44,9 @@ def sync():
     jobs = find_where(Job, job_request_id__in=job_request_ids)
     jobs_data = [job_to_remote_format(i) for i in jobs]
     log.debug(f"Syncing {len(jobs_data)} jobs back to job-server")
-    api_post("jobs", json=jobs_data)
+
+    if jobs_data:
+        api_post("jobs", json=jobs_data)
```
evansd, line 49: Now I think about it, it might be better to do this check earlier as it would save us a useless query to the db as well.
I was thinking on line 40:
```py
# Bail early if there's nothing to do
if not job_requests:
    return
```

evansd (approved):
Very sensible changes. I've made one small comment below.

Also, you'll need at least one commit to start with `fix: ` in order to pass the semantic PR check.

jobrunner/sync.py
ghickman: done and rebased in 2cb24dc

PR #128, opened 2020-12-18
ci: add windows service docs and scripts

bloodearnest:


sebbacon (approved):
Ahhh, `REM`-nostalgia

evansd (approved):
One thing worth noting is that when we're running natively we need a couple of extra bits of config on top of what's used for the docker service:
```
LOG_TIMESTAMPS='true'
ENABLE_PERMISSIONS_WORKAROUND='true'
```
```

bloodearnest:
> One thing worth noting is that when we're running natively we need a couple of extra bits of config on top of what's used for the docker service:
> 
> ```
> LOG_TIMESTAMPS='true'
> ENABLE_PERMISSIONS_WORKAROUND='true'
> ```

Ah, ok, good information.

I'll add these when I work test it next week.

PR #129, opened 2020-12-21
feat: refactor logging setup and formatting

bloodearnest:
Move the formatting logic into a filter rather than the formatter. This avoids
the need for parameterising the formatter, it can be customised just using a
regular format string, which I then used in the 2 places that need it
(local_run and threaded service).

Additionally, changed the log format to put various tags at the end of the
output. This pushes line noise out to the right, but still allows us to filter,
e.g:

    grep project=foo log.txt | grep status=bar

### DIFF
--- a/jobrunner/service.py
+++ b/jobrunner/service.py
@@ -24,6 +24,7 @@ log = logging.getLogger(__name__)
 
 def main():
     """Run the main run loop after starting the sync loop in a thread."""
+    # extra space to align with other thread's "sync" label.
     threading.current_thread().name = "run "
     fmt = "{asctime} {threadName} {message} {tags}"
     configure_logging(fmt)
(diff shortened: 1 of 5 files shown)


jobrunner/service.py
```diff
@@ -24,6 +24,9 @@
 
 def main():
     """Run the main run loop after starting the sync loop in a thread."""
+    threading.current_thread().name = "run "
```
ghickman, line 27: Did you mean to leave a trailing space here?

bloodearnest: Yes, so that it lines up with "sync" for the other thread. But worth a comment, will add.

inglesp:
image

ghickman (approved):

evansd (approved):
:+1: Definite improvement, thanks for doing this

PR #132, opened 2021-01-04
feat: parse .env file on module import.

bloodearnest:
This is done before we import config, so the env vars are set when it imports


evansd (approved):
Ah yeah, I should have spotted this issue earlier!

PR #134, opened 2021-01-05
fix: respect log context again and fix log status_code filtering

bloodearnest:
Also includes some black changes

### DIFF
--- a/jobrunner/log_utils.py
+++ b/jobrunner/log_utils.py
@@ -15,7 +15,7 @@ import threading
 import time
 
 
-DEFAULT_FORMAT="{asctime} {message} {tags}"
+DEFAULT_FORMAT = "{asctime} {message} {tags}"
 
 
 def formatting_filter(record):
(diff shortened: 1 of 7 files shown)


jobrunner/log_utils.py
```diff
@@ -26,20 +26,23 @@ def formatting_filter(record):
     record.status = ""
 
     tags = {}
-
+    ctx = set_log_context.current_context
+    job = getattr(record, "job", None) or ctx.get("job")
+    req = getattr(record, "job_request", None) or ctx.get("job_request")
+    
     if hasattr(record, "status_code"):
         record.status = record.status_code
```
evansd, line 34: I don't quite understand the renaming here

jobrunner/log_utils.py
```diff
@@ -103,7 +105,8 @@ def __init__(self, status_codes_to_ignore):
         self.status_codes_to_ignore = set(status_codes_to_ignore)
 
     def filter(self, record):
-        return record.status_code not in self.status_codes_to_ignore
+        status = getattr(record, "status", None)
```
evansd, line 108: Shouldn't this be `status_code`?

evansd (approved):
There were a couple of things that confused me here, but generally happy for it to go out

jobrunner/log_utils.py
bloodearnest: So, the thinking was .status is always available, .status_code may be.

It might be too confusing, though.

jobrunner/log_utils.py
bloodearnest: It's correct but confusing - status is always be available.

I think I'll change it, as it's not clear.

evansd:
Ah, OK I see where you were going with the `status` thing. I think it is a bit confusing though. That said, the fact that we have `state`, `status_message` and `status_code` is itself a bit confusing. Did you see this comment where I tried to explain my thinking?
https://github.com/opensafely/job-runner/blob/73bcbc3b8a42ed30f32439cf70f75bf5f7487eb1/jobrunner/models.py#L26-L35

PR #135, opened 2021-01-05
Switch to Token Auth

ghickman:
Use a token in the `Authorization` HTTP header for authentication with job-server.

This is the job-runner side of opensafely/job-server#303.

### DIFF
--- a/dotenv-sample
+++ b/dotenv-sample
@@ -1,12 +1,11 @@
 # The name of this backend
 OPENSAFELY_BACKEND=tpp
 
-# Credentials for logging into the job server
-OPENSAFELY_QUEUE_USER=user
-OPENSAFELY_QUEUE_PASS=pass
-
 # The endpoint to poll for jobs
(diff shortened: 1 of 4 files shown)


dotenv-sample
```diff
@@ -1,13 +1,12 @@
...
 
-# Credentials for logging into the job server
-OPENSAFELY_QUEUE_USER=user
-OPENSAFELY_QUEUE_PASS=pass
-
 # The endpoint to poll for jobs
 OPENSAFELY_JOB_SERVER_ENDPOINT=https://jobs.opensafely.org/jobs/
 
+# Credentials for logging into the job server
+export JOB_SERVER_TOKEN=pass
```
evansd, line 8: I think you want:
```
OPENSAFELY_JOB_SERVER_TOKEN=pass
```
(without the `export`)

evansd (approved):
Looks good, although the sample file needs amending (not that that's a blocker)

dotenv-sample
ghickman: ooh yes, definitely copy/pasta from my own `.envrc`!

PR #137, opened 2021-01-05
feat: refactor environment variable loading

bloodearnest:
1. Environment files are now loaded via `jobrunner/__init__.py`, which
ensures they are done early before the config module is imported.

2. Environment variables can be with or without their OPENSAFELY_
prefix. Previously we relied on docker-compose's .env parsing and
mapping the OPENSAFELY_ version to the plain version. This new way works
for both docker-compose and other methods (i.e. windows service).

3. Given a local .env file is parsed on import now, some test
configuration was needed to ensure correct config values for the tests to
run.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -6,31 +6,6 @@ services:
     image: ghcr.io/opensafely/job-runner:latest
     init: true
     restart: unless-stopped
-    environment:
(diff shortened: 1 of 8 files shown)


ghickman (approved):

bloodearnest:
Superseeded by #138

PR #138, opened 2021-01-05
fix: make windows service use a bash runner

bloodearnest:
This changes the nssm config to launch a bash script that sources .env
and then execs to python.

This means we do not need to do any env parsing in python, which is
good.

Also, we drop the OPENSAFELY_ prefix used by the current .env and
docker-compose tooling, just using the actual environment variables
directly.

NOTE: the .env file in prod will need updating to remove OPENSAFELY_ prefix when this is deployed.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -6,31 +6,6 @@ services:
     image: ghcr.io/opensafely/job-runner:latest
     init: true
     restart: unless-stopped
-    environment:
-      &jobrunner_environment
-      # The job server endpoint
-      - JOB_SERVER_ENDPOINT=${OPENSAFELY_JOB_SERVER_ENDPOINT}
-      # The backend requests which we should monitor
-      - BACKEND=${OPENSAFELY_BACKEND}
-      # Credentials for acessing the job server
-      - JOB_SERVER_TOKEN=${OPENSAFELY_JOB_SERVER_TOKEN}
-      # Working directory (see the volume mounts below)
-      - WORK_DIR=/work_dir
-      # A location where outputs should be stored
-      - HIGH_PRIVACY_STORAGE_BASE=${OPENSAFELY_HIGH_PRIVACY_STORAGE_BASE}
-      - MEDIUM_PRIVACY_STORAGE_BASE=${OPENSAFELY_MEDIUM_PRIVACY_STORAGE_BASE}
-      # A Github developer access key which can read private repos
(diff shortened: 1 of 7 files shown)


ghickman (approved):

docker-compose.yaml
```diff
@@ -6,31 +6,6 @@ services:
...
-    environment:
-      &jobrunner_environment
-      # The job server endpoint
-      - JOB_SERVER_ENDPOINT=${OPENSAFELY_JOB_SERVER_ENDPOINT}
-      # The backend requests which we should monitor
-      - BACKEND=${OPENSAFELY_BACKEND}
-      # Credentials for acessing the job server
-      - JOB_SERVER_TOKEN=${OPENSAFELY_JOB_SERVER_TOKEN}
-      # Working directory (see the volume mounts below)
-      - WORK_DIR=/work_dir
```
evansd, line 18: This still needs setting as it's not in the `.env` file. That's because it's a reference to a mount point defined within docker-compose that doesn't make sense in any other context.

docker-compose.yaml
```diff
@@ -6,31 +6,6 @@ services:
...
-      # Database URLs
-      - FULL_DATABASE_URL=${OPENSAFELY_FULL_DATABASE_URL}
-      - SLICE_DATABASE_URL=${OPENSAFELY_SLICE_DATABASE_URL}
-      - DUMMY_DATABASE_URL=${OPENSAFELY_DUMMY_DATABASE_URL}
-      - TEMP_DATABASE_NAME=${OPENSAFELY_TEMP_DATABASE_NAME}
-      # Polling
-      - POLL_INTERVAL=${OPENSAFELY_POLL_INTERVAL}
-      - JOB_LOOP_INTERVAL=${OPENSAFELY_JOB_LOOP_INTERVAL}
-      # Parallelism
-      - MAX_WORKERS=${OPENSAFELY_MAX_WORKERS}
```
evansd, line 33: I don't think this will work as is (though apologies if you've tested it and it does in fact work!) because as I understand it `.env` only sets defaults for environment variables but doesn't automatically pass them through unless you explicitly list them. I think you need the [env_file](https://docs.docker.com/compose/environment-variables/#the-env_file-configuration-option) option if you want to say "pass through all vars in this file". I guess we could use `env_file: ".env"` to do this.

evansd (changes requested):

evansd (approved):

docker-compose.yaml
bloodearnest: TIL.

I tested this version w/docker-compose, it seems to work correctly.

PR #140, opened 2021-01-07
feat: Add command to kill jobs

evansd:
It turns out we need this when certain jobs are in danger of using up
all the disk space.


bloodearnest (approved):
Good tool to have, although I don't think it needs docker compose to run in the long term its fine for now

PR #142, opened 2021-01-11
ci: refactor windows install script to be less shoddy

bloodearnest:
- it no longer removes the service, as that breaks the security set up.
 - explicit path for nssm
 - better output, suitable for copy/pasting
 - better error handling

### DIFF
--- a/scripts/install.bat
+++ b/scripts/install.bat
@@ -21,7 +21,7 @@ if %ERRORLEVEL% NEQ 0 (
 REM now we know the service definitely exists, configure it
 
 %nssm% set opensafely Application "C:\Program Files\Git\usr\bin\bash"
-%nssm% set opensafely AppParameters "%directory%\scripts\run.sh"
+%nssm% set opensafely AppParameters "%directory%\scripts\run.sh -m jobrunner.service"
 %nssm% set opensafely AppDirectory %directory%
 %nssm% set opensafely DisplayName "OpenSAFELY job runner"
 %nssm% set opensafely Start SERVICE_AUTO_START
--- a/scripts/install.bat
+++ b/scripts/install.bat
@@ -2,12 +2,15 @@
 REM This scripts installs/configures the opensafely service, designed to be
 REM installed on the opensafely TPP backend
 
+REM We make some assumtions about paths for the TPP backend
 set directory=E:\job-runner
 set nssm=C:\nssm-2.24\win64\nssm.exe
 
+REM we require both user name and a password to run, so check
 if "%1" == "" GOTO :usage
 if "%2" == "" GOTO :usage
 
+REM check the service is installed, and either stop it or install it.
 sc query opensafely | find "does not exist" >nul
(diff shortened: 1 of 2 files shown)


scripts/install.bat
```diff
@@ -1,25 +1,52 @@
...
-nssm set opensafely AppRestartDelay 1000
+@echo off
+REM This scripts installs/configures the opensafely service, designed to be
+REM installed on the opensafely TPP backend
+
+set directory=E:\job-runner
+set nssm=C:\nssm-2.24\win64\nssm.exe
+
+if "%1" == "" GOTO :usage
+if "%2" == "" GOTO :usage
```
ghickman, line 9: Can you put a comment before these two lines to explain that they're checking both arguments exist?  (assuming that's what it does!)

scripts/install.bat
```diff
@@ -1,25 +1,52 @@
...
-nssm set opensafely AppExit Default Restart
-nssm set opensafely AppRestartDelay 1000
+@echo off
+REM This scripts installs/configures the opensafely service, designed to be
+REM installed on the opensafely TPP backend
+
+set directory=E:\job-runner
+set nssm=C:\nssm-2.24\win64\nssm.exe
+
+if "%1" == "" GOTO :usage
```
evansd, line 8: Ah, GOTO my old friend ...

ghickman (changes requested):
Just the one quick change, the inline if really caught me out so probably best to be explicit about those checks.

Otherwise this looks great, pulling `nssm` out into a variable really makes the other lines a lot less dense 👍

evansd (approved):
:+1: I can almost feel the painful discovery process behind every line ... Thanks for taking this on

scripts/install.bat
bloodearnest: Beyond the fact that we require both arguments?

ghickman: Yep, just to make it clear that's what we're enforcing here.  I've approached this in terms of "if Simon is not around and I have to debug this, which lines made me pause to consider what they do", so having an idea that arguments 1&2 are necessary is really useful.

bloodearnest: Done

ghickman (approved):

evansd:
@bloodearnest Quick thought here: is it easy enough to change the `run.sh` exec line to use `"@"` and then make the service do `run.sh -m jobrunner.service`

This would make `run.sh` a generic "run python with the environment set up correctly" script which we could then use for running `retry_job`, `kill_job` etc. (Assuming we can fix db permissions.)

bloodearnest:
> @bloodearnest Quick thought here: is it easy enough to change the `run.sh` exec line to use `"@"` and then make the service do `run.sh -m jobrunner.service`
> 
> This would make `run.sh` a generic "run python with the environment set up correctly" script which we could then use for running `retry_job`, `kill_job` etc. (Assuming we can fix db permissions.)

This is a Good Idea, will do.

evansd (approved):

PR #144, opened 2021-01-11
fix: Generate consistent job IDs

evansd:
Rather than randomly generating a job ID each time, we now use the hash
of the job request ID and the action name – which together are
guaranteed unique. This means even if we lose the entire job-runner
database while jobs are running we should be able to gracefully
recover because the same set of jobs with the same IDs should be
automatically recreated by the sync service.

Losing state like this still won't be _entirely_ unproblematic: in
particular, jobs which have already completed but are part of a job
request which is still active won't have their statuses correctly
reported. But it will at least avoid the problem of orphan jobs which
require a manual cleanup.

### DIFF
--- a/jobrunner/add_job.py
+++ b/jobrunner/add_job.py
@@ -3,10 +3,12 @@ Development utility for creating and submitting a JobRequest without having a
 job-server
 """
 import argparse
+import base64
 import dataclasses
 import pprint
 from pathlib import Path
 from urllib.parse import urlparse
+import secrets
 import textwrap
 
 from .log_utils import configure_logging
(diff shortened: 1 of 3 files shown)


sebbacon (approved):
LGTM

I if we should document what the consequences or a loss of state followed by restart are/could be (not sure where, though)

ghickman (approved):

evansd:
I did spend some time thinking this through at one point and I think the answer is "it's complicated and we should try to avoid it ever happening". But this change fixes one issue without (I'm reasonably confident) creating other issues so I thought it was worth chucking in.

I think it would theoretically be possible to fiddle things around so that the database become totally ephemeral and could be safely deleted at any point. But it would add quite a bit of complexity and I didn't feel it was worth it.

bloodearnest:
Question - why use a hash? Could we just use '{request_id}-{action}'? That would be more transparent to us developers when debugging, I think.

evansd:
As I say in the comments, I think it would make sense to move towards that. But switching from a fixed length alphanumeric ID to a variable length one with a larger character set is a more invasive change than I wanted to make as a drive-by.

PR #145, opened 2021-01-11
fix: Pass environment through when fetching license

evansd:
Without this (at least on my particular Linux setup) git can't
authenticate with Github.


bloodearnest (approved):

bloodearnest:
Interesting. This worked for me without my environment being present. But it makes sense to include it.

PR #146, opened 2021-01-14
fix: Don't pass env vars on command line to avoid leakage

evansd:


jobrunner/docker.py
```diff
@@ -278,10 +279,15 @@ def run(name, args, volume=None, env=None, allow_network_access=False, label=Non
...
-            run_args.extend(["--env", f"{key}={value}"])
-    subprocess_run(run_args + args, check=True, capture_output=True)
+    # To avoid leaking the values into the command line arguments we set them
+    # in the evnironment and tell Docker to fetch them from there
+    if env is None:
+        env = {}
+    for key, value in env.items():
+        run_args.extend(["--env", key])
+    subprocess_run(
+        run_args + args, check=True, capture_output=True, env=dict(os.environ, **env)
```
bloodearnest, line 289: Hmm, not sure about passing all of os.environ in here, as that exposes e..g DB creds to the docker container?

Would env=env not work? Or perhaps a whitelist of env keys from os.environ?

bloodearnest (changes requested):
This is great, much neater approach that I was l looking at.

I think there may be security issue though, easily fixable.

jobrunner/docker.py
bloodearnest: Oh, it only exposes them to docker, not to the action, my bad

bloodearnest: But still, what is the need for providing os.environ? Can we just get away with env?

bloodearnest (approved):
Ignore me, I was being dumb.

evansd:
@bloodearnest FWIW, I went through exactly the same thought process in my head earlier :)

I think the argument for passing all the environ is that that's what we currently do by default (by not setting `env` at all) and for things like changing `DOCKER_HOST` when testing from Windows it's useful if `docker` inherits the environment.

PR #151, opened 2021-01-20
Log data being sent to job-server

ghickman:
We have a log line for data coming from the server, this mirrors that in an effort to debug opensafely/job-server/issues/335 (and any other maybe-sync issues!).


bloodearnest (commented):
How big is this data gonna be do we think?

ghickman:
Not that big I think, we're generating this dictionary for each Job: https://github.com/opensafely/job-runner/blob/c10d598e9ab98fecdf1ac2b178ecc7d8fb8f8495/jobrunner/sync.py#L109-L119

bloodearnest (approved):

PR #165, opened 2021-02-02
Move to opensafely-core

bloodearnest:


evansd (approved):
Looks good, but I think you'll need some kind of semantic commit tag to trigger a new PyPI release (which we'll need to get it into opensafely-cli).

PR #174, opened 2021-02-26
feat: add PRESTO_TLS_* configs

bloodearnest:
PRESTO_TLS_KEY_PATH and PRESTO_TLS_CERT_PATH contain paths to the
relevant files, and we load them into PRESTO_TLS_KEY and PRESTO_TLS_CERT
respectively. If present, these values are passed to the cohortextractor
invocation.

This involved some error handing with in config.py that I wanted to
test, so I wrote a stupid subprocess test harness to do so.

### DIFF
--- a/jobrunner/config.py
+++ b/jobrunner/config.py
@@ -59,25 +59,30 @@ PRESTO_TLS_KEY = PRESTO_TLS_CERT = None
 PRESTO_TLS_CERT_PATH = os.environ.get("PRESTO_TLS_CERT_PATH")
 PRESTO_TLS_KEY_PATH = os.environ.get("PRESTO_TLS_KEY_PATH")
 
-if PRESTO_TLS_KEY_PATH and PRESTO_TLS_CERT_PATH:
-    key_path = Path(PRESTO_TLS_KEY_PATH)
-    cert_path = Path(PRESTO_TLS_CERT_PATH)
+if bool(PRESTO_TLS_KEY_PATH) != bool(PRESTO_TLS_CERT_PATH):
+    raise ConfigException(
+            "Both PRESTO_TLS_KEY_PATH and PRESTO_TLS_CERT_PATH must be defined if either are"
+    )
 
-    if key_path.exists() and cert_path.exists():
+if PRESTO_TLS_KEY_PATH:
+    key_path = Path(PRESTO_TLS_KEY_PATH)
+    if key_path.exists():
         PRESTO_TLS_KEY = key_path.read_text()
-        PRESTO_TLS_CERT = cert_path.read_text()
-    elif key_path.exists():
-        raise ConfigException(
-            f"PRESTO_TLS_CERT_PATH={cert_path}, but file does not exist"
-        )
-    elif cert_path.exists():
+    else:
         raise ConfigException(
             f"PRESTO_TLS_KEY_PATH={key_path}, but file does not exist"
         )
+
+if PRESTO_TLS_CERT_PATH:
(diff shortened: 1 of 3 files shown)


jobrunner/config.py
```diff
@@ -51,6 +51,9 @@
 
 TEMP_DATABASE_NAME = os.environ.get("TEMP_DATABASE_NAME")
 
+PRESTO_TLS_KEY = os.environ.get("PRESTO_TLS_KEY")
+PRESTO_TLS_CERT = os.environ.get("PRESTO_TLS_CERT")
```
evansd, line 55: Just a thought: do we want to pass in paths here and have `config.py` read their contents. I suppose it would minimise the amount of time these values spend floating around in environment variables, and would mean that we don't have them on screen every time we edit the `.env` file.

bloodearnest: That's a good idea

bloodearnest: It also doesn't leave them in the processes environment table while its running

jobrunner/config.py
```diff
@@ -51,6 +55,30 @@
 
 TEMP_DATABASE_NAME = os.environ.get("TEMP_DATABASE_NAME")
 
+PRESTO_TLS_KEY = PRESTO_TLS_CERT = None
+PRESTO_TLS_CERT_PATH = os.environ.get("PRESTO_TLS_CERT_PATH")
+PRESTO_TLS_KEY_PATH = os.environ.get("PRESTO_TLS_KEY_PATH")
+
+if PRESTO_TLS_KEY_PATH and PRESTO_TLS_CERT_PATH:
```
evansd, line 62: If you set one value but not the other (e.g. because you typo'd it) then I think this ends up silently ignoring both values.

I wonder if it would be better to do the two checks independently like:
```py
if PRESTO_TLS_KEY_PATH:
    key_path = Path(PRESTO_TLS_KEY_PATH)
    if key_path.exists():
        PRESTO_TLS_KEY = key_path.read_text()
    else:
        raise ConfigException(
            f"PRESTO_TLS_KEY_PATH={key_path}, but file does not exist"
        )

if PRESTO_TLS_CERT_PATH:
    cert_path = Path(PRESTO_TLS_CERT_PATH)
    if cert_path.exists():
        PRESTO_TLS_CERT = cert_path.read_text()
    else:
        raise ConfigException(
            f"PRESTO_TLS_CERT_PATH={cert_path}, but file does not exist"
        )

if bool(PRESTO_TLS_KEY_PATH) != bool(PRESTO_TLS_CERT_PATH):
    raise ConfigException(
            "Both PRESTO_TLS_KEY_PATH and PRESTO_TLS_CERT_PATH must be defined if either are"
    )
```

evansd:
Wow, Windows failure is a new one on me:
```
    def test_config_presto_paths(tmp_path):
        key = tmp_path / "key"
        key.write_text("key")
        cert = tmp_path / "cert"
        cert.write_text("cert")
        cfg, err = import_cfg(
            {"PRESTO_TLS_KEY_PATH": str(key), "PRESTO_TLS_CERT_PATH": str(cert)}
        )
>       assert err is None
E       AssertionError: assert 'Fatal Python error: _Py_HashRandomization_Init: failed to get random numbers to initialize Python\n\n' is None
```

evansd (approved):
Looks good to me. A bit of bikeshedding about config stuff but all ignoreable stuff.

jobrunner/config.py
bloodearnest: Good call, done

evansd (approved):
Nice

PR #175, opened 2021-03-01
fix: pass through EMIS_ORGANISATION_HASH to cohortextractor

bloodearnest:


jobrunner/manage_jobs.py
```diff
@@ -84,6 +84,8 @@ def start_job(job):
             if config.PRESTO_TLS_KEY and config.PRESTO_TLS_CERT:
                 env["PRESTO_TLS_CERT"] = config.PRESTO_TLS_CERT
                 env["PRESTO_TLS_KEY"] = config.PRESTO_TLS_KEY
+            if config.EMIS_ORGANISATION_HASH:
+                env["EMIS_ORGANISATION_HASH"] = config.EMIS_ORGANISATION_HASH
```
evansd, line 88: Idle thought: I wonder if at some point we should switch to a `COHORT_EXTRACTOR_ENV` config dict that we pass through here as the number of vars here just keeps growing.

evansd (approved):

PR #176, opened 2021-03-01
feat: Add `--debug` flag to disable cleanup

evansd:
This adds a `--debug` flag to the local_run command which leaves all
Docker containers and volumes in place for debugging purposes.

### DIFF
--- a/jobrunner/local_run.py
+++ b/jobrunner/local_run.py
@@ -77,20 +77,38 @@ def add_arguments(parser):
         help="Don't stop on first failed action",
         action="store_true",
     )
+    parser.add_argument(
+        "--debug",
+        help="Leave docker containers and volumes in place for debugging",
+        action="store_true",
+    )
     return parser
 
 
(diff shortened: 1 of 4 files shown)


bloodearnest (approved):
LGTM

I'd love to figure out a way to test more of the docker related stuff some point soon.

jobrunner/local_run.py
```diff
@@ -77,20 +77,38 @@ def add_arguments(parser):
         help="Don't stop on first failed action",
         action="store_true",
     )
+    parser.add_argument(
+        "--debug",
```
sebbacon, line 81: I feel `--debug` is slightly overloaded - implies "log verbosely" - but I can't think of a better alternative!

sebbacon (approved):

jobrunner/local_run.py
evansd: It is a bit. But I sort of imagine this flag will increase log verbosity as well at some point when I get round to it.

PR #182, opened 2021-03-11
Support config file giving job "resource weightings"

evansd:
This gives us a very crude mechanism for controlling how many of what
type of jobs are running together.

The config file itself should be placed at `job-resource-weights.ini`
in the job-runner working directory.

It is structured as:
```ini

[some-workspace-name]
my-ram-hungry-action = 4
other-actions.* = 1.5

[other-workspace-name]
...
```

Any jobs in the specified workspace will have their action names matched
against the regex patterns specified in the config file and will be
assigned the weight of the first matching pattern. All other jobs are
assigned a weight of 1.

We also take this opportunity to randomly shuffle the jobs list before each
processing loop so as to prevent a single set of jobs hogging all available
resources.

### DIFF
--- a/jobrunner/config.py
+++ b/jobrunner/config.py
@@ -1,5 +1,7 @@
+import configparser
 import os
 from pathlib import Path
+import re
 from multiprocessing import cpu_count
 
 
@@ -100,3 +102,35 @@ STATA_LICENSE_REPO = os.environ.get(
     "STATA_LICENSE_REPO",
     "https://github.com/opensafely/server-instructions.git",
 )
+
+
+def parse_job_resource_weights(config_file):
+    """
+    Parse a simple ini file which looks like this:
+
+        [some-workspace-name]
+        my-ram-hungry-action = 4
+        other-actions.* = 1.5
+
+        [other-workspace-name]
+        ...
+
+    Any jobs in the specified workspace will have their action names matched
+    against the regex patterns specified in the config file and will be
+    assigned the weight of the first matching pattern. All other jobs are
+    assigned a weight of 1.
+    """
+    config_file = Path(config_file)
+    weights = {}
+    if config_file.exists():
+        config = configparser.ConfigParser()
+        config.read_string(config_file.read_text(), source=str(config_file))
+        for workspace in config.sections():
+            weights[workspace] = {
+                re.compile(pattern): float(weight)
+                for (pattern, weight) in config.items(workspace)
+            }
+    return weights
+
+
+JOB_RESOURCE_WEIGHTS = parse_job_resource_weights("job-resource-weights.ini")
--- a/jobrunner/run.py
+++ b/jobrunner/run.py
@@ -11,7 +11,7 @@ import time
 
 from .log_utils import configure_logging, set_log_context
 from . import config
-from .database import find_where, count_where, update, select_values
+from .database import find_where, update, select_values
 from .models import Job, State, StatusCode
 from .manage_jobs import (
     JobError,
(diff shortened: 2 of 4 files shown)


jobrunner/config.py
```diff
@@ -100,3 +102,35 @@ class ConfigException(Exception):
...
+    assigned a weight of 1.
+    """
+    config_file = Path(config_file)
+    weights = {}
+    if config_file.exists():
+        config = configparser.ConfigParser()
+        config.read_string(config_file.read_text(), source=str(config_file))
+        for workspace in config.sections():
+            weights[workspace] = {
+                re.compile(pattern): float(weight)
```
bloodearnest, line 130: Not seen compiled regex's as keys before. Nice idea, makes a lot of sense.

jobrunner/run.py
```diff
@@ -24,17 +25,25 @@
...
 
-def handle_jobs(raise_on_failure=False):
+def handle_jobs(raise_on_failure=False, shuffle_jobs=True):
     active_jobs = find_where(Job, state__in=[State.PENDING, State.RUNNING])
+    # Randomising the job order is a crude but effective way to ensure that a
+    # single large job request doesn't hog all the workers. We make this
+    # optional as, when running locally, having jobs run in a predictable order
+    # is preferable
+    if shuffle_jobs:
+        random.shuffle(active_jobs)
```
bloodearnest, line 46: I wonder about shuffling a subset of the jobs, e.g. the top 5 or top len() * 0.5

This preserves a degree of fair queuing - jobs will queue in FIFO order until they enter the top n - whilst keeping the head-of-line unblocking of randomisation.

Randomising the entire list means a job just submitted might gazump one that's been waiting for hours. I also suspect full randomisation might mean highly weighted jobs are often delayed as they lose their chance to be considered before the resources are filled up with small jobs. If there's *some* ordering to the queue, this keeps some fairness next to the unblocking randomness.

jobrunner/config.py
```diff
@@ -100,3 +102,35 @@ class ConfigException(Exception):
...
+                for (pattern, weight) in config.items(workspace)
+            }
+    return weights
+
+
+JOB_RESOURCE_WEIGHTS = parse_job_resource_weights("job-resource-weights.ini")
```
bloodearnest, line 136: Local config file is fine for now, and you've probably already considered this, but:

I think in future the weight should come in from the job-server with the JobRequest.  Doing so would work across backends, and allow us to automatically generate those weights based on stats in the future without changing the job-runner. We could also allow users to prioritse amongst their own (or their groups) queued jobs in the same way.

bloodearnest (approved):
Looks good as a quick attempt to address the current problem.

I've left one suggestion, and a comment regards future design of queuing.

jobrunner/config.py
evansd: Yes, I think putting the job-server in control of this makes a lot of sense longer term, assuming it has access to the necessary data. The weights thing was just a quick and dirty way of solving the immediate problem of not running too many high memory jobs at the same time and I'm not sure we'll keep using that. I'd imagine we'll track expected RAM, CPU and disk usage independently and make decisions based on availability of specific resources.

jobrunner/run.py
evansd: I'm not sure that will do what we need. The specific case that prompted this was someone submitting 70 odd jobs which they expected to take all weekend to run and this meaning that no one else could run anything until it was finished. I think having a newly submitted job gazump an old job is fine in the circumstances.

It's a good point that highly weighted jobs won't get scheduled if there are loads of of smaller jobs in the queue. I don't think that's a disaster either though. It's an improvement on the status quo. And the queue spends most of the time empty anyway at the moment, so in practice I don't think it will be a problem.

PR #183, opened 2021-03-15
Add crude service to log system stats to a SQLite file

evansd:
This dumps the output of various docker and system load viewing commands as timestamped blobs of JSON in a SQLite file.

Configured using:

    STATS_DATABASE_FILE
    STATS_POLL_INTERVAL

Deliberately added as a new service rather than bundled up with `sync` so that we can back it out easily when we replace this with something better.

### DIFF
--- a/jobrunner/docker_stats.py
+++ b/jobrunner/docker_stats.py
@@ -1,0 +1,52 @@
+import json
+
+from .subprocess_utils import subprocess_run
+
+
+def get_volume_and_container_sizes():
+    response = subprocess_run(
+        ["docker", "system", "df", "--verbose", "--format", "{{json .}}"],
+        capture_output=True,
+        check=True,
+    )
+    data = json.loads(response.stdout)
+    volumes = {row["Name"]: _parse_size(row["Size"]) for row in data["Volumes"]}
+    containers = {row["Names"]: _parse_size(row["Size"]) for row in data["Containers"]}
+    return volumes, containers
+
+
+def get_container_stats():
+    response = subprocess_run(
+        ["docker", "stats", "--no-stream", "--format", "{{json .}}"],
+        capture_output=True,
+        check=True,
+    )
+    data = [json.loads(line) for line in response.stdout.splitlines()]
+    return {
+        row["Name"]: {
+            "cpu_percentage": float(row["CPUPerc"].rstrip("%")),
+            "memory_used": _parse_size(row["MemUsage"].split()[0]),
+        }
+        for row in data
+    }
+
+
+CONVERSIONS = {
+    "B": 1,
+    "KB": 10 ** 3,
+    # Is this correct?
+    "kB": 10 ** 3,
+    "KiB": 2 ** 10,
+    "MB": 10 ** 6,
+    "MiB": 2 ** 20,
+    "GB": 10 ** 9,
+    "GiB": 2 ** 30,
+    "TB": 10 ** 12,
+    "TiB": 2 ** 40,
+}
+
+
+def _parse_size(size):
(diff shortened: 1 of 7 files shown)


jobrunner/docker_stats.py
```diff
@@ -0,0 +1,52 @@
...
+    "MiB": 2 ** 20,
+    "GB": 10 ** 9,
+    "GiB": 2 ** 30,
+    "TB": 10 ** 12,
+    "TiB": 2 ** 40,
+}
+
+
+def _parse_size(size):
+    units = size.lstrip("0123456789.")
```
bloodearnest, line 50: Nice!

jobrunner/record_stats.py
```diff
@@ -0,0 +1,80 @@
...
+        log_stats(connection)
+        time.sleep(config.STATS_POLL_INTERVAL)
+
+
+def get_database_connection(filename):
+    filename.parent.mkdir(parents=True, exist_ok=True)
+    conn = sqlite3.connect(filename)
+    # Enable autocommit
+    conn.isolation_level = None
+    schema_count = list(conn.execute("SELECT COUNT(*) FROM sqlite_master"))[0][0]
```
bloodearnest, line 45: This is nice an simple.

Also saw this nice way of doing native sqlite migrations - don't think we need it in this case, though

https://levlaz.org/sqlite-db-migrations-with-pragma-user_version/

tests/test_system_stats.py
```diff
@@ -0,0 +1,14 @@
...
+
+@pytest.mark.needs_docker
+def test_get_system_stats():
+    stats = get_system_stats()
+    assert {
+        "total_disk_space",
+        "available_disk_space",
+        "total_memory",
+        "available_memory",
+    } <= stats.keys()
```
bloodearnest, line 14: TIL you can use order comparison on sets/dict views! This is nice, but was also very confusing! But I don't know a nicer way of expressing that, since dict views are annoyingly not quite sets

ghickman (approved):

bloodearnest (approved):
LGTM

tests/test_docker_stats.py
```diff
@@ -0,0 +1,39 @@
...
+import pytest
+
+from jobrunner import docker
+from jobrunner.docker_stats import get_container_stats, get_volume_and_container_sizes
+
+
+@pytest.mark.needs_docker
+# This runs file locally but fails in CI, despite the retry logic added below.
+# Don't have time to diagnose this properly and this isn't core functionality
+# in any case
```
sebbacon, line 12: :sob:

jobrunner/service.py
```diff
@@ -50,5 +55,16 @@ def sync_wrapper():
             time.sleep(sleep_after_error)
 
 
+def record_stats_wrapper():
+    """Wrap the record_stats call with logging context and an exception handler."""
+    while True:
+        try:
+            record_stats.main()
```
sebbacon, line 62: What's the return for, OOI?

evansd: Good point, have added comment:
```py
            # `main()` should loop forever, if it exits cleanly that means it
            # wasn't configured to run so we should exit the thread rather than
            # looping
```

sebbacon (approved):
I actually love this.
I actually really like this!

Little thought, we already support labelling a docker container when we run it; if record job id  then we can also write that to the stats db. Would make spelunking a bit easier?

evansd:
I don't know if it would make that much difference really. The job id is always the last segment of the container name so it's easily greppable/extractable with what we already have.

PR #185, opened 2021-03-18
docs: Add script for mounting job volumes

evansd:


scripts/mount-job-volume.sh
```diff
@@ -0,0 +1,68 @@
...
+if [[ -n "$name_search" && "$volume_count" == 1 ]]; then
+
+    selected_volume=${volume_array[0]}
+
+# Otherwise display a menu
+else
+
+  echo $"Found $volume_count $verb volumes:\n"
+  PS3=$'\nEnter volme number: '
+  select selected_volume in ${volume_array[@]}; do
```
bloodearnest, line 60: Huh, not seen that one before! 

There's always a more bashy way to do things!

bloodearnest (approved):

scripts/mount-job-volume.sh
evansd: It was a copy/paste/edit of the `watch-job-logs.sh` script and I've now entirely forgotten where I got that pattern from!

PR #190, opened 2021-03-23
fix: add option for local_run to emit timestamps

bloodearnest:
This is useful in CI settings.

### DIFF
--- a/jobrunner/local_run.py
+++ b/jobrunner/local_run.py
@@ -77,6 +77,12 @@ def add_arguments(parser):
         help="Don't stop on first failed action",
         action="store_true",
     )
+    # This particularly useful in CI.
+    parser.add_argument(
+        "--timestamps",
+        help="Include timestamps in output",
+        action=store_true,
+    )
     parser.add_argument(
         "--debug",
         help="Leave docker containers and volumes in place for debugging",
@@ -91,6 +97,7 @@ def main(
     force_run_dependencies=False,
     continue_on_error=False,
     debug=False,
+    timestamps=False,
 ):
     if not docker_preflight_check():
         return False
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,9 +1,9 @@
-VENV_DIR ?= .venv
```
evansd, line 1: What's with the Makefile changes?

jobrunner/local_run.py
```diff
@@ -77,6 +77,11 @@ def add_arguments(parser):
         help="Don't stop on first failed action",
         action="store_true",
     )
+    parser.add_argument(
+        "--timestamps",
+        help="Emit timestamps on log out. Useful in CI",
```
evansd, line 82: Nitpick, but the primary audience for the help text here are researchers looking at `opensafely run --help`. So I think the CI comment is just extra noise and potential confusion here. "Log" is also slightly confusing here as there are separate log file which always include timestamps. So I'd just go for:
```
Include timestamps in output
```
The CI thing could go in a Python comment above. And actually, would be worth doing the same for `--continue-on-error`.

evansd (approved):

Makefile
bloodearnest: It was broken, so I fixed it so I could use it to run the tests

jobrunner/local_run.py
bloodearnest: Good call. There's already such a comment on the `--continue-on-error` definition :)

Makefile
evansd: Ah, I see. Feels like it belongs a separate commit in that case.

bloodearnest: reverted

evansd: :+1: didn't necessarily mean a separate PR btw. Drive by fixes are great, but in my experience it pays off over time to keep commits logically atomic (as far as reasonably possible)

PR #191, opened 2021-03-23
fix: timestamps cli arg definition was broken

bloodearnest:
Added a test to ensure local_run cli args work.


evansd (approved):
Given that it's exposed the complete lack of testing of `add_arguments` I'm very glad you made this typo!

PR #194, opened 2021-03-25
fix: Safely handle case-insensitive filesystems

evansd:
Previously, if the case of an output file changed we would end up
deleting it every time we copied it out :(

The WindowsPath object already contains some case-normalisation logic
for comparison/hashing, but we also need to run correctly on macOS we is
case-insensitive but uses PosixPath. Doing the comparisons directly with
inodes is a bit of blunt tool but should give us the correct behaviour
here.


ghickman (approved):
🍻

PR #202, opened 2021-04-29
feat: Add script to extract usage stats

evansd:
This is quick and dirty, but I've done something like this enough times
now that it's worth scripting up.


inglesp (approved):
Neat.

PR #203, opened 2021-04-30
Validate that commits belong to OpenSAEFLY repos

evansd:
This is already enforced by the job-server but this means that even if
job-server is compromised it's impossible to run code unless you have
commit access to an OpenSAFELY repo.

### DIFF
--- a/jobrunner/config.py
+++ b/jobrunner/config.py
@@ -104,6 +104,11 @@ STATA_LICENSE_REPO = os.environ.get(
 )
 
 
+ALLOWED_GITHUB_ORGS = (
+    os.environ.get("ALLOWED_GITHUB_ORGS", "opensafely").strip().split(",")
+)
+
+
 def parse_job_resource_weights(config_file):
     """
     Parse a simple ini file which looks like this:
--- a/jobrunner/git.py
+++ b/jobrunner/git.py
@@ -71,6 +71,32 @@ def checkout_commit(repo_url, commit_sha, target_dir):
     )
 
 
+def commit_reachable_from_ref(repo_url, commit_sha, ref):
+    """
+    Given a `ref` (branch name, tag, etc) on a remote repo, check whether the
+    supplied commit is reachable from that ref.
+    """
+    ref_sha = get_sha_from_remote_ref(repo_url, ref)
+    # The easy case and the case I expect to be hit almost every time as the UI
+    # currently only supports running against the branch head
+    if commit_sha == ref_sha:
+        return True
+    # However a well (or badly) timed push could cause the target sha and the
+    # branch head to diverge, so we need to handle that. In order to do so we
+    # need to fetch the history of the branch. We first fetch just the last 10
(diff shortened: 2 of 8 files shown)


jobrunner/git.py
```diff
@@ -71,6 +71,32 @@ def checkout_commit(repo_url, commit_sha, target_dir):
...
+    if commit_sha == ref_sha:
+        return True
+    # However a well (or badly) timed push could cause the target sha and the
+    # branch head to diverge, so we need to handle that. In order to do so we
+    # need to fetch the history of the branch. We first fetch just the last 10
+    # commits on the assumption that it's probably one of those. If that fails
+    # we fetch the entire branch history.
+    repo_dir = get_local_repo_dir(repo_url)
+    ensure_git_init(repo_dir)
+    fetch_commit(repo_dir, repo_url, ref_sha, depth=10)
```
ghickman, line 91: Not at all relevant to this PR but these three lines feel like something we do in various places, would an abstraction to them be useful at some point?

tests/test_git.py
```diff
@@ -54,6 +59,22 @@ def test_read_file_from_private_repo(tmp_work_dir):
...
+        "2170fadfeca7baf04f50376677253513e0a12bb1",
+        "1.6.0",
+    )
+    assert should_be_reachable
+    should_not_be_reachable = commit_reachable_from_ref(
+        "https://github.com/opensafely/documentation",
+        "afaee06237df52d78aa85920248572d5803379a3",
+        "1.6.0",
+    )
+    assert not should_not_be_reachable
```
ghickman, line 75: This is a bit of a brain teaser!

Naming suggestion (feel free to ignore 🙂) - call both variables `reachable` (or number them to distinguish), and then use the `assert`s to show the expectation on their outcome.

jobrunner/config.py
```diff
@@ -104,6 +104,9 @@ class ConfigException(Exception):
 )
 
 
+ALLOWED_GITHUB_ORGS = os.environ.get("ALLOWED_GITHUB_ORGS", "opensafely").split()
```
ghickman, line 107: Would splitting on comma cause less potential confusion here?  It's what a few libraries do for "list" env vars, and IIRC shells typically treat space as a delimiter so could cause issues there?

ghickman (approved):
Looks good, just a quick thought on naming 👍

jobrunner/git.py
evansd: I had similar feelings initially but we only actually use them twice and in the other context there's a no-op check (`commit_already_fetched`) which is only applicable when `depth=1` so I think it would end up being less readable to try to make an abstraction out of them.

tests/test_git.py
evansd: Ha, yes! I obviously never vocalised this line in my head. Your suggestion is much better.

jobrunner/config.py
evansd: Yes, good point.

PR #204, opened 2021-05-06
fix: Clean up jobs which fail to start correctly

evansd:
I think I originally didn't do this to aid debugging. However, it leads
to a cascading failure state where a job fails to start because Docker
runs out of disk space while copying in input files, and then the volume
never gets cleaned up and we can't even extract outputs from completed
jobs to free up space because we can't start the container required to
do the copying.

The one case now when we still *don't* automatically clean up is when a
job completes but hits some unexpected internal error while finalising.
In this case it seems premature to discard the results of potentially
long-running job when often the condition is fixable and the
finalisation step can be successfully retried later.

### DIFF
--- a/jobrunner/run.py
+++ b/jobrunner/run.py
@@ -90,6 +90,7 @@ def handle_pending_job(job):
                 cleanup_job(job)
             except Exception:
                 mark_job_as_failed(job, "Internal error when starting job")
+                cleanup_job(job)
                 raise
             else:
                 mark_job_as_running(job)


ghickman (approved):

jobrunner/run.py
```diff
@@ -90,6 +90,7 @@ def handle_pending_job(job):
                 cleanup_job(job)
             except Exception:
                 mark_job_as_failed(job, "Internal error when starting job")
+                cleanup_job(job)
```
bloodearnest, line 93: Is it worth having some specific log output here to indicate we hit this point?

evansd: "Internal error when starting job" _is_ exactly that log output (all job status messages end up in the log as well as being pushed back to job-server)

evansd:
The PyPI package test failure here is unrelated so I'm going to merge and tackle that separately.

PR #207, opened 2021-05-07
fix: Add expiry to the stata license cache for local_run

bloodearnest:


jobrunner/local_run.py
```diff
@@ -438,9 +439,12 @@ def temporary_stata_workaround(image):
     return True
 
 
+
+
```
ghickman, line 443: did you mean to leave these in?

ghickman (approved):

evansd (commented):
I think that if we have an expired cached value but the fetch fails for any reason (like the user's offline) then we should return the stale value. I think we should _probably_ also bump the cache mtime in this case so we don't retry for another couple of hours, otherwise we'll retry on every run and that could potentially be quite slow, especially if it involves waiting for a timeout (cf https://github.com/opensafely-core/opensafely-cli/issues/37).

bloodearnest:
> _obably_ also bump the cache mtime in this case so we don't retry for another couple of hours, otherwise we'll retry on every run and that could potentially be quite slow, especially if it involves waiting for a timeout (cf [opensafely-core/opensafely-cli#37](https://github.com/opensafely-core/opensafely-cli/issues/37)).

That's a good idea, will do

evansd (approved):
Nice

PR #209, opened 2021-05-10
Attempt to fix git reliability issues

evansd:
This PR contains two changes: one to ensure that our use of git-fetch is atomic and prevent the problems caused by partially fetched commits; and another to expand the class of errors on which we automatically retry fetches.

### DIFF
--- a/jobrunner/git.py
+++ b/jobrunner/git.py
@@ -16,6 +16,9 @@ from .subprocess_utils import subprocess_run
 log = logging.getLogger(__name__)
 
 
+# See `commit_already_fetched`
+SENTINEL_TAG_PREFIX = "fetched/"
+
 
 class GitError(Exception):
     pass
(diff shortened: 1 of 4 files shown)


jobrunner/git.py
```diff
@@ -174,12 +177,54 @@ def ensure_git_init(repo_dir):
...
+    """
+    Return whether a given commit exists in a repo directory
+
+    We used to do this with:
+
+        git cat-file -e 'COMMIT_SHA^{commit}'
+
+    However it's possible that an interrupted fetch leaves the commit object in
+    place without all of its associated blobs, meaning that the above check
+    passes but attempting to checkout the commit will fail. To work around this
```
inglesp, line 189: If we're being pedantic, "checkout" is a noun, and you want "check out".

(cf setup/set up, login/log in, etc... I'm not sure about fuckup/fuck up.)

PR #210, opened 2021-05-10
fix: Apply github proxy when resolving remote refs

evansd:
After all the bikeshedding over configuration I missed the fact that
there's another function which talks to Github which didn't have the
proxy transformation applied! By moving this in with the auth stuff it's
now impossible to accidentally miss this.

### DIFF
--- a/jobrunner/git.py
+++ b/jobrunner/git.py
@@ -128,7 +128,7 @@ def get_sha_from_remote_ref(repo_url, ref):
         output = response.stdout
     except subprocess.SubprocessError as exc:
         redact_token_from_exception(exc)
-        log.exception(f"Error resolving {ref} from {repo_url}")
+        log.exception("Error resolving remote git ref")
         output = ""
     results = _parse_ls_remote_output(output)
     if len(results) == 1:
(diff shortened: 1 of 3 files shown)


jobrunner/git.py
```diff
@@ -264,7 +262,7 @@ def fetch_commit(repo_dir, repo_url, commit_sha, depth=1):
         except subprocess.SubprocessError as e:
             redact_token_from_exception(e)
             log.exception(
-                f"Error fetching commit {commit_sha} from {proxied_url}"
+                f"Error fetching commit {commit_sha} from {repo_url}"
```
bloodearnest, line 265: Hmm, the reason I added proxyied_url was so the url actually used was logged. Now it will say a `github.com` url, which is likely to lead to debugging red herrings if it's an issue with the proxy (which is more likely to break than github.com, I suspect).

Can we still separate out the proxied url here?

I'm +1'ing as it sounds like we've got a broken job-runner, but I think the clarity of logging the correct url that failed in case of error is a better long term behaviour?

bloodearnest (approved):

jobrunner/git.py
evansd: Ah, good point. But I think the solution is actually to pull all of that detail out of the logging message altogether: all the relevant stuff is on the SubprocessError which gets logged alongside it and so duplicating it here only gives the potential for drift.

The GitError message is a slightly different case as that's user-facing and I'm inclined to think of the proxy as an internal implementation detail that we don't want to confuse users with. Maybe worth a chat about?

bloodearnest (approved):

PR #211, opened 2021-05-19
feat: add support for running full job-runner against the dummy backend

bloodearnest:
- allow setting USING_DUMMY_DATA_BACKEND from the environment
- don't validate job request database name if using dummy db
- force job.database_name = "dummy" if we are using dummy db


jobrunner/docker_stats.py
```diff
@@ -47,6 +47,9 @@ def get_container_stats():
 
 
 def _parse_size(size):
-    units = size.lstrip("0123456789.")
-    value = float(size[: -len(units)])
+    units = size.lstrip("0123456789.-")
+    try:
+        value = float(size[: -len(units)])
+    except Exception as e:
+        breakpoint()
```
sebbacon, line 54: o_O

bloodearnest: ah, rats, hadn't pushed latest

jobrunner/config.py
```diff
@@ -43,7 +43,10 @@ class ConfigException(Exception):
 
 BACKEND = os.environ.get("BACKEND", "expectations")
 
-USING_DUMMY_DATA_BACKEND = BACKEND == "expectations"
+if "USING_DUMMY_DATA_BACKEND" in os.environ:
```
evansd, line 46: Just in passing, I generally like to check the env var is at least non-empty, rather than just present. If nothing else it makes it easier to override in development by writing `MYVAR= ./my-executable`.

But I think it's probably better to check the value as well because there's less potential for error e.g.
```py
USING_DUMMY_DATA_BACKEND = {"true": True, "false": False}[
    os.environ.get("USING_DUMMY_DATA_BACKEND", "false").strip()
]
```

jobrunner/create_or_update_jobs.py
```diff
@@ -185,13 +185,15 @@ def recursively_add_jobs(job_request, project, action, force_run_actions):
         if required_job:
             wait_for_job_ids.append(required_job.id)
 
+    database_name = 'dummy' if config.USING_DUMMY_DATA_BACKEND else job_request.database_name
```
evansd, line 188: I think I'd prefer to leave this as it's original value and just not validate it when using dummy data. It doesn't do any harm to leave it as is and I think the fewer places in the code we can fork on `config.USING_DUMMY_DATA_BACKEND` the better.

jobrunner/docker_stats.py
```diff
@@ -47,6 +47,6 @@ def get_container_stats():
 
 
 def _parse_size(size):
-    units = size.lstrip("0123456789.")
+    units = size.lstrip("0123456789.-")
```
evansd, line 50: Wow, Docker! That's not one I was expecting

evansd (approved):
Good to go as it, but I made a couple of minor suggestions if you agree and felt like implementing them.

jobrunner/config.py
bloodearnest: Ok, have done so, but with version that honours Postel's law, which I generally prefer for env var boolean flags

evansd: Ah yeah, good shout

PR #212, opened 2021-05-20
Include full log files in Github Actions output

evansd:
Github Actions supports various [magic strings][1] for formatting
output. In particular, it allows you to hide output behind an expandable
toggle. We exploit this to embed the full log file from each action run
in the test output. We also add some very basic text formatting to
improve readability of the test output.

Example of collapsed output:

github com_opensafely-core_job-runner_runs_2630948131_check_suite_focus=true

Example of expanded output:

github com_opensafely-core_job-runner_runs_2630948131_check_suite_focus=true (1)

Addresses https://github.com/opensafely-core/research-action/issues/6

[1]: https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#grouping-log-lines


ghickman (approved):

bloodearnest:
This is lovely

PR #221, opened 2021-06-22
fix: Slightly better behaviour for cancelled jobs

evansd:
Rather than marking jobs as running and then waiting for the next loop
iteration to actually cancel them, we now immediately invoke
`handle_running_job` to do the cancellation. This gives us the same
structural advantages as we had previously, but it makes cancellation
faster and means we no longer misleadingly report cancelled jobs as
having a non-zero runtime.


ghickman (approved):
Nice!

PR #230, opened 2021-07-09
Test filesystem for case sensitivity

iaindillingham:
Previously, we tested the operating system for case sensitivity. However, this fails on some macOS installs because whilst macOS is POSIX-compatible, by default APFS (Apple File System) is case insensitive. Python's standard library doesn't have a function for testing the filesystem; seemingly the best way is to write/read a temporary file. Finally, because this function is specific to the test, rather than the application code, it's next to the test.

### DIFF
--- a/tests/test_manage_jobs.py
+++ b/tests/test_manage_jobs.py
@@ -1,6 +1,16 @@
 import os
+import tempfile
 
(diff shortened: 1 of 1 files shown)


evansd (approved):
Looks good. You'll just need to give the commit a "semantic" prefix like `ci:`

PR #231, opened 2021-07-12
Use timeouts everywhere when interacting with Docker

evansd:
Shelling out to the Docker client on Windows sometimes hangs
indefinitely (for no obvious reason that we can determine). To prevent
this locking up the entire process we use a large-ish default timeout on
almost all calls to Docker.

The exceptions are copying files and pulling images which we expect to
sometimes take a long time. Pulling images is only done locally at the
moment in any case so we don't need to worry too much about that.
There's no simple way I can think of to distinguish a slow copy of a
large file from a broken call to Docker, so we'll just have to live with
that risk for now.

### DIFF
--- a/jobrunner/docker.py
+++ b/jobrunner/docker.py
@@ -25,6 +25,13 @@ VOLUME_MOUNT_POINT = "/workspace"
 LABEL = "job-runner"
 
 
+# Shelling out to the Docker client on Windows sometimes hangs indefinitely
+# (for no obvious reason that we can determine). To prevent this locking up the
+# entire process we use a large-ish default timeout on almost all calls to
+# Docker. (Timeout value is in seconds.)
+DEFAULT_TIMEOUT = 5 * 60
+
+
 class DockerPullError(Exception):
     pass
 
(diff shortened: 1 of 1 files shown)


jobrunner/docker.py
```diff
@@ -37,22 +44,29 @@ class DockerTimeoutError(Exception):
     pass
 
 
+def docker(docker_args, timeout=DEFAULT_TIMEOUT, **kwargs):
+    args = ["docker"] + docker_args
+    try:
+        return subprocess_run(args, timeout=timeout, **kwargs)
+    except subprocess.TimeoutExpired as e:
+        raise DockerTimeoutError from e
```
ghickman, line 52: Separate to the timeout change, this is a nice little abstraction for the reader, all the calls being explicitly `docker()` calls below is great 👍

jobrunner/docker.py
```diff
@@ -25,6 +25,13 @@
 LABEL = "job-runner"
 
 
+# Shelling out to the Docker client on Windows sometimes hangs indefinitely
+# (for no obvious reason that we can determine). To prevent this locking up the
+# entire process we use a large-ish default timeout on almost all calls to
+# Docker.
+DEFAULT_TIMEOUT = 5 * 60
```
ghickman, line 32: I'm guessing this is seconds?  The subprocess.run docs aren't very clear here, could you add a comment to be explicit about it?

ghickman (approved):
Looks great, just a small tweak to a comment I think?

jobrunner/docker.py
evansd: Good call, thanks.

jobrunner/docker.py
evansd: Yeah, wish I'd done this in the first place and then adding the timeout would have been a lot easier!

PR #234, opened 2021-07-19
feat: Handle reusable actions

iaindillingham:
Reusable Actions, Part Deux 🎉 

The best place to start is *jobrunner/project.py*. Here, `handle_reusable_action` rewrites actions that resolve to reusable actions. It's called within job-runner's sync loop, before a transaction is opened on the database.

A small change within *jobrunner/manage_jobs.py* to `create_and_populate_volume` ensures that reusable actions are handled correctly within job-runner's run loop.

Next steps...

- [ ] Address any changes requested following this PR
- [ ] Add a reusable action to the integration test (*tests/test_integration.py*)

This PR follows #232, which was closed after much squashing and force pushing.

### DIFF
--- a/jobrunner/config.py
+++ b/jobrunner/config.py
@@ -109,6 +109,9 @@ STATA_LICENSE_REPO = os.environ.get(
 )
 
 
+ACTIONS_GITHUB_ORG = "opensafely-actions"
+ACTIONS_GITHUB_ORG_URL = f"https://github.com/{ACTIONS_GITHUB_ORG}"
+
 ALLOWED_GITHUB_ORGS = (
     os.environ.get("ALLOWED_GITHUB_ORGS", "opensafely").strip().split(",")
 )
--- a/jobrunner/create_or_update_jobs.py
+++ b/jobrunner/create_or_update_jobs.py
@@ -91,7 +91,13 @@ def create_jobs(job_request):
     validate_job_request(job_request)
     # In future I expect the job-server to only ever supply commits and so this
     # branch resolution will be redundant
-    if not job_request.commit:
+
+    # job_request.commit will be:
+    # 1. None when in local run mode
+    # 2. None when not in local run mode and job-server didn't supply a commit
+    # 3. A non-empty string when not in local run mode and job-server supplied a commit
+    # Are we 2?
+    if not job_request.commit and not config.LOCAL_RUN_MODE:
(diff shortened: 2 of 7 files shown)


evansd (commented):
Generally this looks great, but (apologies if I'm missing something) I can't see how the repo URL and commit for a resuable action ever makes it on to Job and therefore how the action code ever gets executed.

iaindillingham:
You're not missing anything. I am 🙁.

iaindillingham:
Instances of `ActionSpecifiction` now have `repo_url` and `commit` attributes, which are `None` for non-reusable actions and strings for reusable actions. (Using `None` isn't ideal, but saves passing values into `get_action_specification` within `recursively_add_jobs`.) These values make it onto the job in `recursively_add_jobs`:

https://github.com/opensafely-core/job-runner/blob/99c9310387113470674eb46383fc426f2d510921/jobrunner/create_or_update_jobs.py#L202-L221

iaindillingham:
Something I'm now unsure about:

https://github.com/opensafely-core/job-runner/blob/99c9310387113470674eb46383fc426f2d510921/jobrunner/project.py#L392-L398

The call to `parse_and_validate_project_file` will mutate the project file on first run. It's called in the sync loop and in the run loop, but when the job isn't associated with a reusable action. Whilst this isn't great modelling, I don't _think_ it's a problem.

evansd (approved):
Great work! Looks good

> The call to parse_and_validate_project_file will mutate the project file on first run. It's called in the sync loop and in the run loop, but when the job isn't associated with a reusable action. Whilst this isn't great modelling, I don't think it's a problem.

I see what you mean here and agree with your assessment: structurally it's not great, but it's not actually a problem in practice. We can think about ways to refactor things later.

Before using this on a real project we'll need to fix the "recording the correct repo and commit in manifest.json" problem. But that doesn't need to be tackled now.

iaindillingham:
I think I'll need to rework this to add a reusable action to the integration test (*tests/test_integration.py*). Updating the project (study) fixture and creating a reusable action fixture is easy enough; the issue is how to get `jobrunner.project.handle_reusable_action` to validate the reusable action's repository URL (call to `validate_repo_url`) and get an SHA from a remote reference (call to `git.get_sha_from_remote_ref`) when the URL must contain `github.com` but the remote reference must be a local path.

PR #235, opened 2021-07-20
fix: Support specifying dummy data file in project.yaml

inglesp:


evansd (approved):
This looks good and follows the pattern used elsewhere in supplying additional arguments to cohortextractor.

I'm not 100% happy with that pattern, but that's a separate wider question that isn't directly relevant to that PR. I wonder if it would be better to allow cohortextractor to accept arguments related to dummy data which it just ignores when extracting real data. That way they'd just go as extra args on the `run` line and we reduce the amount that job-runner needs to know about cohortextractor's internals.

PR #240, opened 2021-07-26
fix: Fix non-deterministic integration test

evansd:
The addition of the `generate_cohort_with_dummy_data` action made the dependency graph no longer have a single root. Combined with the job order randomisation that we apply this made it non-deterministic which job was started first which made the test intermittently fail.

This PR disables job order randomisation during the test and slightly refactors how the behaviour of the main `run` loop is configured.

It also re-organises the test to restore the single graph root and hopefully make things clearer.

### DIFF
--- a/jobrunner/config.py
+++ b/jobrunner/config.py
@@ -93,6 +93,12 @@ if PRESTO_TLS_CERT_PATH:
 
 MAX_WORKERS = int(os.environ.get("MAX_WORKERS") or max(cpu_count() - 1, 1))
 
+# This is a crude mechanism for preventing a single large JobRequest with lots
+# of associated Jobs from hogging all the resources. We want this configurable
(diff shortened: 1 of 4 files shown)


sebbacon (approved):
LGTM. Is there any value in also making the dependency unwinding itself deterministic? Just thinking we've fixed the test here but future tests might hit the same problem.

evansd:
Ha! Hilariously, it wasn't until you asked this question that I remembered that we deliberately made it random as a crude way to prevent resource hogging by a single large job request. But there's a switch to disable this and make it deterministic again which we use in local run mode. The right fix here would have been just to enable this in test too. I'll do that now. I think the other changes are overall positive for the shape of the test though so I'll leave them in place.

sebbacon (approved):

PR #241, opened 2021-07-29
Test config

iaindillingham:
I wrote these to fix a bug that turned out to be elsewhere. However, I think they're useful.


evansd (approved):
Looks good. Thanks!

PR #242, opened 2021-07-30
Add debug logging to `run` loop

evansd:
This adds support for a `DEBUG_LOG_FILE` environment variable to which we log at level DEBUG while leaving the primary logs at their default log level. This is necessary because DEBUG logging is extremely noisy and we want to avoid polluting the primary logs.

We also add extra `log.debug()` calls around critical points in the run loop to try to determine where it is locking up.

### DIFF
--- a/jobrunner/log_utils.py
+++ b/jobrunner/log_utils.py
@@ -6,7 +6,6 @@ output. It also includes the stderr output from any failed attempts to shell
 out to external processes.
 """
 import contextlib
-import datetime
 import logging
 import os
 import subprocess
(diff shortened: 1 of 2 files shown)


jobrunner/log_utils.py
```diff
@@ -18,46 +17,36 @@
...
+
+    log_level = os.environ.get("LOGLEVEL", "INFO")
+    handlers = [handler]
+
+    # Support a separate log file at level DEBUG, while leaving the default
+    # logs untouched. DEBUG logging can be extremely noisy and so we want a way
+    # to capture these that doesn't pollute the primary logs.
+    debug_log_file = os.environ.get("DEBUG_LOG_FILE")
+    if debug_log_file:
+        debug_handler = logging.FileHandler(debug_log_file, encoding="utf-8")
```
bloodearnest, line 36: Maybe add a simple limits to avoid accidentally leaving this on and filling up disk

I've done this kind of thing before: https://github.com/canonical-ols/talisker/blob/master/talisker/logs.py#L156

jobrunner/log_utils.py
```diff
@@ -18,46 +17,36 @@
...
+        debug_handler.setFormatter(formatter)
+        debug_handler.addFilter(formatting_filter)
+        debug_handler.setLevel("DEBUG")
+        handlers.append(debug_handler)
+        # Set the default handler to the originally specified log level and
+        # then increase the base log level to DEBUG
+        handler.setLevel(log_level)
+        log_level = "DEBUG"
```
bloodearnest, line 44: Maybe do an INFO log saying sth like "enabled debug logging to f{debug_log_file}" so it's visible in the regular logs

bloodearnest (approved):
I'm a big fan of this approach, having used it a lot previously in talisker.

evansd:
Thanks @bloodearnest, both good suggestions. I've made it keep two weeks worth of daily log files, which should give us plenty of scope for historical debugging while keeping an upper bound on the size. (We've got loads of space on the native disk so I'm not too worried about keeping it small, just bounded.)

PR #244, opened 2021-08-03
Validate image names for reusable actions

evansd:
This ensures that reusable actions have valid images names, and also prohibits them from invoking data extracts. In theory we could allow these, but we know of no valid use case for that at present and limiting the powers of reusable actions seems sensible.

This PR also moves some common validation code into its own module to avoid circular import problems, and splits out the git-related and non-git-related parts of the "handle" function so we can test independently without needing to mock.

### DIFF
--- a/jobrunner/project.py
+++ b/jobrunner/project.py
@@ -211,7 +211,15 @@ def apply_reusable_action(action_id, action, reusable_action):
         # developer didn't make an error; the reusable action developer did.
         action_config = parse_yaml_file(reusable_action.action_file)
         assert "run" in action_config
-    except (ProjectYAMLError, AssertionError):
+        action_run_args = shlex.split(action_config["run"])
+        action_image, action_tag = action_run_args[0].split(":")
+        if action_image not in config.ALLOWED_IMAGES:
+            raise ProjectValidationError(f"Unrecognised runtime: {action_image}")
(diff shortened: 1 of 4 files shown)


jobrunner/project.py
```diff
@@ -211,7 +211,15 @@ def apply_reusable_action(action_id, action, reusable_action):
...
-    except (ProjectYAMLError, AssertionError):
+        action_run_args = shlex.split(action_config["run"])
+        action_image, action_tag = action_run_args[0].split(":")
+        if action_image not in config.ALLOWED_IMAGES:
+            raise ProjectValidationError(f"Unrecognised runtime: {action_image}")
+        if is_generate_cohort_command(action_run_args):
+            raise ProjectValidationError(
+                "Re-usable actions cannot invoke cohortextractor"
+            )
+    except (ProjectYAMLError, AssertionError, ProjectValidationError):
```
inglesp, line 222: It feels a bit odd to be raising one of these errors and then immediately converting to a `ReusableActionError` and throwing away any useful information.  Is that worth addressing now?

inglesp (approved):

evansd:
Thanks @inglesp. I agree with both of your comments, but I've probably already gone slightly too far with what was intended to be a simple extra validation condition so I minded to leave these for another time.

PR #248, opened 2021-08-06
Dereference annotated tags

iaindillingham:
Previously, if `get_sha_from_remote_ref` was passed a ref that resolved to [an annotated tag](http://git-scm.com/book/en/v2/Git-Basics-Tagging), then it would return the SHA of the annotated tag and not of the associated commit. This was a pain, because we tend to assume that SHAs are commit SHAs and not tag SHAs; for example, when checking that a branch ref is an ancestor of a commit ref.

This PR ensures that if `get_sha_from_remote_ref` is passed a ref that resolves to an annotated tag, then it returns the SHA of the associated commit.

### DIFF
--- a/jobrunner/git.py
+++ b/jobrunner/git.py
@@ -3,15 +3,14 @@ Utility functions for interacting with git
 """
 import logging
 import os
-from pathlib import Path, PurePath
 import subprocess
 import time
+from pathlib import Path, PurePath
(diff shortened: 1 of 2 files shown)


evansd (approved):
This is both more powerful and more readable than the original, which is always a nice combination!

PR #250, opened 2021-08-11
Pull initial docker images

bloodearnest:
- Change the default docker registry to be the proxy.
- Pull docker images if they are not present
- Fix stata license test and ensure it is not skipped in CI

There are no real tests for this specific change, due to the difficulty of
requiring certain local docker state, but I manually deleted cohortextractor
and ran the integration tests, which correctly pulled the image.


bloodearnest:
This is old and has been addressed in other ways.

PR #251, opened 2021-08-11
Refactor job creation to remove db access

evansd:
This refactors the job creation and dependency resolution logic so it is no longer intertwined with database access and so it no longer needs to run within a transaction. As well as making the code clearer (I think) this means we can now do expensive operations like pulling from Docker or fetching from Github after we've worked out what we're going to run but before we written anything to the database or opened a transaction.

### DIFF
--- a/jobrunner/create_or_update_jobs.py
+++ b/jobrunner/create_or_update_jobs.py
@@ -19,12 +19,7 @@ from .database import (
     transaction,
     update_where,
 )
-from .git import (
-    GitError,
-    GitFileNotFoundError,
-    get_sha_from_remote_ref,
-    read_file_from_repo,
-)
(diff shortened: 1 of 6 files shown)


inglesp (approved):
A definite improvement!

I'm not clear on what the exact transaction guarantees are though.  For instance, in `create_jobs_with_project_file`, `get_active_jobs_for_workspace` is not in the transaction block.  This is presumably because its outputs are required for `get_jobs_to_run`, which shouldn't be in the transaction.

I'm not sure whether this matters!  And if it does, I'm not sure how to provide consistency guarantees across multiple databases (the database, the filesystem, git).

evansd:
@inglesp You're absolutely right to ask these questions. In this particular case it all works out as safe but I should explain why this is in comments as it's not immediately clear.

PR #257, opened 2021-08-12
Record reusable actions in manifest

### LINKED ISSUE #236: Fix manifest.json for reusable actions
Reusable actions (opensafely-core/job-runner#234) are run by, in effect, rewriting *project.yaml*. However, this isn't reflected in *manifest.json*: the repo and commit point to the study's repo and commit and not to the reusable action's repo and commit.

inglesp:
THIS REQUIRES A DATABASE MIGRATION, SO ANY DEPLOY NEEDS BABYSITTING.

There are no existing tests for what gets logged and stored in manifest.json, so I haven't added any new tests here.

Fixes #236.

### DIFF
--- a/jobrunner/manage_jobs.py
+++ b/jobrunner/manage_jobs.py
@@ -410,6 +410,8 @@ def write_log_file(job, job_metadata, filename):
             "state",
             "commit",
             "docker_image_id",
+            "action_repo_url",
+            "action_commit",
             "exit_code",
             "job_id",
             "run_by_user",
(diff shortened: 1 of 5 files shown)


jobrunner/manage_jobs.py
```diff
@@ -134,7 +134,9 @@ def create_and_populate_volume(job):
     if config.LOCAL_RUN_MODE and not job.commit:
```
evansd, line 149: I think this now needs to be:
```py
if config.LOCAL_RUN_MODE and not job.commit and not job.action_commit
```
Otherwise running re-usable actions locally won't work.

evansd (changes requested):
I think this looks good apart from the one change I mentioned.

I think it would make a lot of sense to merge this soon if we can and then for me to babysit the migration. It's either that or do it when I'm not around at all, or wait until the end of the month.

evansd (approved):

PR #259, opened 2021-08-19
Further refactoring of `create_job`

evansd:
This refactors the job creation process to do all disk access upfront in
a single batch, as we now already do for database access.

This means that the bulk of the job creation logic is now in a pure
function which takes a list of current jobs and returns a list of new
jobs.

As well as being a general structural improvement this paves the way for
support something other than local disk as the primary data store.

### DIFF
--- a/jobrunner/create_or_update_jobs.py
+++ b/jobrunner/create_or_update_jobs.py
@@ -70,11 +70,8 @@ def create_or_update_jobs(job_request):
     else:
         if job_request.cancelled_actions:
             log.debug("Cancelling actions: %s", job_request.cancelled_actions)
-            update_where(
-                Job,
-                {"cancelled": True},
-                job_request_id=job_request.id,
-                action__in=job_request.cancelled_actions,
+            set_cancelled_flag_for_actions(
+                job_request.id, job_request.cancelled_actions
             )
         else:
             log.debug("Ignoring already processed JobRequest")
@@ -82,6 +79,15 @@ def create_or_update_jobs(job_request):
 
 def related_jobs_exist(job_request):
     return exists_where(Job, job_request_id=job_request.id)
+
+
+def set_cancelled_flag_for_actions(job_request_id, actions):
+    update_where(
+        Job,
+        {"cancelled": True},
+        job_request_id=job_request_id,
+        action__in=actions,
+    )
 
 
 def create_jobs(job_request):
--- a/jobrunner/create_or_update_jobs.py
+++ b/jobrunner/create_or_update_jobs.py
@@ -5,6 +5,7 @@ It handles all logic connected with creating or updating Jobs in response to
 JobRequests. This includes fetching the code with git, validating the project
 and doing the necessary dependency resolution.
 """
+import dataclasses
 import logging
 import re
 import time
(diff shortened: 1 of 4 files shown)


jobrunner/create_or_update_jobs.py
```diff
@@ -107,11 +113,11 @@ def create_jobs(job_request):
...
 
-    active_jobs = get_active_jobs_for_workpace(job_request.workspace)
-    new_jobs = get_jobs_to_run(job_request, project, active_jobs)
+    current_jobs = get_latest_job_for_each_action(job_request.workspace)
+    new_jobs = get_jobs_to_run(job_request, project, current_jobs)
 
     if not new_jobs:
-        if active_jobs:
+        if any(job.state in [State.PENDING, State.RUNNING] for job in current_jobs):
             raise JobRequestError("All requested actions were already scheduled to run")
```
inglesp: Is this error message correct?  It talks about "all requested actions", but the condition uses `any()`.

tests/test_create_or_update_jobs.py
```diff
@@ -210,3 +211,9 @@ def make_job_request(action="generate_cohort", **kwargs):
     for key, value in kwargs.items():
         setattr(job_request, key, value)
     return job_request
+
+
+def create_jobs_with_project_file(job_request, project_file):
+    with mock.patch("jobrunner.create_or_update_jobs.get_project_file") as f:
```
inglesp, line 217: Is mocking unavoidable here?

inglesp (approved):
Looks sensible, have added a couple of small notes.

A meta point: this is at the limit of what I can comfortably review without spending significantly more time getting to grips with the codebase.  All the trees in this PR look like they're healthy, but I've not got a good enough view of the overall health of the arboreal ecosystem.  (I'm not sure what to suggest here!)

tests/test_service.py
```diff
@@ -10,8 +10,11 @@
 @pytest.mark.skipif(
     platform.system() == "Windows", reason="tricky to do ctrl-c in windows"
 )
-def test_service_main():
-    p = subprocess.Popen([sys.executable, "-m", "jobrunner.service"])
+def test_service_main(tmp_path):
+    p = subprocess.Popen(
+        [sys.executable, "-m", "jobrunner.service"],
+        env={"WORK_DIR": str(tmp_path), "JOB_SERVER_ENDPOINT": "https://example.com"},
+    )
```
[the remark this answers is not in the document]
evansd: This is a slightly odd test in that it does't actually matter whether the requests succeed or not here. I've tried to add some comments to clarify and switched to using an unrouted IP block rather than example.com, which hangs rather than returning an error.

jobrunner/create_or_update_jobs.py
evansd: It is correct, but I agree that the logic isn't particularly clear here partly, I think, because we're trying to mirror behaviour which was implemented a long time ago and could probably be made more sensible. I've tried to restructure this to make it a bit less confusing though.

tests/test_create_or_update_jobs.py
evansd: It might be, but I'm not quite sure. The reason I used it here was that it was the easiest way of getting the existing tests working without doing some hard thinking and rewriting them, which I didn't fancy doing right then.

evansd:
Thanks @inglesp. As to your meta point I totally appreciate that. Part of the problem is that the full arboreal ecosystem lives in no one's head but mine at the moment. But hopefully that should change over the course of the next couple of months.

For now, I'm very grateful for any sanity/comprehensibility checking, even if that's at a local level.

PR #266, opened 2021-09-02
Remove need for LOCAL_RUN_MODE flag

evansd:
Rather than having production and local_run share a single codepath peppered with switches on `config.LOCAL_RUN_MODE` we provide a separate `create_jobs` implementation for local_run, and use common functions to share common behaviour.

This simplifies the code overall, and makes it easier to further customise behaviour between the two execution modes (e.g. to support automatically pulling new docker images in production, but only warning about available updates in local_run).

### DIFF
--- a/jobrunner/create_or_update_jobs.py
+++ b/jobrunner/create_or_update_jobs.py
@@ -9,7 +9,6 @@ import dataclasses
 import logging
 import re
 import time
-from pathlib import Path
 
 from jobrunner import config
 from jobrunner.lib.database import (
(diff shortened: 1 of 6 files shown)


jobrunner/create_or_update_jobs.py
```diff
@@ -90,6 +90,10 @@ def create_or_update_jobs(job_request):
 
 
 def create_jobs(job_request):
+    # NOTE: Similar but non-identical logic is implemented for running jobs
+    # locally in `jobrunner.cli.local_run.create_job_request_and_jobs`. If you
+    # make changes below then consider what the appropriate corresponding
+    # changes are for locally run jobs.
```
inglesp, line 95: Should we have a corresponding comment in the corresponding function?

jobrunner/manage_jobs.py
```diff
@@ -149,16 +149,17 @@ def create_and_populate_volume(job):
...
-    else:
-        # We're either not in local run mode or we're running a reusable action, so we
-        # need to copy a git commit to a volume.
-        repo_url = job.action_repo_url or job.repo_url
-        commit = job.action_commit or job.commit
+
+    # Jobs which are running reusable actions pull their code from the reusable
+    # action repo, all other jobs pull their code from the study repo
+    repo_url = job.action_repo_url or job.repo_url
+    commit = job.action_commit or job.commit
```
inglesp, line 156: Is there any way we could end up with a commit from the wrong repo?  (that is, looking up a commit from the action repo in the workspace repo, or vice versa)

inglesp (approved):

jobrunner/manage_jobs.py
evansd: Not as things currently stand, but it's good to be defensive. I'll add an assert.

PR #268, opened 2021-09-03
Move reusable actions code into its own module

evansd:
This PR moves all the code related to reusable actions into its own module. This restores `project.py` to being a pure parsing module which doesn't do anything I/O related. The reusable actions code is instead invoked via an explicit call during the job creation process.

This has the side-effect that we only need to call out to Github for the actions that are actually being run, rather than needing to do so for every reusable action in the project. In fact, this is closer to @iaindillingham's original design for the code but it could only be implemented once we'd refactored the job creation process to no longer require a database transaction.

The PR initially moves the reusable actions code verbatim, but later commits apply some refactoring to the exception handling to try to provide some "what do I do now?" hints to the user.


iaindillingham:
I'm not going to be able to do this PR justice this afternoon, so will return to it on Monday. However, I've checked out the branch, started job-runner locally, and added a job: [`test-reusable-actions`](https://github.com/opensafely-actions/test-reusable-actions). (I've also modified `jobrunner.lib.github_validators.validate_repo_url` to return without complaining about repository URLs.)

When doing this from `master`, the job completes. However, not so when doing this from `refactor-reusable-actions` -- both the `test-safetab` and `test-cohort-report` actions, which call the `safetab` and `cohort-report` reusable actions respectively, fail.

At first glance, they seem to fail because they received arguments from the `config` section of *project.yaml* that they didn't recognize.

evansd:
Thanks @iaindillingham. As discussed in Slack I've added a test for this issue and a fix in the final two commits of the PR.

iaindillingham:
Success! The `test-reusable-actions` job completes successfully.

iaindillingham (approved):

PR #270, opened 2021-09-06
fix: Remove default timeout for copies out of Docker volume

evansd:
Large files may legimately take a long time to copy.  We had already
removed this from copies _into_ the volume but neglected to do it for
copies out.

### DIFF
--- a/jobrunner/lib/docker.py
+++ b/jobrunner/lib/docker.py
@@ -158,7 +158,14 @@ def copy_to_volume(volume_name, source, dest, timeout=None):
     )
 
 
-def copy_from_volume(volume_name, source, dest):
+def copy_from_volume(volume_name, source, dest, timeout=None):
+    """
+    Copy the contents of `source` from the root of the named volume to `dest`
+    on local disk
+
+    As this command can potentially take a long time with large files it does
+    not, by default, have any timeout.
+    """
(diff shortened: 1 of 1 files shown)


CarolineMorton (approved):
Looks good to me

bloodearnest (approved):

CarolineMorton (commented):
Sure this is a silly question but `timeout=None` is not used in the function, but is added in as an argument. Is this intentional?

jobrunner/lib/docker.py
```diff
@@ -158,7 +158,14 @@ def copy_to_volume(volume_name, source, dest, timeout=None):
     )
 
 
-def copy_from_volume(volume_name, source, dest):
+def copy_from_volume(volume_name, source, dest, timeout=None):
```
ghickman, line 161: Do you need to use the timeout somewhere?

ghickman (changes requested):

evansd:
Thanks @ghickman and @CarolineMorton for pointing out that I hadn't actually done the one thing I said I did in the PR

jobrunner/lib/docker.py
evansd: I do, George. I really do.

ghickman (approved):

PR #273, opened 2021-09-13
job-executor API spike

benbc:


jobrunner/local_docker_job_executor.py
```diff
@@ -0,0 +1,451 @@
...
+        medium_privacy_dir = get_medium_privacy_workspace(workspace)
+        if medium_privacy_dir:
+            copy_file(
+                workspace_dir / METADATA_DIR / f"{action}.log",
+                medium_privacy_dir / METADATA_DIR / f"{action}.log",
+            )
+            for filename, privacy_level in outputs.items():
+                if privacy_level == "moderately_sensitive":
+                    copy_file(workspace_dir / filename, medium_privacy_dir / filename)
+        cleanup_job(job_id)
```
evansd, line 142: This makes the function non-idempotent. We can't safely clean up until the caller has persisted its state, in whatever form it does that.

Unless I'm missing something, the only safe way to do this is to have an additional API call with the semantics of "make job garbage-collectable". Whether that's done via an inline operation, or just updating some metadata on the job or whatever it may be, is up the executor. But it needs to be the job-runner that tells it that it's OK to do so.

evansd (commented):
Thanks for this @benbc. It's really helpful to have something concrete to look at it.

Initial comments/questions below.

jobrunner/job_executor.py
```diff
@@ -0,0 +1,23 @@
...
+
+def run_job(job_id, image, args, workspace, input_files, env, repo_url, commit, allow_network_access):
+    pass
+
+
+def terminate_job(job_id):
+    pass
+
+
+def get_job_status(job_id, workspace, action, output_spec):
```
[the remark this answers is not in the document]
benbc: Yes, agreed. I'm actually working up to suggesting that the answer is neither.

jobrunner/job_executor.py
```diff
@@ -0,0 +1,23 @@
...
+    # return state, status_code, status_message, outputs, unmatched_outputs
```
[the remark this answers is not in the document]
benbc: Yup, agreed. And maybe also therefore not have to pass output_spec.

jobrunner/job_executor.py
```diff
@@ -0,0 +1,23 @@
+from enum import Enum
+
+
+class Privacy(Enum):
+    HIGH = "high"
+    MEDIUM = "medium"
+
+
+def run_job(job_id, image, args, workspace, input_files, env, repo_url, commit, allow_network_access):
```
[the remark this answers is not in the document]
benbc: Oh, wow. That's surprising. I think we'll have to discuss this.

jobrunner/local_docker_job_executor.py
benbc: Ooh interesting. Thanks. I'll have a think about it.

jobrunner/job_executor.py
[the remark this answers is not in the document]
evansd: Absolutely. It's sometimes hard to remember what was discussed as a group and what was just @bloodearnest and me!

jobrunner/local_docker_job_executor.py
benbc: Ugh. Distributed systems are hard.

jobrunner/job_executor.py
```diff
@@ -0,0 +1,85 @@
...
+from jobrunner.models import State, StatusCode
+
+
+class Privacy(Enum):
+    HIGH = "high"
+    MEDIUM = "medium"
+
+
+InputSpec = Mapping[str, str]  # file paths to actions that produced them
+OutputSpec = Mapping[str, Mapping[str, str]]  # privacy levels to names to patterns
```
evansd, line 14: I think we can simplify `InputSpec` to just a list of files (unless I'm missing some reason we need the action too). And `OutputSpec` can just be a mapping of patterns to privacy levels.

But maybe you're just trying to capture the types of the current status quo before refactoring?

benbc: Yes, you're right twice. :-) The actions from the input spec are logged at the moment, but I think we can live without that.

jobrunner/job_executor.py
```diff
@@ -0,0 +1,114 @@
...
+        contact and authenticate with the database should be provided as environment variables.
+
+        The job must be run with a workspace directory at /workspace in the filesystem (this is expected to be a
+        volume mounted into the container, but other implementations are permitted). The workspace must contain a
+        checkout of the study and any inputs specified in the job definition, copied from long-term storage.
+
+        When a job finishes, whether or not it has succeeded, any output files that match the output spec in the
+        definition must be copied up to long-term storage, both to the high privacy workspace and, if called for in
+        the spec, to the medium privacy workspace. A non-disclosive log file containing useful metadata about the job
+        should also be written to both workspaces in long-term storage.
```
evansd, line 65: This doesn't sound quite right to me. I think we want to say something like:

> Anything written to stdout/stderr by the job should be captured and written to a file named `metadata/{action_name}.log` in both the high and medium privacy workspaces. (Ideally, with each line prefixed by timestamps.)

There's a separate point here which is useful for debugging and audit but I suppose doesn't strictly form part of the contract. That's writing the above log file and the "big bag of useful metadata" into a separate log directory in persistent storage, named using the job_id. I'm not sure where these kind of nice-to-haves are best documented. Or maybe this is more than just a nice-to-have, I'm not sure.

bloodearnest: I'd lean towards more than just nice to have, to be honest, as I think debugging anything will be nigh on impossible without it

benbc: > named using the job_id

I don't see this anywhere. Looks like the filename has the action?

benbc: Oh no, I've found it now. Thanks.

evansd: Yes, it's not done in the most transparent way!

The log and metadata dump happens here:
https://github.com/opensafely-core/job-runner/blob/2214002be42ac478a8111a42901cd179cb4de65c/jobrunner/manage_jobs.py#L296-L300

And the directory it's dumped into uses the container name, which contains the job_id (along with some human readable stuff):
https://github.com/opensafely-core/job-runner/blob/2214002be42ac478a8111a42901cd179cb4de65c/jobrunner/manage_jobs.py#L481-L484

I'm not suggesting that specific scheme needs to be replicated. But some way of getting logs and metadata for historical jobs is very useful, and they're small enough that we can effectively keep them forever.

benbc: @evansd I've rewritten this. I'm trying to make it both clear and correct. I've also taken into account the discussion we had about the relationship between the worksapce and the privacy levels in that call just now.

(The explanation here isn't intended to be completely standalone. I assume there will be a separate doc, covered in another card, which explains the workspace concept and how long-term storage and privacy levels work.)

Could you take another look, please?

benbc: I think I'm exactly mirroring the current implementation apart from not specifying the appending of a subset of the metadata to the log file, which feels a bit fiddly and may be optional. But I can add it too if you think we really need it.

evansd: This looks good to me. I agree about not appending anything to the logs. That's too fiddly and I think we can do everything we need using the entrypoint script in a much cleaner way.

jobrunner/job_executor.py
```diff
@@ -0,0 +1,135 @@
...
+class JobDefinition:
+    id: str  # a unique identifier for the job
+    study: Study  # the study defining the action for this job
+    workspace: str  # the workspace to run the job in
+    action: str  # the name of the action that the job is running
+    image: str  # the Docker image to run
+    args: List[str]  # the arguments to pass to the Docker container
+    env: Mapping[str, str]  # the environment variables to set for the Docker container
+    inputs: List[str]  # the files that the job requires
+    output_spec: Mapping[str, str]  # the files that the job should produce (globs mapped to privacy levels)
```
evansd, line 29: Drive-by thought (not sure of the best place to capture it). Can we/should we specify that output patterns are supplied as some dialect of regular expression (e.g POSIX or POSIX-extended) rather than "globs"? That gives us a reasonable amount of flexibility in our matching logic, and it avoids the problem that the interpretation of glob-like syntax is under-specified (e.g do wildcards match slashes or not?)

PR #274, opened 2021-09-21
chore: docker refactor

bloodearnest:
This is based on the work with the release-hatch docker image, and
provides:

 - lighter and more secure docker prod image, w/o any build
   dependencies.
 - docker dev/test image workflow, which is used in parallel in CI, in
   order to test the docker image.
 - running jobrunner as non-root user inside docker.
 - faster iterative local docker builds.

The `Dockerfile` and `docker-compose.yaml` files are liberally commented
to aid future travellers.

A few implementation notes:

1. We use a hardcoded uid of 10000, which matches the uid we use for the
   jobrunner on our linux hosts, so that file permissions on our mounted
   volumes match. The dev image uses your local uid/gid, so it should be
   seamless in dev.

2. We access docker over SSH, rather than mounting the socket, to
   avoid issues with file permissions around the docker group now that
   we are running as non-root. This requires some local dev setup, which
   is handled automatically by the Makefile, and can be cleaned up with
   `make docker-clean`.

3. We contain all the docker files and commands inside a `docker/`
   subdirectory to avoid bloating the root project directory, as per
   decision in release-hatch project.

Note: some the Makefile parts a not great, but the anticipation is that we'll replace make with just at somepoint, which will simpliy them


.github/workflows/build_and_publish.yaml
```diff
@@ -66,7 +66,9 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v1
     - name: Build image
-      run: docker build . --file Dockerfile --tag $IMAGE_NAME  --build-arg=pythonversion=3.8.3
+      run: make docker-build
+    - name: Test image
+      run: make docker-test ARGS='-e PRIVATE_REPO_ACCESS_TOKEN'
```
bloodearnest, line 71: Because we use `env_file: ../.env` in the docker-compose.yaml, there is an invalid PRIVATE_REPO_ACCESS_TOKEN set, which causes a couple of tests to fail.

It maybe better to add a `.test.env` file that is used by docker-compose, perhaps, but we would need to generate that and include the PRIVATE_REPO_ACCESS_TOKEN and possibly others.

docker/Dockerfile
```diff
@@ -0,0 +1,159 @@
...
+# ensure fully working base python3 installation
+# see: https://gist.github.com/tiran/2dec9e03c6f901814f6d1e8dad09528e
+# use space efficient utility from base image
+RUN --mount=type=cache,target=/var/cache/apt \
+  /root/docker-apt-install.sh python3 python3-venv python3-pip python3-distutils tzdata ca-certificates
+
+# install any additional system dependencies
+COPY docker/dependencies.txt /tmp/dependencies.txt
+RUN --mount=type=cache,target=/var/cache/apt \
+    /root/docker-apt-install.sh /tmp/dependencies.txt
```
madwort, line 25: as an aside - I'm not really into the overloading of the argument(s) to `docker-apt-install.sh`, in which they can be filenames or package names - it's a little too "magical" for my taste - but that's for the other repo not for this one!

docker/Dockerfile
```diff
@@ -0,0 +1,159 @@
...
+##################################################
+#
+# Base project image
+#
+# Ok, we've built everything we need, build an image with all dependencies but
+# no code.
+#
+# Not including the code at this stage has two benefits:
+#
+# 1) this image only rebuilds when the handlful of files needed to build job-runner-base
```
madwort, line 70: typo `handful`

docker/Dockerfile
```diff
@@ -0,0 +1,159 @@
...
+
+RUN mkdir /app
+WORKDIR /app
+VOLUME /app
+
+# We set command rather than entrypoint, to make it easier to run different
+# things from the cli
+CMD ["/opt/venv/bin/python", "-m", "jobrunner.service"]
+
+# This may not be nessecary, but it probably doesn't hurt
```
madwort, line 99: typo `necessary`

docker/docker-compose.yaml
```diff
@@ -0,0 +1,71 @@
...
+        - USERID=${DOCKER_USERID:-1000}
+        - GROUPID=${DOCKER_GROUPID:-1000}
+    # paths relative to docker-compose.yaml file
+    env_file:
+      - ../.env
+    volumes:
+      - ..:/app
+      - ../workdir:/work_dir
+
+  # test runner service - uses dev-image with a different commnd
```
madwort, line 65: typo `command`

madwort (commented):
couple of typos

docker/Dockerfile
bloodearnest: right, it's a bash script and I didn't want to do proper argument parsing. I'm ok with it, but happy to switch if you prefer (although we'd need to update all the images that use it too)

docker/Makefile
```diff
@@ -0,0 +1,82 @@
...
+docker-exec: CMD ?= bash
+docker-exec: docker-build $(SSH_KEY).authorized
+	docker-compose exec dev $(CMD)
+
+
+.PHONY: docker-clean
+docker-clean:
+	# clean up local ssh config
+	rm -f ssh/known_hosts ssh/id_ed25519*
+	sed -i '/$(SSH_COMMENT)/d' ~/.ssh/authorized_keys
```
madwort, line 82: if I was nitpicking I'd feel /slightly/ more comfortable if this was based on the key rather than the comment, but really this is fine.

docker/Dockerfile
```diff
@@ -0,0 +1,159 @@
...
+COPY requirements.dev.txt /tmp/requirements.dev.txt
+# using cache mount instead
+# hadolint ignore=DL3042
+RUN --mount=type=cache,target=/root/.cache \
+    python -m pip install --requirement /tmp/requirements.dev.txt
+
+# in dev, ensure appuser uid matches host user id
+ARG USERID=1000
+ARG GROUPID=1000
+RUN usermod -u $USERID appuser && groupmod -g $GROUPID appuser
```
madwort, line 156: so - this fails for me on macOS, but release-hatch also fails in the same way - not sure if you want to fix now or update them both later. Failure is 
```
[job-runner-dev 3/3] RUN usermod -u 501 appuser && groupmod -g 20 appuser:
#25 0.231 groupmod: GID '20' already exists
```
which are my user (`501`) and first group in my groups listing, which is `staff` aka `20`. not a blocker for this PR, I think.

madwort (approved):

docker/Dockerfile
madwort: I don't think it's particularly worth changing, mostly mentioned it for kind of calibrating our judgements with each other for future choices

docker/Dockerfile
bloodearnest: Well, that is a pain. I'll add an issue to fix, somehow

PR #276, opened 2021-09-24
feat: initial api support

bloodearnest:
Inital implementation of the spike API
    
This adds an EXECUTION_API feature flag, which switches the main run
loop to us alternate version of `handle_pending_jobs` and
`handle_running_jobs`, which use the API.
   
It adds unit tests for those new functions, using the fact we now have
an API! Fairly basic branch coverage to start with, but more to add.
    
Things this branch does not do:
    
 - implement selection and loading of the API implementation
 - handle manifest state
   
The spike code is mostly unaltered, except for moving the `set_message`
into `sync_job_status`, and adding an additional `update(job)` call.

### DIFF
--- a/jobrunner/job_executor.py
+++ b/jobrunner/job_executor.py
@@ -27,7 +27,9 @@ class JobDefinition:
     args: List[str]  # the arguments to pass to the Docker container
     env: Mapping[str, str]  # the environment variables to set for the Docker container
     inputs: List[str]  # the files that the job requires
-    output_spec: Mapping[str, str]  # the files that the job should produce (globs mapped to privacy levels)
+    output_spec: Mapping[
+        str, str
+    ]  # the files that the job should produce (globs mapped to privacy levels)
     allow_database_access: bool  # whether this job should have access to the database
 
 
@@ -108,7 +110,7 @@ class JobAPI(Protocol):
 
             Raises:
                 JobError: if there is a problem retrieving the status of the job
-         """
+        """
         ...
 
     def cleanup(self, job: JobDefinition) -> None:
@@ -134,8 +136,6 @@ class WorkspaceAPI(Protocol):
         This method must be idempotent; if any of the files specified doesn't exist then it must ignore them.
         """
         ...
-
-
 
 
 class NullJobAPI(JobAPI):
@@ -165,6 +165,3 @@ def get_job_api():
 
 def get_workspace_api():
     return NullWorkspaceAPI()
-
-
-
--- a/jobrunner/job_executor.py
+++ b/jobrunner/job_executor.py
@@ -100,6 +100,8 @@ class JobAPI(Protocol):
         irreversible cleanup which loses information about must be deferred to JobAPI.cleanup() which will only be
         called once the results have been persisted.
 
+        If no matching job can be found, it should raise a JobError exception.
+
         The results must include a list of output files that the job produced which matched its output spec. It
         should also include a list of files that it produced but which did not match the output spec,
         to aid in debugging during study development.
--- a/jobrunner/job_executor.py
+++ b/jobrunner/job_executor.py
@@ -1,6 +1,6 @@
 from dataclasses import dataclass
 from enum import Enum
-from typing import Protocol, Mapping, List, Tuple, Optional
+from typing import Mapping, List, Tuple, Optional
 
 from jobrunner.models import State, StatusCode
 from jobrunner import config
@@ -43,7 +43,7 @@ class JobResults:
     image_id: str
 
 
-class JobAPI(Protocol):
+class JobAPI:
     def run(self, job: JobDefinition) -> None:
         """
         Run a job.
@@ -132,7 +132,7 @@ class JobAPI(Protocol):
         ...
 
 
-class WorkspaceAPI(Protocol):
+class WorkspaceAPI:
     def delete_files(self, workspace: str, privacy: Privacy, paths: [str]):
         """
         Delete files from a workspace.
@@ -158,7 +158,7 @@ class NullJobAPI(JobAPI):
         raise NotImplemented
 
 
-class NullWorkspaceAPI(Protocol):
+class NullWorkspaceAPI:
     def delete_files(self, workspace: str, privacy: Privacy, paths: [str]):
         raise NotImplemented
 
--- a/jobrunner/job_executor.py
+++ b/jobrunner/job_executor.py
@@ -1,0 +1,170 @@
+from dataclasses import dataclass
+from enum import Enum
+from typing import Protocol, Mapping, List, Tuple, Optional
+
+from jobrunner.models import State, StatusCode
+from jobrunner import config
+
+
+class Privacy(Enum):
+    HIGH = "high"
+    MEDIUM = "medium"
+
+
+@dataclass
+class Study:
+    git_repo_url: str
+    commit: str
+
+
+@dataclass
+class JobDefinition:
+    id: str  # a unique identifier for the job
+    study: Study  # the study defining the action for this job
+    workspace: str  # the workspace to run the job in
+    action: str  # the name of the action that the job is running
+    image: str  # the Docker image to run
+    args: List[str]  # the arguments to pass to the Docker container
+    env: Mapping[str, str]  # the environment variables to set for the Docker container
+    inputs: List[str]  # the files that the job requires
+    output_spec: Mapping[str, str]  # the files that the job should produce (globs mapped to privacy levels)
+    allow_database_access: bool  # whether this job should have access to the database
+
+
+@dataclass
+class JobResults:
+    state: State
+    status_code: Optional[StatusCode]
+    status_message: str
+    outputs: Mapping[str, str]
+
+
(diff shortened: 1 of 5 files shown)


jobrunner/job_executor.py
```diff
@@ -0,0 +1,170 @@
...
+        The results must include a list of output files that the job produced which matched its output spec. It
+        should also include a list of files that it produced but which did not match the output spec,
+        to aid in debugging during study development.
+
+            Returns:
+                state (State): the state of the job
+                results (Optional[JobResults]): the results, if the job has finished
+
+            Raises:
+                JobError: if there is a problem retrieving the status of the job
```
benbc, line 110: Could you explicitly add the "or the job does not exist", case?

jobrunner/run.py
```diff
@@ -99,6 +111,44 @@ def handle_pending_job(job):
...
+    else:
+        not_started_reason = get_reason_job_not_started(job)
+        if not_started_reason:
+            set_message(job, not_started_reason, code=StatusCode.WAITING_ON_WORKERS)
+        else:
+            try:
+                set_message(job, "Preparing")
+                api.run(job_to_job_definition(job))
+            except JobError as exception:
+                mark_job_as_failed(job, exception)
```
benbc, line 144: I wonder if we should make a courtesy call to `api.cleanup()` here, since until we do that the implementation has no way of knowing that we won't ask for the status of this job.

jobrunner/run.py
```diff
@@ -99,6 +111,44 @@ def handle_pending_job(job):
...
+            except Exception:
+                mark_job_as_failed(job, "Internal error when starting job")
+                raise
+            else:
+                mark_job_as_running(job)
```
benbc, line 149: My preference would be to put this in the `try` block along with the rest of the happy path.

tests/factories.py
```diff
@@ -0,0 +1,92 @@
...
+        self.jobs_terminated = {}
+        self.jobs_cleaned = {}
+        self.results = {}
+        self.errors = {}
+
+    def add_test_job(self, **kwargs):
+        """Create and track a db job object."""
+        job = job_factory(**kwargs)
+        self.jobs[job.id] = job
+        insert(job)
```
benbc, line 46: What's this needed for?

jobrunner/run.py
```diff
@@ -136,6 +186,97 @@ def handle_running_job(job):
...
+def sync_job_status(job, api):
+    """Query API for job status."""
+    state, results = api.get_status(job_to_job_definition(job))
+
+    if state == State.RUNNING:
+        return True
+    assert state != State.PENDING
+
+    # TODO: implement workspace state tracking
+    #delete_obsolete_files(job, results)
```
benbc, line 265: I suggest that you leave the manifest-dependent stuff in. That gives us the opportunity to implement the local docker version of the API before the manifest stuff is replaced, reaching around behind the scenes to write it to the workspace directory.

benbc (commented):
This looks good. Lovely to have those tests. Some questions and suggestions.

benbc:
@bloodearnest Also I meant to say: I realize that some of my feedback is on code I originally wrote. Thank you for tidying up my mess.

jobrunner/job_executor.py
bloodearnest: Yep, will do

jobrunner/run.py
bloodearnest: Right, that might be useful, will do

jobrunner/run.py
bloodearnest: Yeah, I was just preserving the original function structure, but I agree

jobrunner/run.py
```diff
@@ -136,6 +186,97 @@ def handle_running_job(job):
...
+    for action in job.requires_outputs_from:
+        for filename in list_outputs_from_action(action):
+            input_files.append(filename)
+
+    outputs = {}
+    for privacy_level, named_patterns in job.output_spec.items():
+        for name, pattern in named_patterns.items():
+            outputs[pattern] = privacy_level
+
+    return job_executor.JobDefinition(job.id, study, job.workspace,
```
bloodearnest, line 251: This is another change from the spike - I used `job.id` rather than `job.slug` for the JobDefinition.id

tests/factories.py
bloodearnest: So, `get_reason_job_not_started` ends up querying the db for all running jobs.

I added it at one point to make things work, but then I changed how the db fixtures work, so it may not be needed any more. But there might be other places in the flow where the db is queryed for job data rather than using the provided `job` object.

bloodearnest: Ah, yes, and to check if a dependency is running or failed or what not.

benbc: It seems surprising that it's in the TestAPI. I think that if the tests do need each job to be added to the DB then that should be done explicitly in the test, or in a helper function that saves it and adds it to the TestAPI.

bloodearnest: `add_test_job` is exactly that helper function?

bloodearnest: Ah, I think I see your point now - the factory for job create should do this, rather than the TestAPI. You're right - I'd come to this conclusion in parallel because we need SavedJobRequest state in the db to support the current manifest flow.

benbc: Great.

jobrunner/run.py
evansd: I don't particularly care about this at all, but to my mind the original structure was correct. Exceptions raised by `mark_job_as_running` are from a different layer than those raised by "start_job" and shouldn't be handled by the final `except Exception` block - it wasn't the job the failed here, it was the job-runner.

benbc: I think you're right. I'm afraid the case where the `mark_*()` function raises an exception didn't occur to me.

jobrunner/run.py
```diff
@@ -136,6 +194,117 @@ def handle_running_job(job):
...
+def sync_job_status(job, api):
+    """Query API for job status."""
+    state, results = api.get_status(job_to_job_definition(job))
+
+    if state == State.RUNNING:
+        return True
+    assert state != State.PENDING
+
+    # TODO: implement workspace state tracking
+    # delete_obsolete_files(job, results)
```
benbc, line 284: I think this can go in now that you've added the manifest stuff back, can't it?

jobrunner/job_executor.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+@dataclass
+class JobResults:
+    state: State
+    status_code: Optional[StatusCode]
+    status_message: str
+    outputs: Mapping[str, str]
+    exit_code: int
+    image_id: str
```
bloodearnest, line 43: Note: I added these to be able to write the manifest via API. They may not survive, perhaps.

I think `exit_code` would still be good debugging info, but image_id will hopefully be determined a priori rather than post at some point.

jobrunner/run.py
bloodearnest: I haven't implemented a test WorkspaceAPI, I was focussing on the JobAPI. Feel like it's maybe a separate PR?

benbc: Oh -- that's fine. As long as it's a second PR on this stream of work rather than waiting for the manifest replacement.

tests/factories.py
```diff
@@ -0,0 +1,121 @@
...
+    "run_command": "python myscript.py",
+    "output_spec": {},
+}
+
+
+def job_request_factory(**kwargs):
+    if "id" not in kwargs:
+        kwargs["id"] = base64.b32encode(secrets.token_bytes(10)).decode("ascii").lower()
+
+    values = JOB_REQUEST_DEFAULTS.copy()
```
evansd, line 37: Don't you need a `deepcopy()` here (and with `JOB_DEFAULTS` below) because you've got mutable values in the dicts (`requested_actions`, `output_spec` etc)?

bloodearnest: Hmm, I do, yes, but I don't actually modify them

bloodearnest: fixed

jobrunner/run.py
```diff
@@ -99,6 +118,45 @@ def handle_pending_job(job):
...
+        )
+    else:
+        not_started_reason = get_reason_job_not_started(job)
+        if not_started_reason:
+            set_message(job, not_started_reason, code=StatusCode.WAITING_ON_WORKERS)
+        else:
+            try:
+                set_message(job, "Preparing")
+                api.run(job_to_job_definition(job))
+                mark_job_as_running(job)
```
benbc, line 150: Sorry, I think Dave is right. This suggestion of mine has introduced a bug where we mark jobs as failed because we encountered an error marking them as running.

benbc: This needs to go back into an `else` clause.

benbc (changes requested):

jobrunner/run.py
bloodearnest: Ah, yes, will do

benbc (approved):

PR #277, opened 2021-09-28
Add image id to job record

benbc:
SCHEMA CHANGE -- DEPLOY NEEDS MANUAL INTERVENTION


jobrunner/manage_jobs.py
```diff
@@ -269,6 +269,7 @@ def finalise_job(job):
     container_metadata = get_container_metadata(job)
     outputs, unmatched_patterns = find_matching_outputs(job)
     job.outputs = outputs
+    job.image_id = container_metadata["Id"]
```
evansd, line 272: Ah, confusingly that's the _container_ ID. We want the ID of the image used to run it:
https://github.com/opensafely-core/job-runner/blob/5e421c4164369a14d46eeee333e14895f2988363/jobrunner/manage_jobs.py#L426

evansd (changes requested):
One small change. Plus ideally one of the commits would have a "semantic prefix" to generate a new release, although we don't strictly need that to deploy a new version.

jobrunner/manage_jobs.py
benbc: Doh! Fixed.

evansd (approved):

PR #279, opened 2021-09-28
Remove manifest in favour of database

benbc:


jobrunner/create_or_update_jobs.py
```diff
@@ -173,39 +165,12 @@ def get_latest_job_for_each_action(workspace):
...
-    # reconstruct Job instances to return so instead we create a JobPlaceholder
-    # containing just the two fields we care about in this context: the action
-    # and the final state.
-    return [
-        JobPlaceholder(action=action, state=state)
-        for action, state in get_states_for_actions(workspace).items()
-    ]
+    all_jobs = find_where(Job, workspace=workspace)
+    latest_jobs = []
+    for _, jobs in groupby(sorted(all_jobs, key=attrgetter("action")), key=attrgetter("action")):
```
ghickman, line 170: This might be a little easier to parse if you sorted on the line above the groupby?  The double key= on this line caught me out a bit at first despite being quite familiar with both both functions! (it's a shame there isn't a way to do both given you're using the same key)

bloodearnest: Is this more readable? I'm not sure it is!

```
key = attrgetter("action")
all_jobs.sort(key=key)
for _, jobs in groupby(all_jobs, key=key)
```

jobrunner/run.py
```diff
@@ -20,7 +20,7 @@
     finalise_job,
     job_still_running,
     kill_job,
-    start_job,
+    start_job, list_outputs_from_action,
```
ghickman, line 23: It looks like this repo has isort configured but not enforced by CI, could you pop this on the line before `start_job`?

jobrunner/create_or_update_jobs.py
evansd: I'd be tempted by this:
```py
latest_by_action = {job.action: job for job in sorted(all_jobs, key=attrgetter("created_at")}
return list(latest_by_action.values())
```
But I don't know if that's clearer to anyone except me :smile:

ghickman: I only meant to pull the `sorted(all_jobs, key=attrgetter("action"))` portion out as `key=attrgetter("action")), key=attrgetter("action"))` tripped me up a little bit 🙂

benbc: My inclination (which I considered when originally writing it) is to pull out a helper function that implements the group by operation without requiring it to be pre-sorted.

jobrunner/run.py
benbc: Will do. Thanks.

jobrunner/create_or_update_jobs.py
benbc: I don't like having the sort separate because it's purely an implementation detail of the group by, not something that we actually want to achieve.

jobrunner/manage_jobs.py
```diff
@@ -340,12 +336,25 @@ def finalise_job(job):
...
+        # below).
+        write_manifest_file(
+            medium_privacy_dir, {"repo": job.repo_url, "workspace": job.workspace}
+        )
+
+    # We no longer use the manifest file in the high privacy workspace. Delete any old instances of it so that we're
+    # not left with an out-of-date version sitting around.
+    manifest_file = workspace_dir / METADATA_DIR / MANIFEST_FILE
+    if manifest_file.exists():
+        manifest_file.unlink()
```
bloodearnest, line 355: On windows, `unlink()` has a non-negligable chance of failing, so it might we worth adding exception handling

jobrunner/job_executor.py
```diff
@@ -146,21 +145,21 @@ class NullJobAPI(JobAPI):
     """Null implementation of JobAPI."""
 
     def run(self, job: JobDefinition) -> None:
-        raise NotImplemented
+        raise NotImplementedError
```
bloodearnest, line 148: Yeah, found this in my branch too! Oops.

jobrunner/lib/database.py
```diff
@@ -28,10 +28,9 @@ def insert(item):
     get_connection().execute(sql, encode_field_values(fields, item))
 
 
-def update(item, update_fields=None):
+def update(item):
     assert item.id
-    if update_fields is None:
-        update_fields = [f.name for f in dataclasses.fields(item)]
+    update_fields = [f.name for f in dataclasses.fields(item)]
```
bloodearnest, line 33: Would it be worth excluding 'cancelled' field from this generic update?

It seems to me that only the sync thread can write cancelled, and main thread should always read, that might be a nicer solution that relying on a reset on next sync?

bloodearnest (approved):
A few minor suggestions, but looks good to me

jobrunner/manage_jobs.py
benbc: We don't have any equivalent error handling here: https://github.com/opensafely-core/job-runner/blob/master/jobrunner/manage_jobs.py#L517.

I guess we just let the exception bubble up to the `catch Exception` clause in `run.py`.

jobrunner/lib/database.py
benbc: Good. That's much better, thank you.

I've rewritten the commit to incorporate that improvement to make it easier for you to review the change in its entirety. Nothing else has changed since you reviewed.

https://github.com/opensafely-core/job-runner/pull/279/commits/ba193b015cc47b6e62959bae5389fbf2fec93e14

bloodearnest: Looks good!

jobrunner/manage_jobs.py
bloodearnest: Ok. Given it's not a fatal failure, I figured it was worth suppressing any error, but not that bothered by it.

tests/test_integration.py
```diff
@@ -132,29 +134,6 @@ def test_integration(tmp_work_dir, docker_cleanup, requests_mock, monkeypatch):
     manifest = json.load(manifest_file.open())
     assert manifest["workspace"] == "testing"
     assert manifest["repo"] == str(repo_path)
-    assert set(manifest["actions"]) == set(
```
evansd, line 135: I think we should now be asserting equivalent things about the state of the database, rather than just removing these assertions.

evansd (changes requested):
This looks good. I have one small thing that I think needs to be addressed, and one larger thing that needs thinking about, although maybe we decide to do nothing about it.

The small thing is that I think we still want _locally_ the behaviour whereby deleting an output file causes the system to behave like the action never ran, rather than throwing an error. This should just be a case of adding some extra logic after this line:
https://github.com/opensafely-core/job-runner/blob/cae9624fd738316a53ff92deae0e4c61201f3fc8/jobrunner/cli/local_run.py#L409

Something roughly along the lines of:
```py
current_jobs = [
    job for job in current_jobs if all(project_dir.joinpath(f).exists() for f in job.output_files)
]
```

The larger question is whether we need a migration path from the old system to the new. As it stands, deploying this will cause all local outputs to effectively disappear and need to be regenerated. Possibly that's not too much of a problem, but it seems at best a bit impolite.

More importantly, we should check to what extent _production_ files will be impacted by this. The database has been treated as ephemeral and deleted on multiple occasions during its life. We should be able to check the earliest `created_at` date in the jobs table to get a sense for how many files will be affected.

Both the local and production issues could be solved by a function which reads the manifest and creates any necessary Job entries (and then deletes it). We could run this over all workspaces in production, and then remove it completely once enough time has elapsed that any actively developed projects will have run it locally.

benbc:
@evansd, thanks for the feedback.

I've added the logic you suggest for re-running actions with missing files locally. Rebased into the relevant commit here: https://github.com/opensafely-core/job-runner/pull/279/commits/1772292ef4bacc4a68ecd65df16f44fcf3ef23c1.

benbc:
@evansd I was planning a migration process originally, but decided it wasn't necessary when we reallized that all the information we needed was already in the database. That decision could well be wrong.

I'll check what the oldest job is in the database today, so we have a feel for the impact of not providing a migration on production.

I just don't know what the impact will be locally. Will most researchers even notice if some of their actions re-run? Are any actions slow locally, or are they only slow when running on real-scale data? (I don't necessarily expect you to have answers to those questions, but I think they're worth asking.)

tests/test_integration.py
benbc: Well now. I think I disagree. I think we should have tests that check that the system behaves correctly in terms of knowing what has been run and therefore correctly calculating what needs to be run. But the actual persistence is an implementation detail of that, isn't it?

jobrunner/cli/local_run.py
```diff
@@ -406,14 +406,31 @@ def create_job_request_and_jobs(project_dir, actions, force_run_dependencies):
...
+    latest_jobs = get_latest_job_for_each_action(job_request.workspace)
+
+    # On the server out-of-band deletion of an existing output is considered an error, so we ignore that case when
+    # scheduling and allow jobs with missing dependencies to fail loudly when they are actually run. However for local
+    # running we should allow researchers to delete outputs on disk and automatically rerun the actions that create
+    # if they are needed. So here we check whether any files are missing for completed actions and, if so, treat them
+    # as though they had not been run -- this will automatically trigger a rerun.
+    latest_jobs_with_files_present = filter(
+        all_output_files_present(project_dir), latest_jobs
+    )
```
evansd, line 418: I don't think this will work because `filter()` gives you an iterator not a list and it's being passed to two functions below, so the first will consume it. Wrapping it in a `list()` call will do the trick, though you might decide that with the extra nesting the end result is less clear than a list comprehension would have been.

tests/test_integration.py
evansd: It was mainly the list of files produced that I was concerned with. There's now nothing that asserts that the jobs which ran produced the files we expected them to. (Checking the manifest wasn't _exactly_ asserting that either, but it was better than nothing.) Probably we should assert that each of those listed files now exists in the workspace.

I don't have any principled feelings here, I'm just aware that test coverage is already sub-optimal and I worry about lots of red assert statements.

evansd:
@benbc 

> I'll check what the oldest job is in the database today, so we have a feel for the impact of not providing a migration on production.

It occurs to me we don't need to do this with timestamps. It would be an easy enough script to grab every job ID from every manifest file in every workspace, and check whether every corresponding job exists in the database.

> I just don't know what the impact will be locally. Will most researchers even notice if some of their actions re-run? Are any actions slow locally, or are they only slow when running on real-scale data? 

I don't really know the answer either. I will say that I've definitely seen long-running local jobs. Particularly projects involving measures over long time periods as they generate a lot of files and there's a per-file overhead to copying things in and out of docker. Also sometimes people have to generate quite large amounts of dummy data to have enough to get their models to converge, which can also be quite slow to run. It won't be more than an annoyance for anyone I don't think, but it's possible it might be quite annoying for some.

It's probably worth checking the state of the TPP/EMIS workspaces first. If we have to write migration code for production we may as well use it locally as well. It _seems_ like a fairly small thing to write as well (though I've probably missed some horrible complexity!).

jobrunner/cli/local_run.py
benbc: Ugh. Well spotted -- thank you. The UX for things-you-can-iterate-over in Python is really shot to pieces. `yield` is amazing when it's just what you want, but the knock-on implications like this really make me doubt that it's a good trade-off.

tests/test_integration.py
benbc: Oh well, if that assertion is missing altogether then I absolutely agree we need it. I'll have a look at adding a check for what's actually on disk.

benbc: Done. Thank you.

benbc: https://github.com/opensafely-core/job-runner/pull/279/commits/6c562fc3024d240db6b2ab10b9a8cd93ba0f8ad6

benbc: https://github.com/opensafely-core/job-runner/pull/279/commits/20a17896012b698227c5615c7e89bf69fb71ebd5

benbc:
@evansd I have added migration. Please could you look at this again from [test: Check we're producing the right outputs in the integration test](https://github.com/opensafely-core/job-runner/pull/279/commits/154bb7d3f8ea006c1a0d45a562f603c58f9dff06) onwards?

benbc:
@evansd 

> As far as deployment goes, I guess it's just a case of bringing down job-runner, running the migration and then restarting it, right?

Right.

PR #280, opened 2021-09-29
feat: refactor JobAPI.run() into separate preprare/execute/finalize stages

bloodearnest:
Split the `JobAPI.run()` into separate prepare/execute/finalize steps.

This also makes JobAPI properly async, and splits out the state it's tracking into a separate ExecutorState.

This simplifies and changes a lot of things, so the main loop for the API is quite different from the the current one, so we diverge and split a function call higher up than previously.

### DIFF
--- a/jobrunner/job_executor.py
+++ b/jobrunner/job_executor.py
@@ -47,7 +47,7 @@ class ExecutorState(Enum):
 @dataclass
 class JobStatus:
     state: ExecutorState
-    message: Optional[str]
+    message: Optional[str] = None
 
 
 @dataclass
@@ -67,8 +67,9 @@ class JobAPI:
 
     They should return either:
      - the next state if successful
-     - the current state to indicate back pressure and to retry later
      - the ERROR state with an appropriate message if something has gone wrong.
+     - the current state if is different from the expected state, with a message
+     - the initial state to indicate back pressure and to retry later.
 
     Given the long running nature of jobs, it is an asychronous API, and calls should not block for a more than a few
     seconds.
@@ -84,13 +85,16 @@ class JobAPI:
 
         1. Validate the JobDefinition. If there are errors, return an ERROR state with message.
 
-        2. Check the resources are available to prepare the job. If not, return the UNKNOWN state with an appropriate
+        2. Check the job is currently in UNKNOWN state. IF not return its current state with a message indicated invalid
+           state.
+
+        3. Check the resources are available to prepare the job. If not, return the UNKNOWN state with an appropriate
            message.
 
-        3. Create an ephemeral workspace to use for executing this job. This is expected to be a volume mounted into the
+        4. Create an ephemeral workspace to use for executing this job. This is expected to be a volume mounted into the
            container, but other implementations are allowed.
 
-        4. Launch a prepare task asynchronously. If launched successfully, return the PREPARING state. If not, return an
+        5. Launch a prepare task asynchronously. If launched successfully, return the PREPARING state. If not, return an
            ERROR state with message.
 
         The prepare task must do the following:
@@ -111,13 +115,15 @@ class JobAPI:
         """
         Launch the execution of a job that has been prepared, transitioning from PREPARED to EXECUTING.
 
-        1. Validate the the ephememeral workspace created by prepare for this job exists.  If not, return an ERROR
+        1. Check the job is in the PREPARED state. If not, return its current state with a message.
+
+        2. Validate the the ephememeral workspace created by prepare for this job exists.  If not, return an ERROR
            state with message.
 
-        2. Check there are resources availabe to execute the job. If not, return PREPARED status with an appropriate
+        3. Check there are resources availabe to execute the job. If not, return PREPARED status with an appropriate
            message.
 
-        3. Launch the job execution task asynchronously. If launched successfully, return the EXECUTING state. If not,
+        4. Launch the job execution task asynchronously. If launched successfully, return the EXECUTING state. If not,
            return an ERROR state with message.
 
         The execution task must do the following:
@@ -142,9 +148,11 @@ class JobAPI:
         """
         Launch the finalization of a job, transitioning from EXECUTED to FINALIZING.
 
-        1. Validate that the job's ephemeral workspace exists. If not, return an ERROR state with message.
+        1. Check the job is in the EXECUTED state. If not, return its current state with a message.
 
-        2. Launch the finalize task asynchronously. If launched successfully, return the FINALIZING state. If not,
+        2. Validate that the job's ephemeral workspace exists. If not, return an ERROR state with message.
+
+        3. Launch the finalize task asynchronously. If launched successfully, return the FINALIZING state. If not,
            return an ERROR state with message.
 
         The finalize task should do the following:
--- a/jobrunner/job_executor.py
+++ b/jobrunner/job_executor.py
@@ -34,14 +34,13 @@ class JobDefinition:
 
 
 class ExecutorState(Enum):
-    PENDING = "pending"
     PREPARING = "preparing"
     PREPARED = "prepared"
     EXECUTING = "executing"
     EXECUTED = "executed"
     FINALIZING = "finalizing"
     FINALIZED = "finalized"
-    GONE = "gone"
+    UNKNOWN = "unknown"
     ERROR = "error"
 
 
@@ -81,11 +80,11 @@ class JobAPI:
 
     def prepare(self, job: JobDefinition) -> JobStatus:
         """
-        Launch a prepare task for a job, transitioning from PENDING to PREPARING.
+        Launch a prepare task for a job, transitioning to the initial PREPARING state.
 
         1. Validate the JobDefinition. If there are errors, return an ERROR state with message.
 
-        2. Check the resources are available to prepare the job. If not, return the PENDING state with an appropriate
+        2. Check the resources are available to prepare the job. If not, return the UNKNOWN state with an appropriate
            message.
 
         3. Create an ephemeral workspace to use for executing this job. This is expected to be a volume mounted into the
@@ -180,15 +179,15 @@ class JobAPI:
 
     def cleanup(self, job: JobDefinition) -> JobStatus:
         """
-        Clean up any remaining state for a finished job, transitioning to the GONE state.
+        Clean up any remaining state for a finished job, transitioning to the UNKNOWN state.
 
         1. Initiate the cleanup, do not wait for it to complete.
 
-        2. Return the GONE status.
+        2. Return the UNKNOWN status.
 
         This method must be idempotent; it will be called at least once for every finished job. The implementation
         may defer resource cleanup to this method if necessary in order to correctly implement idempotency of
-        get_status() or get_results(). If the job is unknown, it should still return GONE successfully.
+        get_status() or get_results(). If the job is unknown, it should still return UNKNOWN successfully.
 
         This method will not be called for a job that raises an unexpected exception from JobAPI in order
         to facilitate debugging of unexpected failures. It may therefore be necessary for the backend to provide
@@ -199,7 +198,7 @@ class JobAPI:
         """
         Return the current status of a job.
 
-        1. Check the job is known to the system. If not, return the PENDING state.
+        1. Check the job is known to the system. If not, return the UNKNOWN state.
 
         2. Return the current state of the job from the executors perspective.
 
--- a/jobrunner/job_executor.py
+++ b/jobrunner/job_executor.py
@@ -85,7 +85,7 @@ class JobAPI:
 
         1. Validate the JobDefinition. If there are errors, return an ERROR state with message.
 
-        2. Check the job is currently in UNKNOWN state. IF not return its current state with a message indicated invalid
+        2. Check the job is currently in UNKNOWN state. If not return its current state with a message indicated invalid
            state.
 
         3. Check the resources are available to prepare the job. If not, return the UNKNOWN state with an appropriate
--- a/jobrunner/job_executor.py
+++ b/jobrunner/job_executor.py
@@ -33,26 +33,95 @@ class JobDefinition:
     allow_database_access: bool  # whether this job should have access to the database
 
 
+class ExecutorState(Enum):
+    PENDING = "pending"
+    PREPARING = "preparing"
+    PREPARED = "prepared"
+    EXECUTING = "executing"
+    EXECUTED = "executed"
+    FINALIZING = "finalizing"
+    FINALIZED = "finalized"
+    GONE = "gone"
+    ERROR = "error"
+
+
+@dataclass
+class JobStatus:
+    state: ExecutorState
+    message: Optional[str]
+
+
 @dataclass
 class JobResults:
-    state: State
-    status_code: Optional[StatusCode]
-    status_message: str
-    outputs: Mapping[str, str]
+    outputs: Mapping[str, str]  # mapping of outputs to privacy levels
+    unmatched_patterns: List[str]  # list of patterns that matched no outputs
     exit_code: int
     image_id: str
 
 
 class JobAPI:
-    def run(self, job: JobDefinition) -> None:
-        """
-        Run a job.
+    """
+    API for managing job execution.
 
-        This method must be idempotent; it may be called more than once with the same job_id, in which case only one
-        job should be created. It must also be idempotent in the face of errors; if it throws an exception because
-        job creation has failed due to a transient error it may be called again to retry the operation and this
-        should succeed if possible. (The implementation may provide a configuration option which breaks this
-        idempotency by preserving resources after a failure to aid debugging.)
+    This API is called by the job-runner to manage each job it is tracking. It models the running of a job as a state
+    machine, and the methods below are the transitions between states.
+
+    They should return either:
+     - the next state if successful
+     - the current state to indicate back pressure and to retry later
+     - the ERROR state with an appropriate message if something has gone wrong.
+
+    Given the long running nature of jobs, it is an asychronous API, and calls should not block for a more than a few
+    seconds.
+
+    All the state transition methods (prepare(), execute(), finalize(), terminate(), cleanup()) must be idempotent. If
+    the relevent task they are responsible for is already running for that job, they must not start a new task, and
+    instead return successfully the current state.
+    """
+
+    def prepare(self, job: JobDefinition) -> JobStatus:
+        """
+        Launch a prepare task for a job, transitioning from PENDING to PREPARING.
+
+        1. Validate the JobDefinition. If there are errors, return an ERROR state with message.
+
+        2. Check the resources are available to prepare the job. If not, return the PENDING state with an appropriate
+           message.
+
+        3. Create an ephemeral workspace to use for executing this job. This is expected to be a volume mounted into the
+           container, but other implementations are allowed.
+
+        4. Launch a prepare task asynchronously. If launched successfully, return the PREPARING state. If not, return an
+           ERROR state with message.
+
+        The prepare task must do the following:
+
+          - check out the supplied study repo via the OpenSAFELY github proxy into the ephemeral workspace, erroring if
+            there are any failures.
+          - copying the supplied file inputs from the long-term workspace storage into the ephermal workspace, erroring
+            if there are any missing.
+
+        When the prepare task finishes, the get_status() call should now return PREPARED for this job.
+
+        This method must be idempotent. If called with a job that is already running a prepare task, it must not
+        launch a new task, and simply return succesfully with PREPARING.
+
+        """
+
+    def execute(self, job: JobDefinition) -> JobStatus:
+        """
+        Launch the execution of a job that has been prepared, transitioning from PREPARED to EXECUTING.
+
+        1. Validate the the ephememeral workspace created by prepare for this job exists.  If not, return an ERROR
+           state with message.
+
+        2. Check there are resources availabe to execute the job. If not, return PREPARED status with an appropriate
+           message.
+
+        3. Launch the job execution task asynchronously. If launched successfully, return the EXECUTING state. If not,
+           return an ERROR state with message.
+
+        The execution task must do the following:
 
         The specified image must be run, with the provided arguments and environment variables. The implementation
         may add environment variables to those in the job definition as necessary for the backend.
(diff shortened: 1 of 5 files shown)


jobrunner/job_executor.py
```diff
@@ -33,26 +45,86 @@ class JobDefinition:
...
 
+class ExecutorState(Enum):
+    PREPARING = "preparing"
+    PREPARED = "prepared"
+    EXECUTING = "executing"
+    EXECUTED = "executed"
+    FINALIZING = "finalizing"
+    FINALIZED = "finalized"
+    GONE = "gone"
+    ERROR = "error"
```
benbc, line 56: The text below refers to this as `ERRORED`. I don't feel strongly which we use as long as it's consistent.

jobrunner/job_executor.py
```diff
@@ -33,26 +45,86 @@ class JobDefinition:
...
+
+    This API is called by the job-runner to manage each job it is tracking. It models the running of a job as a state
+    machine,and the methods below are the transitions between states.
+
+    Given the long running nature of jobs, it is an asychronous API, and calls should not block for a more than a few
+    seconds, returning a transition to the next state.
+
+    All the state transition methods (prepare(), execute(), finalize(), terminate(), cleanup()) should be idempotent. If
+    the relevent task they are responsible for is already running for that job, they should not start a new task, and
+    instead return successfully the current state.
```
benbc, line 85: Also I think we need a section explaining that any async errors (but not failure of the job itself) should be reacted to by moving to `ERROR` state (and returning a helpful message when the status is next retrieved).

jobrunner/job_executor.py
```diff
@@ -61,75 +133,104 @@ def run(self, job: JobDefinition) -> None:
...
+        It should initiate the cleanup and return the GONE status. It should not wait for cleanup to complete before
+        returning. 
 
         This method must be idempotent; it will be called at least once for every finished job. The implementation
         may defer resource cleanup to this method if necessary in order to correctly implement idempotency of
-        JobAPI.get_status().
+        get_status() or get_results(). If the job is unknown, it should still return GONE successfully.
 
         This method will not be called for a job that raises an unexpected exception from JobAPI.get_status() (anything
         other than JobError), in order to facilitate debugging of unexpected failures. It may therefore be necessary
```
benbc, line 231: Do we still have `JobError` and if so, what for?

benbc (commented):
Previously I tried to keep to RFC2119 usage of must/should/may to minimize the chance of misunderstanding. You've moved to "should" for a bunch of things that I think are non-negotiable. Is that an intentional change?

bloodearnest:
> Previously I tried to keep to RFC2119 usage of must/should/may to minimize the chance of misunderstanding. You've moved to "should" for a bunch of things that I think are non-negotiable. Is that an intentional change?

No. I'm used to those usage being capitalized, so I missed that you done that at all. I will review and change.

jobrunner/job_executor.py
```diff
@@ -33,26 +33,86 @@ class JobDefinition:
...
+    """
+
+
+    def prepare(self, job: JobDefinition) -> JobStatus:
+        """
+        Launch a prepare task for a job, transitioning from PENDING to PREPARING.
+
+        1. Validate the JobDefinition. If there are errors, return an ERRORED state with message.
+
+        2. Check the resources are available to prepare the job. If not, raise ExecutorBusy.
```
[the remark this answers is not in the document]
bloodearnest: Yes, will fix.

jobrunner/job_executor.py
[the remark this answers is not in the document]
bloodearnest: Good idea, will do.

jobrunner/job_executor.py
```diff
@@ -61,75 +133,104 @@ def run(self, job: JobDefinition) -> None:
         in which case it should be run with a network allowing access to the database and any configuration needed to
         contact and authenticate with the database should be provided as environment variables.
 
-        The job must be run with a workspace directory at /workspace in the filesystem (this is expected to be a
-        volume mounted into the container, but other implementations are allowed). The workspace must contain a
-        checkout of the study and any inputs specified in the job definition, copied from the workspace in long-term
-        storage. If any of the specified inputs is not present in long-term storage then a JobError must be raised with
-        details of the missing file.
+        The job must be run with the ephemeral workspace for this job at /workspace in the filesystem (this is expected
+        to be a volume mounted into the container, but other implementations are allowed).
```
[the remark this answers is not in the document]
bloodearnest: Good catch, will do

jobrunner/job_executor.py
bloodearnest: Good call, will add

jobrunner/job_executor.py
benbc: This still seems inconsistent to me. Several instances of "ERRORED" below.

bloodearnest: no ERRORED for me in the branch?

jobrunner/run.py
```diff
@@ -195,28 +155,142 @@ def handle_running_job(job):
...
+
+    definition = job_to_job_definition(job)
+    old_status = api.get_status(definition)
+
     if job.cancelled:
-        log.info("Cancellation requested, killing job")
-        api.terminate(job_to_job_definition(job))
+        # cancelled is driven by user request, so is handled explicitly first
+        # regardless of executor state.
+        api.terminate(definition)
```
benbc, line 194: Since the API returns a state on every call, I think we should at least log it rather than throwing it away. Maybe a decorator is in order which logs every call and the response. That could go in a separate card rather than going to that effort here.

jobrunner/run.py
```diff
@@ -195,28 +155,142 @@ def handle_running_job(job):
...
+        new_status = api.finalize(definition)
+
+    elif old_status.state == ExecutorState.FINALIZED:
+        # final state - we have we have finished!
+        results = api.get_results(definition)
+        # TODO: implement workspace API
+        # delete_obsolete_files(job, results)
+        save_results(job, results)
+        mark_job_as_completed(job)
+        api.cleanup(definition)
```
benbc, line 236: I think the order is wrong here. We shouldn't mark it as completed until we've triggered cleanup, in case there is an error saving the state to the database.

jobrunner/run.py
```diff
@@ -195,28 +155,142 @@ def handle_running_job(job):
...
-        set_message(job, "Failed")
-        mark_job_as_failed(job, "Internal error when finalising job")
-        # We don't clean up here, to facilitate with debugging this unexpected error
-        raise
+            set_message(job, message)
+
+    elif new_status.state == ExecutorState.ERROR:
+        # all transitions can go straight to error
+        mark_job_as_failed(job, new_status.message)
+        api.cleanup(definition)
```
benbc, line 258: I think the order here needs to be the other way round, as above.

jobrunner/job_executor.py
```diff
@@ -33,26 +33,95 @@ class JobDefinition:
...
+    """
+    API for managing job execution.
+
+    This API is called by the job-runner to manage each job it is tracking. It models the running of a job as a state
+    machine, and the methods below are the transitions between states.
+
+    They should return either:
+     - the next state if successful
+     - the current state to indicate back pressure and to retry later
+     - the ERROR state with an appropriate message if something has gone wrong.
```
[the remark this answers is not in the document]
bloodearnest: That's covered under 'something has gone wrong' I think, but we could call it out more specifically. We have to return a state, and ERROR seems appropriate?

We could also raise an exception, but we're currently not raising any, which I think has made control flow nicer.

jobrunner/run.py
```diff
@@ -195,28 +155,142 @@ def handle_running_job(job):
...
+        # cancelled is driven by user request, so is handled explicitly first
+        # regardless of executor state.
+        api.terminate(definition)
+        api.cleanup(definition)
+        mark_job_as_failed(job, "Cancelled by user", StatusCode.CANCELLED_BY_USER)
+        return
 
-    try:
-        is_running = sync_job_status(job, api)
+    if old_status.state == ExecutorState.PENDING:
```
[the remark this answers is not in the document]
bloodearnest: That's a fair point, I was just trying not to add new states.

It does made me realise that this design does leave us open to open to re-running the same job again once we've called cleanup(). But I don't think that's a) likely or b) particularly problematic.

jobrunner/run.py
```diff
@@ -195,28 +155,142 @@ def handle_running_job(job):
...
+        api.cleanup(definition)
+
+    else:
+        raise Exception(
+            f"unexpected state transition from {old_status.state} to {new_status.state}"
+        )
+
+
+def save_results(job, results):
+    """Example the results of the execution and update the job accordingly."""
```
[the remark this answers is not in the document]
bloodearnest: sigh, yes

jobrunner/run.py
```diff
@@ -195,28 +155,142 @@ def handle_running_job(job):
...
+        # typo
+
+        # TODO:  job.unmatched_outputs = ???
+    else:
+        job.state = State.SUCCEEDED
+        job.status_message = "Completed successfully"
+
+    # two updates is messy, but will do for now
+    update(job, update_fields=["state", "outputs", "unmatched_outputs"])
+    set_message(job, results.status_message, results.status_code)
```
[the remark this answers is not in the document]
bloodearnest: so, set_message contains non-trival timestamp update logic that I wanted to preserve. But perhas we do not need it in this specific case.

jobrunner/run.py
bloodearnest: a state transition log is a good idea.

jobrunner/run.py
bloodearnest: but then if there is a db error, the results will be gone, and we can't recover?

bloodearnest: as in, the next time this job is considered, api.get_results() will presumably fail?

jobrunner/run.py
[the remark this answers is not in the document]
benbc: I don't think it's adding new states. I think it's collapsing two into one.

jobrunner/run.py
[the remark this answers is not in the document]
benbc: Yeah, I don't think we do need it here -- isn't it just to avoid spamming updates while we're polling for something?

jobrunner/run.py
benbc: You're quite right. My mistake.

jobrunner/run.py
benbc: Ditto

jobrunner/run.py
bloodearnest: https://linear.app/opensafely/issue/OS-68/add-state-transition-log

jobrunner/run.py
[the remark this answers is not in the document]
bloodearnest: I think knowing that your incoming branch changes update to update all the things, then I think I can just drop the update() call here? And let set_message do it?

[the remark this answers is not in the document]
bloodearnest: Attempted fix in f585f91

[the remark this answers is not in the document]
benbc: I'm sorry to keep going on about this.

We're not just setting the message here -- we're storing all the results of running the job, including the list of outputs. It seems obscure and maybe fragile to rely on a `set_message()` function to have the side effect of persisting all the changes to the job. I really think it would be clearer to call `update()`.

`set_message()` does two other things. It updates `job.updated_at`, which we probably should do here and make sure we're doing elsewhere, and it has spam-avoidance logic which we don't need here.

jobrunner/run.py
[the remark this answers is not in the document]
bloodearnest: Renaming GONE to UNKNOWN wfm.

jobrunner/job_executor.py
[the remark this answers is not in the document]
bloodearnest: Hmm.

So what should it return? No other state seems to fit, do we need new one? Should it raise instead a non-transition error?

jobrunner/run.py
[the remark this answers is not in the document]
bloodearnest: Ok, I was just trying to reuse the existing log as much as possible, but will change.

jobrunner/job_executor.py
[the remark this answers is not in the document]
benbc: The same state as before, I think. The transition was impossible, so the state is unchanged.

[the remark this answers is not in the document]
benbc: It's almost certainly a coding error in the job-runner if this happens, so it _may_ be appropriate for the calling code to raise an exception, but the API should just shrug and get on with its day.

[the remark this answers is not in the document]
bloodearnest: So, we currently use returning the same state as before as an indication of back pressure. So we need to return something else to indicate transition error, or else some other way to indicate back pressure.

benbc:
> I'm used to that usage being capitalized, so I missed that you done that at all. I will review and change.

I'm happy to captitalize them if that makes it clearer.

jobrunner/job_executor.py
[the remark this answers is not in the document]
benbc: I think we should return the same state as before if the state doesn't change. It could not change for various reasons, which can be explained in the message.

If we need to programmatically react to different reasons then we could include a "code" as well as the "message", as we have with `StatusCode` at the moment.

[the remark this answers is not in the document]
bloodearnest: Right, we do need to respond to it programmatically, and I was hoping to avoid adding an additional code, as it this case is currently the only place it's needed.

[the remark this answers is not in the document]
bloodearnest: An invalid transition attempt feels more like hard error to me.

jobrunner/run.py
```diff
@@ -37,42 +38,47 @@
...
-    # single large job request doesn't hog all the workers. We make this
-    # optional as, when running locally, having jobs run in a predictable order
-    # is preferable
-    if config.RANDOMISE_JOB_ORDER:
-        random.shuffle(active_jobs)
+def handle_jobs(active_jobs):
     for job in active_jobs:
         # `set_log_context` ensures that all log messages triggered anywhere
         # further down the stack will have `job` set on them
         with set_log_context(job=job):
```
benbc, line 77: I think that the loop and the setting of log context are now (still?) identical between API and non-API implementations. Could we pull them up to `main` to remove duplication and reduce the chance of them diverging accidentally?

jobrunner/run.py
```diff
@@ -195,28 +161,161 @@ def handle_running_job(job):
...
-        set_message(job, "Failed")
-        mark_job_as_failed(job, "Internal error when finalising job")
-        # We don't clean up here, to facilitate with debugging this unexpected error
-        raise
+    # handle the simple no change needed states.
+    if old_status.state in STABLE_STATES:
+        # no action needed, simply update job message and timestamp
+        message = old_status.state.value.title()
+        set_message(job, message)
+        return
```
benbc, line 216: I like this. Really simplifies things. Should we do something here to reconcile in the case where we think a job is `PENDING`, but the API gives us on one of these states that imply it's actually `RUNNING`?

jobrunner/run.py
```diff
@@ -195,28 +161,161 @@ def handle_running_job(job):
...
+        # cancelled is driven by user request, so is handled explicitly first
+        # regardless of executor state.
+        api.terminate(definition)
+        api.cleanup(definition)
+        mark_job_as_failed(job, "Cancelled by user", StatusCode.CANCELLED_BY_USER)
+        return
 
-    try:
-        is_running = sync_job_status(job, api)
+    old_status = api.get_status(definition)
```
benbc, line 209: I keep stumbling over this name `old_status`, even though I know it's been like this for ages. Would you consider "initial" or "before" as an alternative to "old"? The latter suggests to me that it's somehow out-of-date (or may be), but actually it's the very latest status.

jobrunner/run.py
```diff
@@ -195,28 +161,161 @@ def handle_running_job(job):
...
+            )
+            return
+
+        if any(state != State.SUCCEEDED for state in awaited_states):
+            set_message(
+                job, "Waiting on dependencies", code=StatusCode.WAITING_ON_DEPENDENCIES
+            )
+            return
+
+        next_state = ExecutorState.PREPARING
```
benbc, line 240: Could we call this "expected", instead? I think this better reflects how we actually use it.

jobrunner/run.py
```diff
@@ -195,28 +161,161 @@ def handle_running_job(job):
...
+    definition = job_to_job_definition(job)
+
     if job.cancelled:
-        log.info("Cancellation requested, killing job")
-        api.terminate(job_to_job_definition(job))
+        # cancelled is driven by user request, so is handled explicitly first
+        # regardless of executor state.
+        api.terminate(definition)
+        api.cleanup(definition)
+        mark_job_as_failed(job, "Cancelled by user", StatusCode.CANCELLED_BY_USER)
```
benbc, line 206: Should we be consistent with the calls to `cleanup()` below and always do that after we've persisted the final transition? I can't absolutely convince myself that the ordering here is _wrong_, but I think there's some merit in consistency.

jobrunner/run.py
```diff
@@ -195,28 +161,161 @@ def handle_running_job(job):
...
+
+    elif new_status.state == ExecutorState.ERROR:
+        # all transitions can go straight to error
+        mark_job_as_failed(job, new_status.message)
+        api.cleanup(definition)
+
+    else:
+        raise InvalidTransition(
+            f"unexpected state transition from {old_status.state} to {new_status.state}: {new_status.message}"
+        )
```
benbc, line 287: Do we need to include the job details here, or is it always guaranteed to be logged by whatever handles this exception?

jobrunner/run.py
```diff
@@ -195,28 +161,161 @@ def handle_running_job(job):
...
+
+
+def save_results(job, results):
+    """Extract the results of the execution and update the job accordingly."""
+
+    # this logic is adapted directly from jobrunner.manage_jobs.finalize_job()
```
benbc, line 293: I ignored this before, but I think this kind of comment is so historically contingent that it'll be irrelevant in a few days. I'd drop it altogether.

benbc (approved):
This is great Simon. Good work. Comments are getting pretty picky now.

jobrunner/run.py
bloodearnest: I thought about this, but we use handle_jobs to do one loop through the jobs in tests:

https://github.com/opensafely-core/job-runner/blob/split-run-api/tests/test_integration.py#L71

If we hoist the loop up to `main()`, then the integration test would need to re-implement the loop logic, so another place for divergence.

Not sure which is the lesser evil, tbh!

jobrunner/run.py
bloodearnest: Yeah, we should add a log statement like in a couple of other places.

jobrunner/run.py
bloodearnest: `initial` works for me. I originally used `current`, which did NOT work.

jobrunner/run.py
bloodearnest: Agreed

jobrunner/run.py
bloodearnest: As long as the exception is logged withing the with context manager, it should have an id. I'll add it explicitly in case it bubbles up beyond for some reason.

jobrunner/run.py
benbc: Oh -- if you want it for the tests then you could hoist to a common `handle_jobs()` that does the looping and is called from `main()` and the tests.

jobrunner/job_executor.py
```diff
@@ -33,26 +33,100 @@ class JobDefinition:
...
+    FINALIZING = "finalizing"
+    FINALIZED = "finalized"
+    UNKNOWN = "unknown"
+    ERROR = "error"
+
+
+@dataclass
+class JobStatus:
+    state: ExecutorState
+    message: Optional[str] = None
```
evansd, line 50: A thing it's worth bearing in mind is some mechanism for signalling specific error states that doesn't rely on parsing a free-form `message` field. In the old system we had the (not very well-named) `status_code` field for that. We could just do this by convention here e.g. format message as `[<error code>] <string>`. No strong opinions (yet!) on the right approach, but I think we'll need something like to build a properly helpful UI.

jobrunner/run.py
bloodearnest: That's how it was originally :D

bloodearnest: Have reverted to the original structure

evansd: You can call `main()` with `exit_callback=lambda _: True` and it will exit after the first iteration -- just to throw another evil into the mix

jobrunner/job_executor.py
```diff
@@ -33,26 +33,100 @@ class JobDefinition:
...
+    """
+
+    def prepare(self, job: JobDefinition) -> JobStatus:
         """
-        Run a job.
+        Launch a prepare task for a job, transitioning to the initial PREPARING state.
+
+        1. Validate the JobDefinition. If there are errors, return an ERROR state with message.
+
+        2. Check the job is currently in UNKNOWN state. IF not return its current state with a message indicated invalid
```
evansd, line 88: The F in "IF" is incorrectly capitalised. Never say my reviews don't add value.

jobrunner/run.py
```diff
@@ -195,28 +164,160 @@ def handle_running_job(job):
...
-        api.cleanup(job_to_job_definition(job))
-    except Exception:
-        set_message(job, "Failed")
-        mark_job_as_failed(job, "Internal error when finalising job")
-        # We don't clean up here, to facilitate with debugging this unexpected error
-        raise
+    # handle the simple no change needed states.
+    if initial_status.state in STABLE_STATES:
+        if job.state == State.PENDING:
+            log.warning("state ereror: got {initial_status.state} for a job we thought was PENDING")
```
evansd, line 217: Typo "ereror"

jobrunner/run.py
```diff
@@ -195,28 +164,160 @@ def handle_running_job(job):
...
+        job.state = State.FAILED
+        job.status_message = "No outputs found matching patterns:\n - {}".format(
+            "\n - ".join(results.unmatched_patterns)
+        )
+        # If the job fails because an output was missing its very useful to
+        # show the user what files were created as often the issue is just a
+        # typo
+
+        # Can we figure these out from job.outputs and project.yaml? Do we do
+        # it here or just in local run?
```
evansd, line 312: Yep, I think we should be able to remove this from the API and handle it as a special case in local run. Haven't quite thought through the details yet though.

evansd (approved):
:+1: This looks really nice, and feels like a validation of all that state transition design work we did upfront.

My comments are just typo fixes and a few things to think about later.

jobrunner/job_executor.py
bloodearnest: So, I went without codes, as the specific issues that the codes denote were able to determined by state logic alone, w/o a code.

And we only currently *need* the code to communicate w/job-server, and I was keen to keep job-server specific things out of the executor, so that we can change them easily.

That said, there may be future issues where the executor returns a state that could imply different things that need a code to determine, but right now there aren't any. If that occurs we can model it either as a new state, or add a code, as which ever is appropriate.  We shouldn't never end up parsing a message. We could add a `code` to the `JobStatus` now, with `# reserved for future use`?

jobrunner/run.py
bloodearnest: Ok, should I leave the comment in, or assume we're definitely going to do that and remove it?

evansd: Might be worth leaving the comment in but changing to "we are planning to remove this ...". Or just leave as is if you want, my comment was just for your info.

jobrunner/job_executor.py
evansd: Ah, you're quite right that all the current ones are things you can infer from other state. I keep confusing "things I was planning to do" with "things I have actually done" :smile: The _intention_ was that we'd have a way of signalling things like "this was on out-of-memory error", "this was an our of disk-space error" etc to job-server so it could display a more helpful error message, perhaps with links to advice on what to do next.

Maybe the best thing to do for now is to keep the API simple and worry about how to extend it when we have a concrete need. It's not like we're setting it in stone forever at this point.

PR #285, opened 2021-10-01
Os 48

chrisyeunggraphnet:
For early review only. 

Things have done:
- image for prepare
- single run method to prepare and execute the job

Things have not done yet:
- finalize
- integrate with JobAPI, WorkspaceAPI


jobrunner/k8s/pre.py
```diff
@@ -0,0 +1,63 @@
...
+    copy_input_files(inputs, job_dir)
+
+
+def git_clone_and_checkout(repo_url, commit_sha, repo_dir, job_dir):
+    """
+    Git clone the repo (if not exists) to the workspace repo directory and then checkout the commit to the job directory
+    """
+    
+    # reuse the same method in the docker flow
+    git.ensure_commit_fetched(repo_dir, repo_url, commit_sha)
```
benbc, line 48: I'm not sure it's helpful in this case to separate this out and keep a separate copy of the repo. You're always going to be working on a clean disk, aren't you? In that case maybe it's simpler to unconditionally fetch what you need into the `job_dir`).

jobrunner/k8s/k8s_runner.py
```diff
@@ -0,0 +1,483 @@
...
+    size = config.K8S_STORAGE_SIZE
+    whitelist = config.K8S_EXECUTION_HOST_WHITELIST
+    
+    ws_pv = convert_k8s_name(workspace_name, "pv")
+    ws_pvc = convert_k8s_name(workspace_name, "pvc")
+    job_pv = convert_k8s_name(opensafely_job_id, "pv")
+    job_pvc = convert_k8s_name(opensafely_job_id, "pvc")
+    
+    create_pv(ws_pv, storage_class, size)
+    create_pv(job_pv, storage_class, size)
```
benbc, line 57: I'm a bit surprised that this uses the same storage class and size for both PVs.

As I understand it, the workspace PV is your implementation of "long-term storage" and it needs to be something that supports `ReadWriteMany` -- presumably Azure Files. But the job PV is only needed for this job, so it can be much less fancy.

And the workspace PV has to fit the outputs of all actions in this workspace, plus their logs; but the job PV only needs to fit outputs from one job.

benbc (commented):
@chrisyeunggraphnet This looks very much as I was expecting. Just a couple of comments.

jobrunner/k8s/pre.py
chrisyeunggraphnet: The `repo_dir` will be the one in the long term storage. I wrote this to simulate the current implementation, which stores it in /workdir/repos.

jobrunner/k8s/k8s_runner.py
chrisyeunggraphnet: Yes, you are right. This will be changed.

jobrunner/config.py
```diff
@@ -60,7 +62,8 @@ class ConfigException(Exception):
     "python",
 }
 
-DOCKER_REGISTRY = "ghcr.io/opensafely-core"
+# DOCKER_REGISTRY = "ghcr.io/opensafely-core"
+DOCKER_REGISTRY = os.environ.get("DOCKER_REGISTRY", "")
```
benbc, line 66: @bloodearnest We'll need to provide this in other environments, won't we?

jobrunner/config.py
```diff
@@ -170,6 +173,25 @@ def parse_job_resource_weights(config_file):
...
+K8S_NAMESPACE = os.environ.get("K8S_NAMESPACE", "opensafely")
+
+# Location of the image of the job runner
+K8S_JOB_RUNNER_IMAGE = os.environ.get("K8S_JOB_RUNNER_IMAGE", "ghcr.io/opensafely-core/job-runner:latest")
+
+# Location of the image of the job runner
+K8S_STORAGE_SIZE = os.environ.get("K8S_STORAGE_SIZE", "20Gi")
+
+# The comma separated list of IP:PORT to be whitelisted by the execution job when `allow_database_access` is True
+K8S_EXECUTION_HOST_WHITELIST = os.environ.get("K8S_EXECUTION_HOST_WHITELIST", "")
```
benbc, line 197: @chrisyeunggraphnet It would be good to make this more clearly backend-specific. @bloodearnest do you think there will be a config specific to the local Docker backend?

jobrunner/config.py
bloodearnest: Yeah, we need this, but the default should be "ghcr.io/opensafely-core" if the env is not defined I think

jobrunner/config.py
bloodearnest: Yes - the HIGH_PRIVACY_STORAGE_BASE and such.

Perhaps we could crudely namespace the env vars? eg. `GRAPHNET_*` and `LOCAL_*`?

jobrunner/config.py
chrisyeunggraphnet: Sure, will make ghcr.io/opensafely-core as default

PR #291, opened 2021-10-11
Os 56

chrisyeunggraphnet:
Finished the implementation of the JobAPI and Workspace API.


chrisyeunggraphnet:
PR updated based on the feedback:
- moved to new package `jobrunner.executors.graphnet`
- separated `prepare` and `finalize` as different image (yet, still using some functions in `git`)
- adding DB URL only when `allow_database_access==True`
- refactored the `k8s_runner` and removed static methods
- updated invalid status message

bloodearnest:
> @bloodearnest For new the `job-runner-tools-graphnet` image, since the `prepare.py` reused some functions in git, I don't really want to just copy and paste them. So I would like to make the new dockerfile include the `lib` and the `config.py`:
> 
> ```
> FROM job-runner-base as job-runner-tools-graphnet
> ...
> COPY /jobrunner/lib/. /app/jobrunner/lib
> COPY /jobrunner/__init__.py /app/jobrunner/__init__.py
> COPY /jobrunner/config.py /app/jobrunner/config.py
> ...
> ```
> 
> How do you think?

I think that's fine for now.

You can base the docker-image `FROM base-python` too, as that should have all you need, and no need to mess with root.

bloodearnest:
> You can base the docker-image `FROM base-python` too, as that should have all you need, and no need to mess with root.

It's entirely possible this won't work, in which case, I'd say land as is, and we can refactor the docker parts later.

PR #303, opened 2021-10-22
Executor state logging

benbc:


jobrunner/cli/local_run.py
```diff
@@ -478,13 +469,25 @@ def filter_log_messages(record):
...
         return False
+
     # We sometimes log caught exceptions for debugging purposes in production,
     # but we don't want to show these to the user when running locally
     if getattr(record, "exc_info", None):
         return False
+
+    # Executor state logging is pretty verbose and unlikely to be useful for local running
+    if record.name == executors.logging.LOGGER_NAME:
+        return False
```
bloodearnest, line 489: Nice, this is actually quite clean in the end I think?

jobrunner/executors/local.py
```diff
@@ -120,6 +120,7 @@ def terminate(self, job):
     def cleanup(self, job):
         cleanup_job(job)
         RESULTS.pop(job.id, None)
+        return JobStatus(ExecutorState.UNKNOWN)
```
bloodearnest, line 123: Is this for consistency, or is it actually useful in some way I've missed?

Not a blocker, just curious.

bloodearnest (approved):

jobrunner/cli/local_run.py
benbc: Yes, especially since we already filter the logs for local running, which I didn't realize before I started. And _especially_ especially because @evansd refactored that stuff while I was away to make it cleaner.

jobrunner/executors/local.py
benbc: It's what the contract says. And I think that's right for consistency -- all the state transition methods return the new state.

benbc: (_And_ it's simpler for the decorator if the implementations fulfill the contract rather than having to cope with `None` at that point sometimes.)

PR #309, opened 2021-11-01
Fixes for accessing GitHub when running locally

evansd:
This PR changes the local_run config to access Github directly rather than pushing traffic through the proxy which is unnecessary here.

It also prevents git from prompting for credentials which it will do whenever an action repo is mistyped (Github treats all missing repo errors as authentication failures to avoid leaking the existence of private repos).

### DIFF
--- a/jobrunner/lib/git.py
+++ b/jobrunner/lib/git.py
@@ -18,6 +18,15 @@ log = logging.getLogger(__name__)
 # See `commit_already_fetched`
 SENTINEL_TAG_PREFIX = "fetched/"
 
+# Prevent git from ever prompting for credentials. Hat tip:
+# https://serverfault.com/a/1054253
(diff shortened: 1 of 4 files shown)


jobrunner/lib/git.py
```diff
@@ -18,6 +18,15 @@
 # See `commit_already_fetched`
 SENTINEL_TAG_PREFIX = "fetched/"
 
+# Prevent git from ever prompting for credentials. Hat tip:
+# https://serverfault.com/a/1054253
```
iaindillingham, line 22: What a wail of frustration!

iaindillingham (approved):

jobrunner/lib/git.py
evansd: I'm just delighted that it was someone else's wail, from which I get to beneft!

PR #312, opened 2021-11-02
fix work dir env

bloodearnest:
- chore: fix docker-compose environment for WORK_DIR
- chore: make docker/Makefile able to run the prod docker image more easily

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -91,6 +91,7 @@ ENV VIRTUAL_ENV=/opt/venv/ PATH="/opt/venv/bin:$PATH"
 RUN mkdir /app
 WORKDIR /app
 VOLUME /app
+VOLUME /workdir
 
 # We set command rather than entrypoint, to make it easier to run different
 # things from the cli
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -36,8 +36,9 @@ docker-build: export COMPOSE_DOCKER_CLI_BUILD=1
 docker-build: export BUILD_DATE=$(date -u +'%y-%m-%dT%H:%M:%SZ')
 docker-build: export GITREF=$(git rev-parse --short HEAD)
 docker-build: ENV ?= dev
+docker-build: ARGS ?= --pull
 docker-build:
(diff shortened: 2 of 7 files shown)


benbc (approved):
How much work would it be to change the directory name and env var to be consistent (that is, both either with or without the underscore)?

bloodearnest:
> How much work would it be to change the directory name and env var to be consistent (that is, both either with or without the underscore)?

It's a 2 character change to change `WORK_DIR=/work_dir`. Job runner doesn't care about whatever WORK_DIR points to, so it's up to what we set it to in the docker-compose file really.

Changing it to `WORKDIR=/workdir` also simple, as we currently don't set WORK_DIR in backend-server or on TPP at all, we just use the default of `$PWD/workdir`.

But changing the default to be `$PWD/work_dir` would be effort, as we'd need to migrate

So, I think we want to change to `WORKDIR=/workdir`, and also support WORK_DIR as an alternative?

benbc:
> I think we want to change to WORKDIR=/workdir

That sounds right.

> and also support WORK_DIR as an alternative?

I'm doubtful of the value of this if there aren't lots of places already using it. Better, I think, to make the effort to get everything straight and consistent now rather than allowing for later divergence and suprises (and a tiny bit of extra complexity handling both).

bloodearnest:
Ok.

Will also change `config.WORK_DIR` to `config.WORKDIR` to match.

benbc (commented):
I think maybe you've missed a `work_dir` in `docker/docker-compose.yaml`?

The second commit is a bit weird now, with the `Makefile` stuff alongside the work dir rename. Maybe just squash the whole thing into one commit?

bloodearnest:
> I think maybe you've missed a `work_dir` in `docker/docker-compose.yaml`?

Ah, yes

> The second commit is a bit weird now, with the `Makefile` stuff alongside the work dir rename. Maybe just squash the whole thing into one commit?

d'oh, will fix

benbc (approved):

PR #313, opened 2021-11-04
fix: Restore the missing started_at field for API execution

benbc:


ghickman (approved):

jobrunner/run.py
```diff
@@ -292,6 +292,7 @@ def handle_job_api(job, api):
         # successful state change to the expected next state
         if new_status.state == ExecutorState.PREPARING:
             job.state = State.RUNNING
+            job.started_at = datetime.datetime.now()
```
evansd: You want `int(time.time())` here

evansd (changes requested):

jobrunner/run.py
evansd: Ha! Beat me to it

evansd (approved):

jobrunner/run.py
benbc: Yes, thanks. Helpfully the tests let me know.

PR #316, opened 2021-11-11
ci: fix local docker dev setup to not use SSH.

bloodearnest:
Using SSH by default was slower, and error prone on different
developer's environments (e.g. they had to have SSH set up).

Instead, we default matching the docker GIDs so we can mount the socket.
This avoids the need for SSH in prod and dev

SSH is still supported however, as it is possibly useful in prod in the
longer term.


evansd (approved):
Rubber stamping as I know @madwort's had an initial look over and we need these changes to unbreak CI and merge in some other fixes that we need reasonably urgently.

It would be good for @madwort to do a more thorough post-hoc review at some point though.

PR #317, opened 2021-11-11
fix: handle docker error reponse more robustly

bloodearnest:
In prod we saw instances subprocess_run returning docker errors with
stderr==None, which blew up the previous error handling, and introduced
all kinds of bugs in the error handling.

This change makes the except clause much more robust when handling the
output of a failed docker command.

### DIFF
--- a/jobrunner/lib/docker.py
+++ b/jobrunner/lib/docker.py
@@ -54,13 +54,16 @@ def docker(docker_args, timeout=DEFAULT_TIMEOUT, **kwargs):
     except subprocess.TimeoutExpired as e:
         raise DockerTimeoutError from e
     except subprocess.CalledProcessError as e:
-        stderr = e.stderr
-        if isinstance(e.stderr, bytes):
-            stderr = stderr.decode("utf8")
+        output = e.stderr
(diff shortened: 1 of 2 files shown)


jobrunner/lib/docker.py
```diff
@@ -54,13 +54,16 @@ def docker(docker_args, timeout=DEFAULT_TIMEOUT, **kwargs):
...
         raise DockerTimeoutError from e
     except subprocess.CalledProcessError as e:
-        stderr = e.stderr
-        if isinstance(e.stderr, bytes):
-            stderr = stderr.decode("utf8")
+        output = e.stderr
+        if output is None:
+            output = e.stdout
+        if isinstance(output, bytes):
+            output = output.decode("utf8")
```
evansd, line 61: If we want to be _super_ robust here we should probably use `.decode("utf8", "ignore")` so that malformed output doesn't cause another exception here.

evansd (approved):

jobrunner/lib/docker.py
bloodearnest: done

PR #322, opened 2021-12-07
fix: fix disk space error detection

bloodearnest:
The errors in production had a `.` on the end of the message, meaning
the endswith clause was not an exact match.


evansd (approved):
:facepalm:

PR #329, opened 2021-12-15
fix: ensure that management containers are running after restart

bloodearnest:


evansd (approved):
Makes sense. Do we want to do the same for job containers too?

bloodearnest:
> Makes sense. Do we want to do the same for job containers too?

Hmm, interesting. Possibly, but I think I'd be more inclined to let the job-runner start it, avoid any race conditions between both docker and job-server trying to start the same container? I supposed it might be useful in the current case on Windows where job-runner doesn't always come up after a reboot.

Here's a related thought - what happens if we restart a job on the container it was previously running on, and it's written some files or other wise 'dirtied' it? Would that be a problem? I assume it would just write over the files? But could error?

evansd:
> Here's a related thought - what happens if we restart a job on the container it was previously running on, and it's written some files or other wise 'dirtied' it? Would that be a problem? I assume it would just write over the files? But could error?

That's a good point. My hunch is that in general it would work without problem, but nothing guarantees that. And it doesn't really buy us much over just failing the job and getting the user to re-run it, apart from saving the time copying files into the volume.

PR #330, opened 2021-12-16
fix: more helpful warning if killed by 137

### LINKED ISSUE #208: Indicate likely out-of-memory errors to user
When a Docker container exits with status 137 (process received SIGKILL) this is usually because it's been nuked by the OOM killer. When this happens we should include something helpful in the log "trailer" which lets users know this, otherwise it's quite confusing.

See e.g. https://github.com/opensafely/documentation/discussions/245

bloodearnest:
Fixes #208


evansd (approved):

evansd (approved):
Nice!

PR #339, opened 2022-01-14
Switch to 3.8 as the base Python version

ghickman:
This sets the minimum version of Python we support to 3.8.

### DIFF
--- a/jobrunner/reusable_actions.py
+++ b/jobrunner/reusable_actions.py
@@ -84,8 +84,7 @@ def handle_reusable_action(run_command):
 
     reusable_action = fetch_reusable_action(image, tag)
     new_run_args = apply_reusable_action(run_args, reusable_action)
-    # When we no longer support Python 3.7 we can just use `shlex.join()` here
-    new_run_command = " ".join(shlex.quote(arg) for arg in new_run_args)
+    new_run_command = shlex.join(new_run_args)
     return new_run_command, reusable_action.repo_url, reusable_action.commit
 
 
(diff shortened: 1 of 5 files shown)


jobrunner/reusable_actions.py
```diff
@@ -84,7 +84,7 @@ def handle_reusable_action(run_command):
 
     reusable_action = fetch_reusable_action(image, tag)
     new_run_args = apply_reusable_action(run_args, reusable_action)
-    # When we no longer support Python 3.7 we can just use `shlex.join()` here
+    # we can't use `shlex.join()` here because it fails on Windows
```
iaindillingham, line 87: That's a pain.

evansd (approved):
Looks good. Thanks for doing this! I think if you grep for `3.8` you'll find some more things that can be dropped/simplified, but we don't have to do these now.

iaindillingham (approved):

jobrunner/reusable_actions.py
evansd: The trackback makes it look like that failure was under 3.7:
```
C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\shutil.py:400: in _rmtree_unsafe
```

Are you sure `shlex.join()` was the culprit here?

ghickman: Well now I'm not 😰

ghickman: Yep, looks like I just messed it up, working in the latest actions run.

PR #343, opened 2022-02-28
fix: handle databuilder's generate_dataset command

ghickman:
We renamed `generate_cohort` to `generate_dataset` in opensafely-core/databuilder#383, this special cases that command in job-runner.


evansd (approved):
:+1: Good for now.

I guess this could all do with some refactoring at some point further down the road.

PR #348, opened 2022-03-07
Test with Data Builder

ghickman:
This adds databuilder to a handful of job-runner's tests where we're using cohortextractor as a meaningful part of the test.  A few of these changes are a little messier than I'd like (the integration ones in particular), because databuilder doesn't support the columns we need to match cohortextractor.  I've made these changes on the presumption that we'll update these tests as we add those columns to databuilder.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,11 +10,12 @@ jobs:
   test-job:
     strategy:
       matrix:
+        os: [ubuntu-latest, windows-2019]
         # Python 3.8 is what we currently support for running cohortextractor
         # locally, and 3.9 is what we required for databuilder so we need to make
         # sure we can run with those
         python: [3.8, 3.9]
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     name: Run test suite
     steps:
     - name: Checkout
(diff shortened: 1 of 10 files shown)


evansd (approved):
Thanks for this George. I think this is fine for now as it gets us coverage of the databuilder paths, but I'm not convinced this is the right approach long-term.

We've effectively got two separate projects (a databuilder one and a cohortextractor one) smooshed together into a single fixture. This results in a project.yaml file with two unconnected graphs, which is a bit weird. I get that this comes from the desire to avoid duplicating the entire fixture, but I think the answer here is to remove _both_ cohortextractor and databuilder from this integration test and to create separate tests for them.

The point of the integration test is to check that the basic tasks of job-runner work, so action dependencies get resolved correctly, actions get run, and input and output files get copied to right place. We can do this all with a fixed csv file and a few python scripts, we don't need to involve cohortextractor or databuilder in this.

We also want to test that we invoke cohortextractor and databuilder correctly, as these are special cased images, but that can be done in separate tests.

Of course, back when we only had cohortextractor the simplest (and laziest!) thing to do was to combine it all into one big test. But logically I think they're separate and the issues you're hitting here are a consequence of the fact that we didn't separate them earlier.

ghickman:
Interesting!  I tried having separate project.yamls for them to start with (it was even messier), but didn't consider pulling the extraction tool from the test 🤔  Would there not be some value in checking these tools are also invoked correctly by job-runner given they're special-cased actions?

evansd:
> Would there not be some value in checking these tools are also invoked correctly by job-runner given they're special-cased actions?

Oh yeah, absolutely. I just mean that they don't need to be part of the pipeline integration test and can be done as simpler, standalone integration tests.

PR #351, opened 2022-03-24
Use just stack

ghickman:


requirements.txt
```diff
@@ -1,18 +0,0 @@
-#
```
evansd, line 1: It looks like the switch from `requirements.txt` to `requirements.prod.txt` has also resulted in a version upgrade. Job-runner's deployment is a bit weird because we can't use Docker (because Windows) and neither can we pip install stuff (because we're in the secure networkless environment). So we've ended up shipping in dependencies via this repo:
https://github.com/opensafely-core/job-runner-dependencies

This means that upgrading stuff isn't completely straightforward and so it would be best to keep to the currently specified versions so we don't risk drift between dev/CI and prod.

Hopefully we can revisit this whole situation and do something better once we're running on a Linux VM.

evansd (commented):
In general this looks great, and thanks for doing all the laborious work here!

I've made one comment below about requirements versions which I think needs addressing.

It would also be good to get @bloodearnest's eyes on this as he's a lot more familiar with the Docker build stuff than I am.

requirements.txt
ghickman: This definitely wasn't intended, thanks for catching it!

ghickman: Updated by copying over the versions from requirements.txt -> requirements.prod.txt.  With the way just is configured we're going to see some changes creep into requirements.prod.txt, for instance `ruamel-yaml-clib` is picked up from `ruamel-yaml` (both have the `.` normalised to `-`).  Is there a reason we don't include that already?

ghickman: I think we can leave this as-is and see how we get on with it then?

evansd: Yep, agreed

ghickman: @evansd – could you add the green tick please 🙂

evansd (approved):
Whoops! Sorry.

Did @bloodearnest get a chance to look at this btw?

docker/Makefile
```diff
@@ -1,6 +1,7 @@
 
 # local dev env file for docker-compose to use when building
 .env:
+	cp ../dotenv-sample ../.env
```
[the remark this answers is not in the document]
ghickman: It's two different .env files, this targets `job-runner/.env`, while the one below targets `job-runner/docker/.env`.  

The top level makefile did the same as this line and made it a dependency for all the `docker-*` targets.  It seemed easier to move that down into this makefile, but I don't feel strongly about it.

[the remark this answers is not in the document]
evansd: Oh lol, ignore me. I can't count my path ascenders.

PR #358, opened 2022-03-29
feat: more robust copy from volume

bloodearnest:
Copy to a tempfile and rename, making the filesystem view atomic.

This is going to be useful for our upcoming migration, but also is
a good idea anyway.

### DIFF
--- a/jobrunner/lib/docker.py
+++ b/jobrunner/lib/docker.py
@@ -4,6 +4,7 @@ Utility functions for interacting with Docker
 import json
 import os
 import re
+import secrets
 import subprocess
 
 from jobrunner import config
@@ -184,16 +185,23 @@ def copy_from_volume(volume_name, source, dest, timeout=None):
     not, by default, have any timeout.
     """
     dest.parent.mkdir(parents=True, exist_ok=True)
-    docker(
-        [
-            "cp",
-            f"{manager_name(volume_name)}:{VOLUME_MOUNT_POINT}/{source}",
-            dest,
-        ],
-        check=True,
-        capture_output=True,
-        timeout=timeout,
-    )
+    tmp = dest.with_suffix(dest.suffix + f".{secrets.token_hex(8)}.tmp")
+    try:
+        docker(
+            [
+                "cp",
+                f"{manager_name(volume_name)}:{VOLUME_MOUNT_POINT}/{source}",
+                tmp,
+            ],
+            check=True,
+            capture_output=True,
+            timeout=timeout,
+        )
+    except Exception:
+        tmp.unlink(missing_ok=True)
+        raise
+    else:
+        tmp.replace(dest)
 
 
 def glob_volume_files(volume_name, glob_patterns):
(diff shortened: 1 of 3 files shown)


tests/lib/test_docker.py
```diff
@@ -58,3 +59,35 @@ def run(args, timeout, **kwargs):
...
+    dst = tmp_path / "dst.txt"
+    tmp = dst.with_suffix(dst.suffix + ".tmp")
+
+    src.write_text("I exist")
+    docker.create_volume(volume)
+    docker.copy_to_volume(volume, src, "src.txt")
+    with pytest.raises(Exception):
+        docker.copy_from_volume(volume, "src.txt", "/dev/full")
+        
+    assert not dst.exists()
```
bloodearnest, line 92: I have realised that this assertions is kinda pointless, as it will never exist as we pass in /dev/null

jobrunner/lib/docker.py
```diff
@@ -184,16 +184,22 @@ def copy_from_volume(volume_name, source, dest, timeout=None):
...
-        [
-            "cp",
-            f"{manager_name(volume_name)}:{VOLUME_MOUNT_POINT}/{source}",
-            dest,
-        ],
-        check=True,
-        capture_output=True,
-        timeout=timeout,
-    )
+    tmp = dest.with_suffix(dest.suffix + ".tmp")
```
evansd, line 187: Given that we're doing this in `lib.docker` (which makes sense as it means we get the behaviour for free everywhere else) I've got a very tiny worry about the fixed `.tmp` suffix. We don't know in what contexts we'll end up using this function and it's possible we'll end up blatting something we don't want to.

What about something super crude like:
```py
tmp = dest.with_suffix(f"{dest.suffix}.{secrets.token_hex(8)}.tmp")
```

evansd (approved):
Looks good. I've made a small paranoid comment

jobrunner/lib/docker.py
bloodearnest: good call, fixed

jobrunner/lib/docker.py
```diff
@@ -184,16 +185,23 @@ def copy_from_volume(volume_name, source, dest, timeout=None):
...
-        [
-            "cp",
-            f"{manager_name(volume_name)}:{VOLUME_MOUNT_POINT}/{source}",
-            dest,
-        ],
-        check=True,
-        capture_output=True,
-        timeout=timeout,
-    )
+    tmp = dest.with_suffix(dest.suffix + ".{secrets.token_hex(8)}.tmp")
```
evansd, line 188: Missing an `f"` here!

bloodearnest: Yeah, just spotted that, fixed

tests/lib/test_docker.py
```diff
@@ -58,3 +59,40 @@ def run(args, timeout, **kwargs):
...
+    volume = __name__
+    src = tmp_path / "src.txt"
+    dst = tmp_path / "dst.txt"
+    tmp = dst.with_suffix(dst.suffix + ".tmp")
+    src.write_text("I exist")
+    docker.create_volume(volume)
+    docker.copy_to_volume(volume, src, "src.txt")
+    docker.copy_from_volume(volume, "src.txt", dst)
+    assert dst.read_text() == "I exist"
+    assert not tmp.exists()
```
evansd, line 75: You'll need to do a `glob` here to check there's nothing matching the pattern `*.tmp` as the temporary filename is no longer completely predictable.

bloodearnest: Fixed

PR #371, opened 2022-04-07
Limit moderately sensitive file sizes

inglesp:
Not ready for merging yet: we need to work out what a sensible limit is.

### DIFF
--- a/jobrunner/executors/local.py
+++ b/jobrunner/executors/local.py
@@ -252,8 +252,21 @@ def finalize_job(job):
         image_id=container_metadata["Image"],
         message=message,
     )
+    output_size_bytes = get_output_size_bytes(job, results.outputs)
+    if output_size_bytes > config.MAX_OUTPUT_SIZE_BYTES:
+        if results.message is None:
+            results.message = f"total size of moderately-sensitive outputs exceeded {config.MAX_OUTPUT_SIZE_BYTES}"
     persist_outputs(job, results.outputs, container_metadata)
     RESULTS[job.id] = results
+
+
+def get_output_size_bytes(job, outputs):
+    volume = volume_name(job)
+    return sum(
+        docker.get_file_size_bytes(volume, filename)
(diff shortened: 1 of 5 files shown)


jobrunner/executors/local.py
```diff
@@ -252,10 +252,23 @@ def finalize_job(job):
...
     )
+    output_size_bytes = get_output_size_bytes(job, results.outputs)
+    if output_size_bytes > config.MAX_OUTPUT_SIZE_BYTES:
+        if results.message is None:
+            results.message = f"total size of moderately-sensitive outputs exceeded {config.MAX_OUTPUT_SIZE_BYTES}"
     persist_outputs(job, results.outputs, container_metadata)
     RESULTS[job.id] = results
 
 
+def get_output_size_bytes(job, outputs):
```
bloodearnest, line 263: Possibly miscommunication here - my original suggestion 32mb limit was per file, not across all moderately_sensitive files for an action?

Some actions produce a lot of images (e.g. measures), and this would prevent them from being released.

We may want a perfile limit and also a total limit, perhaps, but the initial 32mb suggestions was a per-file.

I will analyse some file sizes on TPP and see what they currently look like.

jobrunner/lib/docker.py
```diff
@@ -280,6 +280,25 @@ def find_newer_files(volume_name, reference_file):
     return sorted(files)
 
 
+def get_file_size_bytes(volume_name, path):
```
bloodearnest, line 283: The approach of one docker subshell per file is a bit inefficient. I suspect we could do this with a single docker with `find | xargs` subshell that returns all files and their sizes, then pull out the ones we want in python, perhaps?

evansd: If we're going to extract everything to the high priv workspace anyway (which I think we have to) then it would be easier and more efficient to check the sizes there rather than inside the docker volume.

bloodearnest: Even better

PR #372, opened 2022-04-11
Use Pipeline Library to Parse project.yaml

### LINKED ISSUE #366: Switch to pipeline library
Port the `project.yaml` handling code to our [pipeline library](https://github.com/opensafely-core/pipeline/) so we can standardise on one place to handle `project.yaml`s.

ghickman:
This moves the contents of `project.py` into our [pipeline](https://github.com/opensafely-core/pipeline) library, switching us over to instances of the Pipeline model where the output of that is used here.

I've tried to keep each commit readable so they're likely the best way to read through the changes here.

This is currently installing the `support-job-runner` branch of the pipeline so CI can pass, that will need to be switch to a tag before this can be merged.

Fixes #366


requirements.prod.txt
```diff
@@ -10,11 +10,19 @@ chardet==3.0.4
     # via requests
 idna==2.10
     # via requests
+opensafely-pipeline @ https://github.com/opensafely-core/pipeline/archive/refs/heads/support-job-runner.zip
```
bloodearnest, line 13: Is the plan to merge this into pipeline before landing this?

bloodearnest (approved):

requirements.prod.txt
ghickman: Yes, I'll merge the changes to pipeline, and then update the version here so we're always running from a tag 👍

bloodearnest (approved):

PR #373, opened 2022-04-19
fix: atomic file writing and not following symlinks

bloodearnest:
We previously changed `docker.copy_from_volume()` to write to
a temporary file and rename, to gain basic atomic write behaviour.

This change generalises this and extends it to the non-docker
`copy_file()` method.

It also means we do not follow symlinks when writing to a file, as
`shutil.copy()` and friends do. This is useful in migrating between file
systems

### DIFF
--- a/jobrunner/manage_jobs.py
+++ b/jobrunner/manage_jobs.py
@@ -17,7 +17,7 @@ import time
 from pathlib import Path
 
 from jobrunner import config
-from jobrunner.lib import docker
+from jobrunner.lib import atomic_writer, docker
 from jobrunner.lib.database import find_one
 from jobrunner.lib.git import checkout_commit
 from jobrunner.lib.path_utils import list_dir_with_ignore_patterns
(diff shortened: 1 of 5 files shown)


jobrunner/manage_jobs.py
```diff
@@ -493,11 +493,14 @@ def get_log_dir(job):
...
-    # need to stick with 3.7 for some reason we could replace this with a
-    # shellout to `cp`. See:
-    # https://docs.python.org/3/library/shutil.html#shutil-platform-dependent-efficient-copy-operations
-    shutil.copy(source, dest)
+
+    with atomic_writer(dest) as tmp:
+        # shutil.copy() should be reasonably efficient in Python 3.8+, but if we
+        # need to stick with 3.7 for some reason we could replace this with a
+        # shellout to `cp`. See:
+        # https://docs.python.org/3/library/shutil.html#shutil-platform-dependent-efficient-copy-operations
```
evansd, line 501: This comment can go now. We're fully committed to 3.8+ everywhere.

evansd (approved):
Lovely little context manager. I'm going to be borrowing this is in future.

jobrunner/manage_jobs.py
bloodearnest: It's outa here

PR #374, opened 2022-04-20
fix: bump Black from 22.1.0 to 22.3.0 to avoid click v8 issue

ghickman:


requirements.dev.txt
```diff
@@ -94,7 +94,7 @@ pytest-cov==3.0.0
     # via -r requirements.dev.in
 pyyaml==6.0
     # via pre-commit
-requests==2.27.1
+requests==2.25.0
```
evansd, line 97: There's something a bit funny going on here as we seem to have requests (and it's dependencies) listed twice, both here and in [`requirements.prod.txt`](https://github.com/opensafely-core/job-runner/blob/aa8df60616dace7eb14c241d0bcf8454ff2ceb75/requirements.prod.txt). And this PR makes the versions diverge. I don't know exactly what problems this would cause in practice, but it doesn't seem ideal.

I think the simplest thing might be to just upgrade `black` in the PR and leave the other deps as they are.

evansd (changes requested):

requirements.dev.txt
ghickman: This should have only been an upgrade of black, thanks for catching!

ghickman: I see what's happened.  These non-black changes are from the `--constraint` flag we set in `requirements.dev.in`.  It's brought the versions inline with `requirements.prod.txt`.  I'm going to break those changes out into a separate commit since they're the correct state for `requirements.dev.txt` to be in.

evansd: Ah, I'd totally missed that! In fact, I read the diff backwards because I'm so used to seeing upgrades rather than downgrades. So you're PR actually puts things in the correct state :+1:

evansd (approved):

PR #375, opened 2022-04-22
feat: Add `prepare_for_reboot` command

evansd:
We're going to be rebooting a few times in the coming days and it's
better if this can be mostly transparent to users.


bloodearnest (approved):
This is deceptively straight forward.

PR #377, opened 2022-04-25
Add generate codelist report

lucyb:
Like the generate cohort command, this requires access to the database.
It's a custom command that uses some custom SQL to quickly query the
database for OS Interactive

### DIFF
--- a/jobrunner/project.py
+++ b/jobrunner/project.py
@@ -221,12 +221,6 @@ def get_action_specification(project, action_id):
     elif is_generate_cohort_command(run_args):
         raise RuntimeError("Unhandled cohortextractor version")
 
-    elif is_generate_codelist_report_command(run_args):
-        if not args_include(run_args, "--output-dir"):
-            raise ProjectValidationError(
-                "generate_codelist_report command should specify an output directory:\n"
-            )
-
     return ActionSpecifiction(
         run=run_command,
         needs=action_spec.get("needs", []),
--- a/jobrunner/project.py
+++ b/jobrunner/project.py
@@ -243,7 +243,15 @@ def requires_db_access(args):
     """
     By default actions do not have database access, but certain trusted actions require it
     """
-    return is_generate_cohort_command(args) or is_generate_codelist_report_command(args)
+    return (
+        len(args) > 1
+        and [
+            args[0].startswith(image)
+            for image in ("cohortextractor:", "cohortextractor-v2:", "databuilder:")
+        ]
+        and args[1]
+        in ("generate_cohort", "generate_dataset", "generate_codelist_report")
+    )
 
 
 def is_generate_cohort_command(args, require_version=None):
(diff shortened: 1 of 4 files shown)


evansd (approved):
This all looks good to me. One small question about whether we need the validation or not, but nothing blocking.

bloodearnest (approved):

jobrunner/project.py
```diff
@@ -221,6 +221,12 @@ def get_action_specification(project, action_id):
     elif is_generate_cohort_command(run_args):
         raise RuntimeError("Unhandled cohortextractor version")
 
+    elif is_generate_codelist_report_command(run_args):
+        if not args_include(run_args, "--output-dir"):
+            raise ProjectValidationError(
+                "generate_codelist_report command should specify an output directory:\n"
+            )
+
```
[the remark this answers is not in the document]
lucyb: That's fair. We did start going down a complete rabbit hole and backed out, but left this bit in. It wasn't clear where the responsibility for this lies, but I think removing it makes a lot of sense, so I'll do that.

bloodearnest (commented):
LGTM.

I think allowing `databuilder generate_codelist_report` is probably ok, and will make it Just Work :tm: if we add the command to databuilder in future.

bloodearnest:
> > LGTM.
> > I think allowing `databuilder generate_codelist_report` is probably ok, and will make it Just Work tm if we add the command to databuilder in future.
> 
> I changed it again, sorry!
> 
> Are the mappings correct? Dave seemed to indicate that `databuilder generate_cohort` wasn't valid?

It's a change from previous code, but I thought it was probably harmless. But probably better to be stricter.

jobrunner/project.py
```diff
@@ -239,6 +239,23 @@ def add_config_to_run_command(run_command, config):
...
 
 
+def requires_db_access(args):
+    """
+    By default actions do not have database access, but certain trusted actions require it
+    """
+    valid_commands = {
+        "cohortextractor": ("generate_cohort", "generate_codelist_report"),
+        "cohortextractor-v2": ("generate_cohort", "generate_dataset"),
+        "databuilder": ("generate_dataset"),
```
inglesp, line 249: This looks like a single-element tuple but it isn't!

I think you want:

```
         "databuilder": ("generate_dataset",),
```

jobrunner/project.py
```diff
@@ -239,6 +239,23 @@ def add_config_to_run_command(run_command, config):
...
+    }
+    if len(args) <= 1:
+        return False
+
+    image, command = args[0], args[1]
+    image = image.split(":")[0]
+    return True if command in valid_commands.get(image, []) else False
```
inglesp, line 256: You could just do:

```
     return command in valid_commands.get(image, [])
```

jobrunner/project.py
evansd: Well ingl-spied! Missed by three reviewers (incl. myself) I think. The perennial Python footgun!

lucyb: New PR over here https://github.com/opensafely-core/job-runner/pull/380

PR #388, opened 2022-04-27
Add dependency checking workflows

StevenMaude:
# Overview

This PR adds scanning of dependencies and generation of a software bill of materials (SBOM) via GitHub Actions workflows.

This has been discussed in opensafely-core/security#1. There is some other discussion there around the suitability of various tooling.

If this works well, we intend to rollout similar changes to other repositories. This minimises the impact of:

* having unexpected and inconvenient failures
* encountering unanticipated Actions/workflows quirks

# What features are added?

* GitHub's [official dependency review check](https://github.com/actions/dependency-review-action) action
* Creation of releases from tags
* Anchore's [SBOM generating action](https://github.com/anchore/sbom-action) that uses [Syft](https://github.com/anchore/syft)
* Anchore's [Grype action](https://github.com/anchore/scan-action) to scan existing dependencies for vulnerabilities

# How are these features added?

## Dependency review check

A standard workflow that is run on pull request.

This action uses the GitHub API to check vulnerabilities in changed or added dependencies on pull request. It is relatively new. We may find that it has issues that prevent it from being too useful right now.
    
## Grype scan

A reusable workflow that is:

* run on a schedule and 
* ~run within the "build and publish" workflow~ now removed, see reasoning for decision below.

## SBOM creation and upload

A reusable workflow that is run within the "build and publish" workflow as the final step.

This also creates a release from the new tag so that we can add the SBOM as a release asset.

# Outstanding issues/decisions

## Blocking builds on vulnerability detection

The "build and publish" workflow currently runs the Grype scan first and will fail if the scan finds moderate or severe vulnerabilities.

**Should this block a release?**

### Considerations

For the Grype scan:

* This means someone is forced to address the vulnerabilities if trying to merge their code, which is a positive.
* It means that you may end up dealing with this when trying to deal with some other urgent problem, which could be a distraction.
* If we don't require a passing scan, we can either remove it entirely or set `fail-build` to `false`. We can also make it such that the `tag-new-version` job does not list the scan as a requirement to complete: that is, the scan can run concurrently, instead of blocking as it does now.
* An alternative would be to set the failing vulnerability level to be "high", "critical" or similar, instead of the default "medium". This might be best passed in to the reusable workflow as a configurable option.

The GitHub dependency review check has similar considerations, though it is less restrictive: it only fails if changed dependencies have vulnerabilities. This check does not offer any control of failure conditions, at time of writing.

### Decision

Remove this, because it could prevent changes being merged for reasons unrelated to the branch.

## SBOM not generated for manual releases

The SBOM is not currently generated for releases published manually.

### Considerations

* The `sbom-action` only uploads a release asset if you use the `on: release` trigger.
* An `on: release` type trigger won't work for our `build_and_publish` case. We need to create a release from the tag via `GITHUB_TOKEN` first. But, workflow steps using `GITHUB_TOKEN` don't trigger other workflows, so the `release`-triggered workflow won't then trigger. (You could use a personal access token to work around this restriction, but that's a bit messy.)
  * **So, we have to add the release asset within the existing workflow that is already running.**
* One feature of this is that we could actually add another workflow with `on: release`. This would only get triggered by a manual release! But this takes advantage of a restriction — the use of `GITHUB_TOKEN` — that is possibly not guaranteed to always be there.
  * There may also be some messiness around how to share such a workflow between repositories; not entirely sure here.
* **If we never intend to create releases manually, we could leave the existing behaviour as is, for simplicity.**

### Decision

Leave behaviour as is, for automatically published releases only, because we are not using manual releases.

## Sharing of workflow code

There is some attempt here to pull out the steps that are, or might be, reused into reusable workflows. This would remove duplication if we apply similar changes to multiple repositories.
 
### Considerations

* Both [composite actions and reusable workflows](https://github.blog/2022-02-10-using-reusable-workflows-github-actions/#reusable-workflows-vs-composite-actions) are ways to avoid duplication, but have different behaviour and restrictions.
  * Within the same repository, reusable workflows are particularly simple to use.
  * Outside of that, it is less immediately clear which might be most suitable.
  * Maybe the Dependabot action check picks up updated composite actions too? Not sure about Dependabot interacting with reusable workflows.
* Storing several reusable workflows in one place has issues with versioning: versioning is governed by repository tags. So you can't easily version one workflow within a collection such as stored in a single repository.
* There may be better ways of simplifying the duplication; suggestions welcome.

### Decision

Leave this as is for now, because it makes implementation simpler at this stage. Copy-pasting might be OK for adding to other initial test repositories, following rollout here. At that point, we should assess sharing workflow code.

### Considerations

Here, the latest major checkout action is used. We could consider using the same version through all the workflows here.

### Decision

Fix this separately as it's an unrelated change here.

### DIFF
--- a/.github/workflows/schedule_grype.yaml
+++ b/.github/workflows/schedule_grype.yaml
@@ -2,7 +2,7 @@ name: Grype security scan
 on:
   workflow_dispatch:
   schedule:
-    - cron: '0 10 * * * '
+    - cron: '7 7 * * * '
 
 jobs:
   grype-scan:


.github/workflows/schedule_grype.yaml
```diff
@@ -0,0 +1,12 @@
+name: Grype security scan
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 10 * * * '
```
StevenMaude, line 5: comment: Slightly annoying here is that since I create the `cron` line, I'll get the notifications.

It might be better for someone else to modify this line or, disable and re-enable the workflow, so that they get the notification:

> Notifications for scheduled workflows are sent to the user who initially created the workflow. If a different user updates the cron syntax in the workflow file, subsequent notifications will be sent to that user instead. If a scheduled workflow is disabled and then re-enabled, notifications will be sent to the user who re-enabled the workflow rather than the user who last modified the cron syntax.

https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/notifications-for-workflow-runs

.github/workflows/schedule_grype.yaml
```diff
@@ -0,0 +1,12 @@
...
+
+jobs:
+  grype-scan:
+    name: Scan project for vulnerabilities
+    uses: ./.github/workflows/grype_scan.yaml
```
ghickman, line 10: What's the value of splitting out the grype part of this from the scheduler portion?

StevenMaude: In this case, not that much now. For this example, right now, I *could* move it back into the schedule.

Initially, it was running both on schedule and as part of the build, where it was useful to have this to avoid duplication.

More generally, I was also thinking about whether it's worth having potentially common workflows stored elsewhere, so that there's one central place to maintain them. A reusable workflow or composite action would be one way of doing this.

.github/workflows/schedule_grype.yaml
lucyb: That is annoying. Would be better if it was someone on the pipeline team. Happy for you to assign me to receive the notifications or, for me to disable/enable the workflows after merge, so that I get them

lucyb: Side note, would it make sense to schedule it in the early hours of the morning, rather than 10am so the report is ready when we start work?

lucyb (approved):
This is a really complex topic. I was impressed by how much consideration you'd given to each of the decision points and how you were able to communicate them back to the team.

.github/workflows/schedule_grype.yaml
StevenMaude: Done :heavy_check_mark:

@lucyb: if you, or someone else, could deactivate/reactivate, that would be great, thanks!

StevenMaude:
As I've added this, I'll check back on it in case there are any hiccups.

PR #395, opened 2022-05-04
feat: Return custom messages for known db exit codes

### LINKED ISSUE #184: Send non-disclosive, transient errors back to job server
There are certain errors thrown inside jobs which are non-disclosive which are currently hidden and should be dispalyed to Users on the website.

I think we still want to be very careful with what we expose so maybe some sort of allow list for these?

rebkwok:
cohortextractor now exits with custom exit codes for database errors - see https://github.com/opensafely-core/cohort-extractor/pull/791
This PR is just to report user-friendly error messages when job-runner encounters those exit codes
Fixes #184

### DIFF
--- a/jobrunner/executors/local.py
+++ b/jobrunner/executors/local.py
@@ -241,7 +241,16 @@ def finalize_job(job):
     container_metadata = get_container_metadata(job)
     outputs, unmatched_patterns = find_matching_outputs(job)
     exit_code = container_metadata["State"]["ExitCode"]
-    message = config.EXIT_CODES.get(exit_code)
+
+    # First get the user-friendly message for known database exit codes, for jobs
+    # that have db access
+    message = None
+    if job.allow_database_access:
+        message = config.DATABASE_EXIT_CODES.get(exit_code)
+    # No database error, check for other known exit codes
+    if message is None:
+        message = config.EXIT_CODES.get(exit_code)
+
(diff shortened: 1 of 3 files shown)


jobrunner/executors/local.py
```diff
@@ -241,11 +241,7 @@ def finalize_job(job):
     container_metadata = get_container_metadata(job)
     outputs, unmatched_patterns = find_matching_outputs(job)
     exit_code = container_metadata["State"]["ExitCode"]
-    message = None
-    if exit_code == 137:
-        # 137 = 128+9, which means was killed by signal 9, SIGKILL
-        # This usually happens because of OOM killer, or else manually
-        message = "likely means it ran out of memory"
+    message = config.EXIT_CODES.get(exit_code)
```
evansd, line 244: I think we should treat these exit codes as being specific to database jobs. So if, say, a random R job happens to exit 3 we don't report it as a database error.

Fortunately, I think this should just be a case of checking the `job.allow_database_access` flag. But it does make things a bit more awkward in that we need separate handling for the 137 code. Maybe we just need two separate error code dicts, one for database jobs and one for the rest? Not sure what the cleanest approach is here.

evansd (approved):
I've made one suggestion for an improvement, but it's not blocking as I think we're unlikely to be bitten by this in the near future.

PR #396, opened 2022-05-05
feat: Add a basic KV flag implementation.

bloodearnest:
Stored in a new DB table, flag keys and values are both strings. Comes
with a command line module to list and set and clear flags.

The flag lookup is deliberately robust, and will not fail if the new
table doesn't exist, which we'd expect in local_run situations. Flag
writes will fail if the table doesn't exist, the but the cli can create
it if needed.

In the process, I refactored the SQL schema from an external file into
separate per table strings. This allowed for the ability to easily
add the new table with the cli, but also removes a use of `__FILE__`
from the code base, which is good as we cannot use it with pyoxidizer.


jobrunner/lib/database.py
```diff
@@ -19,6 +19,62 @@
 
 CONNECTION_CACHE = threading.local()
 
+TABLE_SCHEMAS = {
```
evansd, line 22: :+1: Using inline Python rather than an external file is a good move, as is splitting it per table. I feel like this deserves a separate `schema.py` file though, rather than bundling it in `database.py` which has already got quite a lot going on.

jobrunner/lib/database.py
```diff
@@ -104,6 +160,35 @@ def select_values(itemclass, column, **query_params):
     return [decode_field_values(fields, row)[0] for row in cursor]
 
 
+def get_flag(name):
```
evansd, line 163: I'm not sure `get_flag` and `set_flag` belong in here, as this is supposed to be generic, schema-agnostic stuff. I think `queries.py` would be the place to put them.

It might also help consistency of the codebase to use the shitty ORM (despite it being shitty) and have `Flag` be a dataclass in `models.py` but I don't feel quite as strongly about that.

jobrunner/cli/flags.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def run(argv):
+    parser = argparse.ArgumentParser(description=__doc__.partition("\n\n")[0])
+    parser.add_argument(
+        "flags",
+        metavar="FLAG[=VALUE]",
+        nargs="*",
+        type=Flag.from_cli,
+        help="Just flag to view, or flag=value to set, flag= to clear",
```
evansd, line 67: This API makes me a bit nervous. I'd prefer explicit `flags get` and `flags set` commands.

jobrunner/cli/flags.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def do_flags(flags):
+    if not flags:
+        all_flags = get_connection().execute("SELECT name FROM flags")
+        flags = [Flag(f[0], NOVALUE) for f in all_flags]
+    for flag in flags:
+        if flag.value is not NOVALUE:
+            set_flag(flag.name, flag.value)
+        print(f"{flag.name}={get_flag(flag.name)}")
```
evansd, line 57: I don't think this is the best named function and I didn't really understand what it was doing until I saw how the API worked. I think explicitly separating the `get` and `set` operations would end up making this code clearer as well as the API safer, and would mean you could do away with the `NOVALUE` sentinel.

jobrunner/cli/flags.py
```diff
@@ -0,0 +1,80 @@
...
+        do_flags(flags)
+    except sqlite3.OperationalError as e:
+        if "no such table" in str(e):
+            if create:
+                create_table(get_connection(), "flags")
+                do_flags(flags)
+            else:
+                sys.exit(
+                    "Flags table does not exists. Run command again with --create to create it."
+                )
```
evansd, line 45: Nice :+1:  I think it's the right call to make this an explicit action, but explain what to do in the error message.

evansd (changes requested):
Broadly speaking this all looks like the right approach, but I've got a few comments on the specifics below.

Refactoring the way the schema is defined is definitely a good call (although the pedantic side of me would have liked that in a separate commit :smile: ).

jobrunner/lib/database.py
```diff
@@ -104,6 +160,35 @@ def select_values(itemclass, column, **query_params):
...
+    """Get the current running mode in the db."""
+    # Note: fail gracefully if the flags table does not exist
+    # This means we don't need to worry about it in local_run.
+    try:
+        cursor = get_connection().execute(
+            f"SELECT value FROM flags WHERE name = '{name}'"
```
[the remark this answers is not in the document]
bloodearnest: Hmm, I was trying to avoid using the "ORM", as I was thinking that you'd get back a Flag object, which you'd then have to deference .value on on, or similar.

But I guess we could define a `Flag` dataclas, an use the ORM querying, but just return the `Flag.value` from `get_flag()`?

jobrunner/lib/database.py
bloodearnest: OK, will do

jobrunner/cli/flags.py
bloodearnest: How do you imagine it going wrong?

jobrunner/lib/database.py
bloodearnest: What do you think about embedding the schema in the dataclass definition?

e.g.

class Job:
   __tablename__ = "job"
   __tableschema__ = """
       CREATE TABLE job (
       ...
   """

jobrunner/lib/database.py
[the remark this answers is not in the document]
evansd: Yeah, I was thinking that logic could go in `get_flag()` which is already going to have to handle the missing table stuff. I agree it's a bit more bureaucratic, but I worry that if we start talking to the db without going through the ORM layer we're going to make it harder to replace it with something better in future.

jobrunner/lib/database.py
bloodearnest: I did a thing for auto registering.

jobrunner/lib/database.py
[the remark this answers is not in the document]
bloodearnest: Have added an ORM object, and an `upsert` ORM method.

jobrunner/cli/flags.py
bloodearnest: Changed. I'm not entirely happy with it, but it'll do for now

jobrunner/cli/flags.py
bloodearnest: Done. It's a bit cleaner

jobrunner/queries.py
```diff
@@ -24,3 +25,24 @@ def calculate_workspace_state(workspace):
...
     return groupby(sorted(iterable, key=key), key=key)
+
+
+def get_flag(name):
+    """Get the current value of a flag, None if not set."""
+    # Note: fail gracefully if the flags table does not exist
+    # This means we don't need to worry about it in local_run.
+    try:
+        return find_one(Flag, id=name).value
+    except ValueError:
```
evansd, line 36: I should have made `find_one` raise something more specific here, something like `DoesNotExist` and `MultipleObjectsReturned` like Django does. Then we could catch exactly that error. I guess given the primary key constraint on ID we're never going to get multiple records back here, but I worry that `ValueError` is still quite broad. If you feel like adding another exception class then great, but I don't think essential.

jobrunner/cli/flags.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def main(action, flags, create=False):
+    if action == "set":
+        flags = set_flags(flags, create)
+    elif not flags:
+        try:
+            flags = select_values(Flag, "id")
+        except sqlite3.OperationalError:
+            pass
```
evansd, line 52: `flags` won't be set so you'll get an error further down when you try to iterate it. Is is worth printing something like "flags database table does not exist" and exiting? Or could just set `flags = []` as the simplest fix.

evansd (approved):
:+1: Looks really good. A couple of small comments.

jobrunner/queries.py
bloodearnest: I'll leave it for now, enough yaks already!

jobrunner/cli/flags.py
bloodearnest: So, currently, when there's no table, `get` "just works", and `set` fails.

I think that's a good mode for the internal API to use.

However, with the flags cli command, there's always a user running it, so perhaps we should be strict about the table's existence for both `get` and `set`.

So we could support --create logic for both?

I think that's probably the best approach, I'll start on that, yell if you think otherwise.

PR #397, opened 2022-05-06
fix: add db maintenance mode.

bloodearnest:
This mode will a) kill any running db jobs and b) prevent
pending db jobs from starting.

Currently, it can be set manually by running

`python -m jobrunner.cli.flags set mode=db-maintenance`

Note: this never has effect in local_run scenarios as it uses the dummy
db - it only changes behaviour when there is a real database configured.


evansd (approved):
It's great how small the diff is considering the feature it gives us.

PR #404, opened 2022-05-13
Support tags as the job commit sha.

bloodearnest:
Previously, they had to be the actual commit sha. This changes handles
derefencing them if they are, then using the actual commit to check
history.

### DIFF
--- a/jobrunner/lib/git.py
+++ b/jobrunner/lib/git.py
@@ -98,13 +98,13 @@ def commit_reachable_from_ref(repo_url, commit_sha, ref):
     supplied commit is reachable from that ref.
     """
     ref_sha = get_sha_from_remote_ref(repo_url, ref)
-    # The commont case is that the cha is the current HEAD of the branch, so optimise for that.
+    # The commont case is that the sha is the current HEAD of the branch, so optimise for that.
     if commit_sha == ref_sha:
         return True
 
-    # The sha could be a tag - try derefence it and recheck against head
+    # The sha could be a tag - try to derefence it and recheck against head
     try:
-        # try deref commit sha in case is a tag
+        # try to deref commit sha in case is a tag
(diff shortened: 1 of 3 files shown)


jobrunner/lib/git.py
```diff
@@ -98,15 +98,27 @@ def commit_reachable_from_ref(repo_url, commit_sha, ref):
     supplied commit is reachable from that ref.
     """
     ref_sha = get_sha_from_remote_ref(repo_url, ref)
-    # The easy case and the case I expect to be hit almost every time as the UI
-    # currently only supports running against the branch head
+    # The commont case is that the cha is the current HEAD of the branch, so optimise for that.
```
ghickman, line 101:
```suggestion
    # The commont case is that the sha is the current HEAD of the branch, so optimise for that.
```

jobrunner/lib/git.py
```diff
@@ -98,15 +98,27 @@ def commit_reachable_from_ref(repo_url, commit_sha, ref):
...
     if commit_sha == ref_sha:
         return True
+
+    # The sha could be a tag - try derefence it and recheck against head
```
ghickman, line 105:
```suggestion
    # The sha could be a tag - try to derefence it and recheck against head
```

jobrunner/lib/git.py
```diff
@@ -98,15 +98,27 @@ def commit_reachable_from_ref(repo_url, commit_sha, ref):
...
+    try:
+        # try deref commit sha in case is a tag
```
ghickman, line 107:
```suggestion
        # try to deref commit sha in case is a tag
```

ghickman (approved):
Just a couple of comments tweaks.  Looks good to me but I'll leave checking the logic is sound for @evansd!

PR #411, opened 2022-05-18
feat: limit job cpu/memory usage by default.

bloodearnest:
The default limits are provided by env var config, and are set
deliberately low, but enough that they shouldn't affect `local_run`.
Actual backends would have higher defaults.

It adds the settings into JobDefinition, so that other executors can use
them. Also, we anticipate these limits being able to be included in the
JobRequest from job-server in the future.


ghickman (approved):

evansd:
Nice!

PR #412, opened 2022-05-19
feat: add pause mode

bloodearnest:
Setting mode flag to pause* will stop new jobs from starting. It does
not affect jobs that are already running.

This is designed to allow use to drain the queue of a jobs a bit before
a reboot or other maintenance. It was pleasantly simple to implement.

*`python3 -m jobrunner.cli.flags set mode=pause`

### DIFF
--- a/jobrunner/run.py
+++ b/jobrunner/run.py
@@ -53,10 +53,8 @@ def main(exit_callback=lambda _: False):
         api = get_executor_api()
 
     while True:
-        mode = get_flag("mode")
-        paused = get_flag("paused", "").lower() == "true"
 
-        active_jobs = handle_jobs(api, mode, paused)
+        active_jobs = handle_jobs(api)
 
         if exit_callback(active_jobs):
             break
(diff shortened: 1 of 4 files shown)


ghickman (approved):

jobrunner/run.py
```diff
@@ -229,6 +229,10 @@ def handle_job_api(job, api, mode=None):
         # reset state to pending and exit
         set_state(job, State.PENDING, "Waiting for database to finish maintenance")
         return
+    elif mode == "pause":
```
evansd, line 232: This feels orthogonal to db maintenance mode to me, so I'm not sure about reusing the same flag. How will this interact with the automatic db maintenance mode thread?

evansd (approved):
Pleasingly simple indeed! I love it when design choices pay off like this.

Just one question below.

jobrunner/run.py
bloodearnest: Well, using the same flag was designed to prevent the behaviours from interacting. Its one or the other (the `if` to this `elif` is `if mode == "db-maintenance" ...`)

Having said that, if it is in pause mode, and switched to db-maintenance mode via the check, then the non-db jobs *will* start executing, which is probably not what was intended.

So maybe another flag is a good idea.

evansd: Yeah, accidentally unsetting the pause mode was the thing I was worried about. I think that as long as there's one clear block of code where all these flags are inspected and the decision logic is implemented (as there is at present) then it will be easy enough to reason about how the different flags interact.

PR #413, opened 2022-05-19
fix: differentiate 137 exit codes between OOM and admin.

bloodearnest:
Previously, the message lists both causes as possible, which actually was more
confusing tothe user, as they didn't know which it was.

This detects when the jobs has been killed by the OOM killer, and provides
useful feedback.

### DIFF
--- a/jobrunner/executors/local.py
+++ b/jobrunner/executors/local.py
@@ -256,7 +256,15 @@ def finalize_job(job):
         message = config.DATABASE_EXIT_CODES.get(exit_code)
     # No database error, check for other known exit codes
     if message is None:
-        message = config.EXIT_CODES.get(exit_code)
+        # special case OOMKilled
+        if container_metadata["State"]["OOMKilled"]:
+            message = "Ran out of memory"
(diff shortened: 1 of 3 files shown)


ghickman (approved):

jobrunner/executors/local.py
```diff
@@ -256,7 +256,15 @@ def finalize_job(job):
         message = config.DATABASE_EXIT_CODES.get(exit_code)
     # No database error, check for other known exit codes
     if message is None:
```
evansd, line 258: This is a really nice UX improvement. I think the `OOMKilled` check could be done before the `allow_database_access` check though, because db jobs can in principle get OOM-killed to. I also wonder if we shouldn't merge the DATABASE_EXIT_CODES and EXIT_CODES dicts when looking up the error messages for database jobs, so that database jobs still get the standard messages for standard error codes. For instance, I can imagine an admin killing a database job and it would be nice to give the right message there.

bloodearnest: That's a good idea. I'll do a follow up PR

bloodearnest: #414

PR #417, opened 2022-05-20
fix: Remove pointless local_run metadata and log

rebkwok:
Previously I thought the "local" executor was used to run actions locally via opensafely-cli, so it might be useful to log that.  Since it's actually how jobs are run on the server too, it doesn't serve any useful purpose to log and could be confusing.


bloodearnest (approved):
I had been meaning to try figure out why jobs logs on the server had local_run = true! :D

PR #419, opened 2022-05-24
fix: send all current flag values to job-server

bloodearnest:
We now support a `mode` flag and a `pause`, as the two are independantly
enabled.

This change communicates all currently set flags to job-server, rather
than just the 'mode' flag.

### DIFF
--- a/jobrunner/sync.py
+++ b/jobrunner/sync.py
@@ -2,6 +2,7 @@
 Script which polls the job-server endpoint for active JobRequests and POSTs
 back any associated Jobs.
 """
+import json
 import logging
 import sys
 import time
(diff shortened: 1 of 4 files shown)


jobrunner/sync.py
```diff
@@ -66,13 +67,23 @@ def api_post(*args, **kwargs):
     return api_request("post", *args, **kwargs)
 
 
-def api_request(method, path, *args, **kwargs):
+def api_request(method, path, *args, headers=None, **kwargs):
```
evansd, line 70: What's the `headers` arg for? We always override it anyway it looks like.

evansd (approved):
Looks good. Just one weird thing below which I guess is left over from a previous version.

jobrunner/sync.py
bloodearnest: ah, I must have left that change in the stash

fixed, thanks

Issue #425, opened 2022-05-26
Abstract output handling from the executor.


bloodearnest:
Currently, executors must understand and implement matching output glob patterns, and handle unmatched patterns/outputs. It uses this to know which files to copy to persistent storage, primarily, but also communicates to the job loop in order to give users better debugging information.

This logic is common to all executors, and should ideally be handled by job-runner's main job handling, rather than having every executor independently implement it.

One way we could do this to split the finalise() api call into two, which I will tentatively call `inspect()` and `persist()`.

Calling inpsect() will create a `JobResults` object like `finalize()`, with approriate exit code and message. But rather than outputs/unmatched_patterns/unmatched_outputs, it will have a single `new_files` property, which will be a flat list of new files that were generated by the action (like we already do with the timestamp file).

The Job loop will then inspect this new files, applying globs to the file paths, filtering unmatched patterns and outputs. Then it will call `persist()` with two lists of files, one for high privacy, the other for medium privacy.  The api implementation of persist should:
a) validate that the all files exist or else error
b) validate that medium file list is a subset of high_privacy file list
c) persist the files to their given long term storage.

This way, the executors only need to implement the "get newer files" logic, which is fairly simple, and don't need to care about output_spec or pattern matching.

evansd:
I've wondered about something like this before myself. I do like the idea of shifting as much logic out of the executor as possible. I think there _are_ a few considerations against it. I'm not saying I think it's definitely a bad idea, just want to make sure we've thought it all through.

One issue is that I think we need to make any API changes with an eye to a future world where job-runner is simple(r) and stateless and all the coordination is handled by job-server. This API would involve shipping the full file list back out of the secure environment. Now, arguably that's fine because we're planning to ship the matched filenames back anyway and, given that we support glob patterns, the opportunity for steganographic data exfiltration via filenames is the same either way. But I'd like to make sure we've thought that through and are comfortable with it before we commit to this API.

Another pair of related issues is:
 (a) The "find unmatched files" feature was only intended a useful local debugging aid, rather than something that ought to be implemented in production. As it happened, given that we used exactly the same code for both in the early days, we ended up getting the feature for free in production. But I'm not sure we need to continue supporting it there.
[...]

Issue #407, opened 2022-05-16
Improve file-copy time in production


sebbacon:
Per https://github.com/opensafely-core/job-runner/issues/169#issuecomment-1127507045 and following comments, now we don't have to support docker-in-Windows in production, we can use temporary bind mounts rather than `docker cp` operations to make the setup phase of jobs _very much_ faster (it can take hours for the copy phase to complete at the moment).

PR #426, opened 2022-05-26
chore: refactor local volume management

bloodearnest:
This is introduces a `volume_api` interface, in order to be able to
support different docker volume implmentations.

The only implementation is the current docker volume one, and it simply
wraps the various functions that were previously used. It does however
change the api to take a `job`, in order to:

a) make the the volume name opaque to the executor, in order that it can
   be a named volume or a host path.
b) `job` has all the information needed various volume api fuctions to
   work.

Doing this refactor has highlighted that the current implementation
places the responsibility of figuring out outputs and error on the
executor, when really this should be shared logic. Have written this up
as #425

This is related to fixing #407

### DIFF
--- a/jobrunner/executors/local.py
+++ b/jobrunner/executors/local.py
@@ -10,6 +10,7 @@ from pathlib import Path
 from pipeline.legacy import get_all_output_patterns_from_project_file
 
 from jobrunner import config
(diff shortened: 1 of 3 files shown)


evansd (approved):
:+1: This makes sense

PR #430, opened 2022-05-26
chore: refactor db codes to be executor agnostic

bloodearnest:
These codes mean the same thing regardless of executor, so handle them
in the job loop rather than in the local executor.

This is also faster, as the tests do not need docker anymore

Also, this was discovered by trying to reorganise config.py in order to
group and documenet the general job-runner config from the local docker
executor config, so that reorganising of config.py is included.


jobrunner/config.py
```diff
@@ -207,13 +183,45 @@ def parse_job_resource_weights(config_file):
...
-EXIT_CODES = {
+
+
+DEFAULT_JOB_CPU_COUNT = float(os.environ.get("DEFAULT_JOB_CPU_COUNT", 2))
+DEFAULT_JOB_MEMORY_LIMIT = os.environ.get("DEFAULT_JOB_MEMORY_LIMIT", "4G")
+
+
+EXECUTOR = os.environ.get("EXECUTOR", "jobrunner.executors.local:LocalDockerAPI")
+
+# LocalDockerAPI execturo specific configuration
```
ghickman, line 194:
```suggestion
# LocalDockerAPI executor specific configuration
```

pyproject.toml
```diff
@@ -9,6 +9,7 @@ exclude = '''
     | \.venv
     | htmlcov
     | venv
+    | lib
```
ghickman, line 12: Could you add a comment on why this is ignored, it's not obvious to me as someone unfamiliar with the project!

ghickman (approved):

evansd:
Retrospective :+1: from me

PR #433, opened 2022-05-27
feat: add a bind mount volume api

bloodearnest:
This volume api implementation uses host directories and bind
mounts them into the docker container.

This approach sould be more efficient than the docker
volume api, which has to `docker cp` everything onto the volume. This
can just fast local filesystem copy, using `os.sendfile` on linux.

This is tested on linux only, and had to be explicitly enabled via
config, and is indended to be used in production backends.

### DIFF
--- a/jobrunner/executors/volumes.py
+++ b/jobrunner/executors/volumes.py
@@ -60,7 +60,18 @@ def host_volume_path(job):
 
 class BindMountVolumeAPI:
     def volume_name(job):
-        return host_volume_path(job)
+        """Return the absolute path to the volume directory.
+
+        In case we're running inside a docker container, make sure the path is
+        relative to the *hosts* POV, not the container.
+        """
+        local_path = host_volume_path(job)
+        if config.DOCKER_HOST_VOLUME_DIR is None:
+            return local_path
+
+        return config.DOCKER_HOST_VOLUME_DIR / local_path.relative_to(
+            config.HIGH_PRIVACY_VOLUME_DIR
+        )
 
     def create_volume(job, labels=None):
         host_volume_path(job).mkdir()
--- a/jobrunner/executors/volumes.py
+++ b/jobrunner/executors/volumes.py
@@ -1,37 +1,122 @@
-from jobrunner.lib import docker
+import importlib
+import shutil
+from collections import defaultdict
+from pathlib import Path
+
+from jobrunner import config
+from jobrunner.lib import atomic_writer, docker
 
 
-def volume_name(job):
+def copy_file(source, dest, follow_symlinks=True):
+    """Efficient atomic copy.
+
+    shutil.copy uses sendfile on linux, so should be fast.
+    """
+    # ensure path
+    dest = Path(dest)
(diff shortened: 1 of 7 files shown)


tests/conftest.py
```diff
@@ -46,6 +48,38 @@ def tmp_work_dir(monkeypatch, tmp_path):
...
+    # host's docker service (which the jobrunner-inna-container uses to run
+    # jobs) can bind mount them into the jobs we've set above available on the
+    # host.
+    #
+    # We do this via:
+    #
+    # a) mounting a host directory into the container as /tmp
+    # b) letting pytest do it's thing in /tmp
+    # c) calculate and set DOCKER_HOST_VOLUME_DIR to point the the host's view
+    # of the tests' temp dirs.  filesystem this is so that we can provide
```
madwort, line 65: > so that we can provide ... ?

bloodearnest: Ah, yeah, will fix, thanks

jobrunner/executors/volumes.py
```diff
@@ -1,37 +1,127 @@
...
+
+    def copy_from_volume(job, src, dst, timeout=None):
+        # this is only use to copy final outputs/logs. So we optimise by doing
+        # a rename. If the volume directory and the workspace directory are on
+        # the same filesystem, this should mean zero copying.
+        # Note: We don't respect the timeout.
+        #
+        path = host_volume_path(job) / src
+        dst.parent.mkdir(parents=True, exist_ok=True)
+        path.rename(dst)
```
evansd, line 104: This isn't idempotent, and it means that the whole finalize step is no longer idempotent which I don't think we can tolerate. I think it would be possible to make it idempotent via some sort of tombstone file trickery but the details get a bit fiddly. It might be easiest in the first instance to make this a copy, which will still be a win as it's local, and then worry about how to do moves idempotently later.

tests/conftest.py
```diff
@@ -46,6 +48,48 @@ def tmp_work_dir(monkeypatch, tmp_path):
         monkeypatch.setattr(
             f"jobrunner.config.{config_var}", tmp_path / config_var.lower()
         )
+
+    # Ok, so this is a bit complex.
```
evansd, line 52: Looks like you had "fun" here

evansd (commented):
Large round of applause for getting the jobrunner-in-docker thing working. The only major wrinkle is the idempotentcy issue below. Might be a case of taking the quicker wins now and working on the full speed up later?

jobrunner/executors/volumes.py
bloodearnest: I have updated with a straight copy.

We of course relying on the duplication aspects of copying to provide various system properties like this. But we way the cost in two ways: latency and capacity. This move to faster coping should help with the latency, but not with the capacity cost. That's the big win that CoW and/or `rename` approach gains us.

evansd (approved):
Yeah, I think we will want to move to using renames if we can. It's just going to take some work (and possibly some API changes) to get it right.

Issue #440, opened 2022-06-15
Handle initial executor state being ERROR


bloodearnest:
Currently we don't have an explicit branch [in this if statement](https://github.com/opensafely-core/job-runner/blob/master/jobrunner/run.py#L150) that handles when `initial_status == ExecutorState.ERROR`. This means `new_status` is not defined in this case, and we blow up [a few lines later](https://github.com/opensafely-core/job-runner/blob/master/jobrunner/run.py#L209).

This is clearly a bug.

It's interesting that we have not to my knowledge hit this case. I suspect is it because we don't actually modify things asynchronously with the local docker executor, the state is only every changed in response to use calling an API function. Whereas, with graphnet's k8s environment, sometimes this has occurred. It's possibly they are not quite using the correct intended semantics of `ExecutorState.ERROR`, I've talked that through with them, and they are double checking that they're not returning it when a user's job failed, which is *not* an executor error.

However, we should clearly not blow up when this happens, and handle it appropriately, which I think is fairy simple - handle it early and fail the job.

PR #445, opened 2022-06-22
fix: fix error when new_status is an error

chrisyeunggraphnet:


bloodearnest:
Thanks for this, @chrisyeunggraphnet, but I already had a fix it the works for this, with tests to avoid it regressing #440

PR #451, opened 2022-07-06
fix: communicate paused state to users

bloodearnest:


jobrunner/run.py
```diff
@@ -77,7 +77,7 @@ def handle_jobs(api: Optional[ExecutorAPI]):
 
 def handle_single_job(job, api):
     mode = get_flag_value("mode")
-    paused = get_flag_value("paused", "").lower() == "true"
+    paused = str(get_flag_value("paused")).lower() == "true"
```
madwort, line 80:
```suggestion
    paused = str(get_flag_value("paused", "False")).lower() == "true"
```

madwort: the nittiest of nits

bloodearnest: Fair

jobrunner/run.py
```diff
@@ -120,6 +120,11 @@ def handle_job(job, api, mode=None, paused=None):
     if paused:
         if job.state == State.PENDING:
             # do not start the job, keep it pending
+            set_state(
```
evansd, line 123: I think you'd be better of using `set_message()` here because you're not actually changing the state , and that will avoid being so spammy in the logs:
https://github.com/opensafely-core/job-runner/blob/75d76821ac0aafc995ae51dfaed184cd5f0e52a6/jobrunner/run.py#L420

evansd (approved):
Looks good. One small recommendation below.

Issue #448, opened 2022-06-30
Remove support for cohortextractor-v2 from CLI and Job Runner


inglesp:
We added this to facilitate end-to-end testing of Cohort Extractor v2 when that still existed, but it has gone now and no one every used it in anger.

I noticed this when running `opensafely pull`. It's not a big problem but it puts a small extra burden on users to download the unused Docker image and it's a small increase in our attack surface.

rebkwok:
Removing support for ce-v2 in opensafely-cli and job-runner will mean study repos that use it won't work.  There are only 2, the [databricks POC study](https://github.com/opensafely/med-safety-PoC) and a [databricks test repo](https://github.com/opensafely/databricks-test).  I don't think it's an issue if we break the current versions of those repos; they're going to need to update to databuilder anyway

ghickman:
Should we drop support for this in the pipeline library too?

rebkwok:
yes, I'd assume so

PR #454, opened 2022-08-10
fix: Remove support for cohortextractor-v2

rebkwok:
See #448 (will also need an update to the vendored jobrunner in opensafely-cli once this PR is merged)


ghickman (approved):
🎉

Issue #459, opened 2022-08-26
Fix missing MAX_WORKERS enforcement in executor api


bloodearnest:
In the move the executor API, we made the decision for the executor to be responsible for applying back pressure when there are no free resources. Different implementations will have different resource levels and ways to query current usage.

However, we never actually implemented this for the local docker executor. So since we switched the the new executor API on prod (Nov 2021), there have been *no concurrent job limits*.

This actually explains why we had repeating resource issues in TPP leading up to the VM switch in April. We haven't seen resource exhaustion much since then, which is likely primarily due to the increased resources. Additionally, many `project.yaml`'s action dependencies enforce a maximum concurrency anyway, except for highly parameterised jobs.

It also explains why we've see a few vague reports from users that local runs have slowed down recently, and they've been forced to run jobs one by one, and hitting memory issue. I originally thought this was to do with the addition of a per-job 4GB memory limit, but it's likely because it's trying to run all the jobs it can at once. Note that executor API usage for `opensafely` tool was switched on by default in early May, quite a bit later than prod.

A few such highly parallel projects ran together on TPP this week, bring the total concurrent job count to 47 and exhausting memory. Due to the cgroup limiting, the VM was still running, but docker itself is included in this limitation, so it was struggling to respond to API queries. We killed 6 jobs, and resource usage dropped enough to unstick things.

As an emergency measure, I manually patched the old concurrency management into the TPP job-runner, and restarted it. This worked, and currently MAX_WORKERS=24 and is being respected, anecdotally running at around 60% mem/cpu when full.

I think the first response is to formally add the old management back in, and release a new version of opensafely tool.  This uses the sqlite database to track currently running job numbers.

However, do we want to apply a concurrency limit in the job-runner scheduler, or is our original goal of the backend providing back pressure still the correct approach?

We could add the current mechanism into the local docker executor, but it relies on accessing the sqlite db, which is a layering violation. We'd probably want to query docker for number of running jobs to get a currently level, rather than relying on the db, but there may be a performance/load issue there (give that docker api it prone to timeouts).

bloodearnest:
The hotfix I've applied to prod in TPP is working, but I'm reluctant to add that as the fix, as it would by default limit concurrency on graphnet and others to the number of CPUs **job-runner** is running on, which is in no way related to it's actually capacity. But it's an easy fix to tell them set graphnet to set MAX_WORKERS to override it, I guess

[...]

PR #460, opened 2022-08-26
fix: re-introduce MAX_WORKERS check

bloodearnest:
This was accidentally removed in executor API refactor. See #459 for
more details.

### DIFF
--- a/jobrunner/run.py
+++ b/jobrunner/run.py
@@ -67,7 +67,7 @@ def handle_jobs(api: Optional[ExecutorAPI]):
     return active_jobs
 
 
-# we do not control the tranisition from these states, the executor does
+# we do not control the transition from these states, the executor does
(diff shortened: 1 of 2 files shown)


jobrunner/run.py
```diff
@@ -182,6 +182,15 @@ def handle_job(job, api, mode=None, paused=None):
             )
             return
 
+        # Temporary fix to reintroduce concurrency limits lost in the move to
+        # the executor API. Ideally this should be the responsiblitt of the
```
rebkwok, line 186: typo "responsiblitt"

rebkwok (approved):
LGTM. One typo in a comment is all.

jobrunner/run.py
bloodearnest: thanks fixed

PR #467, opened 2022-09-21
feat: add schema migrations framework

bloodearnest:
We have some pending db changes to make. In preparation for this, this
change adds a formal migration system to the micro-ORM.

 * To avoid having to package raw sql files, we track the migrations as
   string literals in python code, as dict mapping version to SQL
   string.

 * We use the `PRAGMA user_version` db variable to track the migration
   version, using transactions to ensure it is set appropriately.

 * We add a migrate command, analagous to django `manage.py migrate`.
   This will ensure a db exists and is at the latest schema version.
   This is the only way to create/migrate a db, and should be run as an
   explicit step as part of production deployments

 * Previously, in production schema creation happened lazily when the
   service started up. Extending this to apply migrations automatically
   on service start up seemed risky. Instead we add an explict check
   that the db exists and is up to date, or else we do not start up.
   This is designed to catch configuration issues or unapplied
   migrations in production.

 * For `jobrunner.cli.local_run` however, we do automatically apply the
   migrations on behalf of the user, as each project has its own db. The
   data they have is an optimisation, rather than a requirement, as they
   can always re-run the entire project if needed.

 * Note: there are no actual migrations in this initial change - the
   migration framework will be used in subsequent PRs

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -163,3 +163,51 @@ you can do so:
 2. Add an authorized ed25519 private key for that user to `docker/ssh/id_ed25519`.
 3. Run `touch docker/ssh/id_ed25519.authorized` to let Make know that it is all
    set up.
+
+
+## Database schema and migrations
+
+jobrunner uses a minimal ORM-lite wrapper to talk to the DB.
+
+The current version of a tables schema definition is stored in the the
+`__tableschema__` attribute for that model's class, i.e. `Job.__tableschema__`.
+This is use to create the table in dev and test, so migrations are not usually
+needed in those cases.
+
+### Adding a migration
+
+However, we also occasionally need to apply changes to this schema in
+production, or in a user's local opensafely-cli database.
+
+To do this, we track migrations in `jobrunner/models.py`. Add a migration like so:
+
+```python
+database.migration(1, """
+DDL STATEMENT 1;
+DDL STATEMENT 2;
+""")
+```
+
+These statements are run together in a single transaction, along with
+incrementing the `user_version` in the database.
+
+Note: be aware that there are various restrictions on ALTER TABLE statements in
+sqlite:
+
+https://www.sqlite.org/lang_altertable.html#alter_table_add_column
+
+
+### Applying migrations
+
+
+Trying to run jobrunner as a service will error if the database does not exist
+or is out of date, as a protection against misconfiguration.
+
+To initialise or migrate the database, you can use the migrate command:
+
+```sh
+python -m jobrunner.cli.migrate
+```
+
+Note that for `jobrunner.cli.local_run`, which is used by opensafely-cli,
+migrations are automatically applied.
--- a/jobrunner/lib/database.py
+++ b/jobrunner/lib/database.py
@@ -13,12 +13,14 @@ import json
 import sqlite3
 import threading
 from enum import Enum
+from pathlib import Path
 
 from jobrunner import config
 
 
 CONNECTION_CACHE = threading.local()
 TABLES = {}
+MIGRATIONS = {}
 
 
 def databaseclass(cls):
@@ -29,6 +31,12 @@ def databaseclass(cls):
     assert "id" in fields, "must have primary key 'id'"
     TABLES[dc.__tablename__] = dc
     return dc
+
+
+def migration(version, sql):
+    """Used to record a migration"""
+    assert version not in MIGRATIONS, f"Migration {version} already exists."
+    MIGRATIONS[version] = sql
 
 
 def insert(item):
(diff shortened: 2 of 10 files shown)


DEVELOPERS.md
```diff
@@ -163,3 +163,47 @@ you can do so:
...
+To do this, we track migrations in `jobrunner/models.py`. Add a migration like so:
+
+```python
+database.migration(1, """
+DDL STATEMENT 1;
+DDL STATEMENT 2;
+""")
+```
+
+These statements are run together in a single transaction, allong with
```
ghickman, line 191:
```suggestion
These statements are run together in a single transaction, along with
```

jobrunner/lib/database.py
```diff
@@ -31,6 +32,12 @@ def databaseclass(cls):
     return dc
 
 
+def migration(version, sql):
+    """Used to record a migration"""
+    assert version not in MIGRATIONS, f"Migration {version} alread exists."
```
ghickman, line 37:
```suggestion
    assert version not in MIGRATIONS, f"Migration {version} already exists."
```

jobrunner/lib/database.py
```diff
@@ -153,37 +164,104 @@ def get_connection():
...
     # Enable autocommit so changes made outside of a transaction still get
     # persisted to disk. We can use explicit transactions when we need
     # atomicity.
     conn.isolation_level = None
     # Support dict-like access to rows
     conn.row_factory = sqlite3.Row
-    schema_count = list(conn.execute("SELECT COUNT(*) FROM sqlite_master"))[0][0]
-    if schema_count == 0:
+
+    latest_version = max(migrations) if migrations else 0
```
ghickman, line 211: You could also use the default kwarg here:

    `max(migrations, default=0)`

if you wanted to drop the conditional.  I'm not sure it adds much value though.

pyproject.toml
```diff
@@ -25,6 +25,10 @@ add_job = "jobrunner.cli.add_job:run"
 kill_job = "jobrunner.cli.kill_job:run"
 local_run = "jobrunner.cli.local_run:run"
 retry_job = "jobrunner.cli.retry_job:run"
+prepare_for_reboot = "jobrunner.cli.prepare_for_reboot:run"
+flags = "jobrunner.cli.flags:run"
+migrate = "jobrunner.cli.migrate:run"
```
ghickman, line 30: Could you alphabetise these? Makes them a little easier to skim 🙂

ghickman (approved):
Looks great 👍

Couple of suggestions you might find useful.

jobrunner/lib/database.py
bloodearnest: Thats more readable, thanks

jobrunner/lib/database.py
bloodearnest: fixed

pyproject.toml
bloodearnest: done

DEVELOPERS.md
bloodearnest: fixed

jobrunner/service.py
```diff
@@ -32,7 +32,7 @@ def main():
     configure_logging(fmt)
     # ensure db is set up before we start any threads, otherwise if we don't already have
     # a database (which we don't in tests) then threads will race to create the schema
-    get_connection()
+    init_db()
```
evansd, line 35: This call means that if the database file isn't in the place we expect it we'll just create a new empty file and carry on without error. But I thought the idea was that we'd fail hard unless the database file had already been created?

evansd (approved):

jobrunner/lib/database.py
```diff
@@ -153,37 +164,104 @@ def get_connection():
...
+    if filename not in cache:
+        raise Exception("Database not initialised")
+    return cache[filename]
+
+
+class MigrationNeeded(Exception):
+    pass
+
+
+def init_db(filename=None, migrations=None):
```
evansd, line 176: Related to the comment below, it looks like `init_db()` bundles up both opening the connection and creating the database file if it doesn't exist. But I thought we wanted to separate those?

evansd (approved):

jobrunner/service.py
bloodearnest: Hmm, yes we did discuss that. But my main concern was to detect when migrations needed applying, so that got dropped. And ever

It's a bit messy, as it means init_db will need two modes when a db doesn't exist one where it fails (which is only used here) and one where it doesn't (which is used everywhere else, i.e. tests and local_run.

And how do we know if this is a new backend or not? We'd need a way to create the initial db so this check wouldn't error.  The current migrate command doesn't do that, it assumes a db exists. 

So, if we use django as our model, then an initial migrate *is* needed before a service will run successfully in a new environment. So perhaps the migrate command should do that.

We'll still need to two modes though, or possibly two different functions.

bloodearnest:
Updated PR to answer @dave's point and to clarify the "manual" step.

Basically, to create a db, we now have to run migrate first, just like django.

Added a start up check that the db is in the right state, to be cautious initially.

Also implemented a slightly nicer in memory db set up.

jobrunner/service.py
bloodearnest: Have updated to implement the above

lucyb (approved):

PR #468, opened 2022-09-23
 feat: record user/project/org metadata

bloodearnest:
Job server is now sending project/org data in job requests[1].

This change stores this information in the db. The main use of this so
that we can emit telemetry that is tagged by these values in future.

It includes the first migration using the new migration system.

[1] https://github.com/opensafely-core/job-server/pull/2095/files


ghickman (approved):

bloodearnest:
Refactoring this to not need a migration

PR #469, opened 2022-09-26
Make job-runner aware of sqlrunner

iaindillingham:
As well as adding sqlrunner's image to the list of allowed images, we also let job-runner know that sqlrunner's image requires DB access. We modify `requires_db_access` slightly, because unlike the other images that require DB access, we don't invoke sqlrunner's image with a (sub)command.

See also opensafely-core/sqlrunner#15 for building and publishing sqlrunner's image.

### DIFF
--- a/jobrunner/lib/commands.py
+++ b/jobrunner/lib/commands.py
@@ -5,10 +5,14 @@ def requires_db_access(args):
     valid_commands = {
         "cohortextractor": ("generate_cohort", "generate_codelist_report"),
         "databuilder": ("generate-dataset",),
+        "sqlrunner": None,  # all commands are valid
     }
     if len(args) <= 1:
         return False
 
     image, command = args[0], args[1]
     image = image.split(":")[0]
-    return command in valid_commands.get(image, [])
+    if image in valid_commands:
+        if valid_commands[image] is None or command in valid_commands[image]:
+            return True
+    return False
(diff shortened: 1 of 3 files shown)


jobrunner/lib/commands.py
```diff
@@ -5,10 +5,14 @@ def requires_db_access(args):
...
+        "sqlrunner": (),  # all commands are valid
     }
     if len(args) <= 1:
         return False
 
     image, command = args[0], args[1]
     image = image.split(":")[0]
-    return command in valid_commands.get(image, [])
+    if image in valid_commands:
+        if not valid_commands[image] or command in valid_commands[image]:
```
evansd, line 16: This is fine as it is, but just to note in passing that I would probably have used a different type here as the "matching anything" escape hatch rather than an empty tuple. You could use `None` and have the check as:
```py
if valid_commands[image] is None or command in valid_commands[image]:
    ...
```
I think it slightly reduces the possibility of accidental confusion. But the context is so simple here I doubt it makes much practical difference.

evansd (approved):

jobrunner/lib/commands.py
iaindillingham: Thanks for the comment, Dave. I've replaced the empty tuple with `None` and it makes it easier to reason about the nested `if`. I've force pushed; 1d5b044 includes the change.

PR #474, opened 2022-09-28
feat: add opentelemetry tracing

bloodearnest:
This adds opentelemetry tracing to running jobs.

A new jobrunner/tracing.py module handles most of the work, and we use
it to trace state changes between the fine grained StatusCode states
added previously.  We store as many attributes on each span as possible.

This includes a migration to add somewhere to store a job's trace
context and last changed timestamp.

We also emit "marker spans" that indicate that a the job has entered
a certain state, but not yet exited it yet. This is intedended to be
useful for ongoing jobs.

Also, we add a final "RUN" span once the job has completed that serves
as coarser grained span to use for graphing whole job metics.

### DIFF
--- a/jobrunner/create_or_update_jobs.py
+++ b/jobrunner/create_or_update_jobs.py
@@ -226,7 +226,7 @@ def recursively_build_jobs(jobs_by_action, job_request, pipeline_config, action)
         job_request_id=job_request.id,
         state=State.PENDING,
         status_code=StatusCode.CREATED,
-        # time in ns
+        # time in nanoseconds
         status_code_updated_at=int(timestamp * 1e9),
         repo_url=job_request.repo_url,
         commit=job_request.commit,
--- a/jobrunner/create_or_update_jobs.py
+++ b/jobrunner/create_or_update_jobs.py
@@ -12,7 +12,7 @@ import time
 
 from pipeline import RUN_ALL_COMMAND, ProjectValidationError, load_pipeline
 
-from jobrunner import config
+from jobrunner import config, tracing
 from jobrunner.actions import get_action_specification
 from jobrunner.lib.database import exists_where, insert, transaction, update_where
 from jobrunner.lib.git import GitError, GitFileNotFoundError, read_file_from_repo
@@ -226,6 +226,8 @@ def recursively_build_jobs(jobs_by_action, job_request, pipeline_config, action)
         job_request_id=job_request.id,
         state=State.PENDING,
         status_code=StatusCode.CREATED,
+        # time in ns
+        status_code_updated_at=int(timestamp * 1e9),
         repo_url=job_request.repo_url,
         commit=job_request.commit,
         workspace=job_request.workspace,
@@ -238,6 +240,7 @@ def recursively_build_jobs(jobs_by_action, job_request, pipeline_config, action)
         created_at=int(timestamp),
         updated_at=int(timestamp),
     )
+    tracing.initialise_trace(job)
 
     # Add it to the dictionary of scheduled jobs
     jobs_by_action[action] = job
@@ -331,6 +334,8 @@ def create_failed_job(job_request, exception):
         job_request_id=job_request.id,
         state=state,
         status_code=code,
+        # time in ns
+        status_code_updated_at=int(now * 1e9),
         repo_url=job_request.repo_url,
         commit=job_request.commit,
         workspace=job_request.workspace,
@@ -341,6 +346,7 @@ def create_failed_job(job_request, exception):
         updated_at=int(now),
         completed_at=int(now),
     )
+    tracing.initialise_trace(job)
     insert_into_database(job_request, [job])
 
 
--- a/jobrunner/create_or_update_jobs.py
+++ b/jobrunner/create_or_update_jobs.py
@@ -334,7 +334,7 @@ def create_failed_job(job_request, exception):
         job_request_id=job_request.id,
         state=state,
         status_code=code,
-        # time in ns
+        # time in nanoseconds
         status_code_updated_at=int(now * 1e9),
         repo_url=job_request.repo_url,
         commit=job_request.commit,
--- a/jobrunner/models.py
+++ b/jobrunner/models.py
@@ -14,10 +14,12 @@ import hashlib
 import secrets
 from enum import Enum
 
-from jobrunner.lib.database import databaseclass
+from jobrunner.lib.database import databaseclass, migration
 from jobrunner.lib.string_utils import slugify
 
 
+# this is the overall high level state the job-runner uses to decide how to
+# handle a particular job.
 class State(Enum):
     PENDING = "pending"
     RUNNING = "running"
@@ -140,6 +142,8 @@ class Job:
             updated_at INT,
             started_at INT,
             completed_at INT,
+            trace_context TEXT,
+            status_code_updated_at INT,
 
             PRIMARY KEY (id)
         );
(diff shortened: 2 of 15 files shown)


jobrunner/create_or_update_jobs.py
```diff
@@ -226,6 +226,8 @@ def recursively_build_jobs(jobs_by_action, job_request, pipeline_config, action)
         job_request_id=job_request.id,
         state=State.PENDING,
         status_code=StatusCode.CREATED,
+        # time in ns
```
madwort, line 229:
```suggestion
        # time in nanoseconds
```

jobrunner/create_or_update_jobs.py
```diff
@@ -331,6 +334,8 @@ def create_failed_job(job_request, exception):
         job_request_id=job_request.id,
         state=state,
         status_code=code,
+        # time in ns
```
madwort, line 337:
```suggestion
        # time in nanoseconds
```

jobrunner/create_or_update_jobs.py
madwort: the nittiest of nits

jobrunner/models.py
```diff
@@ -255,7 +275,7 @@ def slug(self):
         Use a human-readable slug rather than just an opaque ID to identify jobs in
         order to make debugging easier
         """
-        return slugify(f"{self.project}-{self.action}-{self.id}")
+        return slugify(f"{self.workspace}-{self.action}-{self.id}")
```
madwort, line 278: ?

jobrunner/run.py
```diff
@@ -469,17 +477,36 @@ def set_code(job, code, message, error=None, **attrs):
...
 
+        # use higher precision timestamp for state change time
+        job.status_code_updated_at = timestamp_ns
         update_job(job)
+
+        if code in FINAL_STATUS_CODES:
+            # transitioning to a final state, so just record that state
+            tracing.finish_current_state(
+                job, timestamp_ns, final=True, error=error, message=message, **attrs
+            )
```
madwort, line 503: would be nice to pull this into a special function to clarify purpose

jobrunner/tracing.py
```diff
@@ -0,0 +1,226 @@
...
+    emitting so few spans that it is easier just to keep it simpler, and it
+    makes testing easier.
+    """
+    if "OTEL_EXPORTER_OTLP_HEADERS" in os.environ:
+        if "OTEL_SERVICE_NAME" not in os.environ:
+            raise Exception(
+                "OTEL_EXPORTER_OTLP_HEADERS is configured, but missing OTEL_SERVICE_NAME"
+            )
+        if "OTEL_EXPORTER_OTLP_ENDPOINT" not in os.environ:
+            os.environ["OTEL_EXPORTER_OTLP_ENDPOINT"] = "https://api.honeycomb.io"
```
madwort, line 39: open question - should this default to functioning (honeycomb url) or safe (invalid url)? for TPP we're confident that if it defaults to honeycomb the firewall will block it, but is that an assumption we can lean on in future?

jobrunner/tracing.py
```diff
@@ -0,0 +1,226 @@
...
+    return {k: "" if v is None else v for k, v in attrs.items()}
+
+
+def initialise_trace(job):
+    """Initialise the trace for this job by creating a root span.
+
+    We store the serialised trace context in the db, so we can reuse it for
+    later spans.
+
+    We create a root span, which is a requirement in OTEl. For this reason we
```
madwort, line 91:
```suggestion
    We create a root span, which is a requirement in OTel. For this reason we
```
nit

jobrunner/tracing.py
```diff
@@ -0,0 +1,226 @@
...
+    assert job.status_code is not None, "job has no initial StatusCode"
+    assert (
+        job.status_code_updated_at is not None
+    ), "job has no initial status_code_updated_at"
+
+    job.trace_context = {}
+    tracer = trace.get_tracer("jobs")
+    attrs = trace_attributes(job)
+
+    # convert created_at to ns
```
madwort, line 104:
```suggestion
    # convert created_at to nanoseconds
```

jobrunner/tracing.py
```diff
@@ -0,0 +1,226 @@
...
+    start_time = int(job.created_at * 1e9)
+
+    # TraceContextTextMapPropagator only works with the current span, sadlmy
```
madwort, line 107:
```suggestion
    # TraceContextTextMapPropagator only works with the current span, sadly
```

jobrunner/tracing.py
```diff
@@ -0,0 +1,226 @@
...
+        if final:
+            # final states have no duration, so make last for 1 sec, just act
+            # as a marker
+            end_time = int(end_time + 1e9)
+
+        record_job_span(job, name, start_time, end_time, error, **attrs)
+
+        if final:
+            # record a full span for the entire run
+            # trace vanity: have the job start 1us before the actualy job, so
```
madwort, line 136:
```suggestion
            # trace vanity: have the job start 1us before the actual job, so
```

jobrunner/tracing.py
```diff
@@ -0,0 +1,226 @@
...
+    # trace the initial job state trace
+    start_new_state(job, job.status_code_updated_at)
+
+
+def finish_current_state(job, timestamp_ns, final=False, error=None, **attrs):
+    """Record a span representing the state we've just exited."""
+    try:
+        name = job.status_code.name
+        # Note: this *must* be timestamp as integer nanoseconds
+        start_time = job.status_code_updated_at
```
madwort, line 123: wondering whether renaming this var to `job.status_code_updated_at_ns` could remove some of these comments?

jobrunner/tracing.py
```diff
@@ -0,0 +1,226 @@
...
+            # final states have no duration, so make last for 1 sec, just act
+            # as a marker
+            end_time = int(end_time + 1e9)
+
+        record_job_span(job, name, start_time, end_time, error, **attrs)
+
+        if final:
+            # record a full span for the entire run
+            # trace vanity: have the job start 1us before the actualy job, so
+            # it shows up first in the trace
```
madwort, line 137: YES!!

jobrunner/tracing.py
```diff
@@ -0,0 +1,226 @@
...
+        # make sure trace failures do not error the job
+        logger.exception(f"failed to trace state for {job.id}")
+
+
+def start_new_state(job, timestamp_ns, error=None, **attrs):
+    """Record a marker span to say that we've entered a new state."""
+    try:
+        name = f"ENTER {job.status_code.name}"
+        start_time = timestamp_ns
+        # fix the time for these synthenic marker events at one second
```
madwort, line 151:
```suggestion
        # fix the time for these synthetic marker events at one second
```

jobrunner/tracing.py
madwort: hmmm, some other things are nanoseconds but not `_ns~`, maybe just leave it as-is

tests/test_create_or_update_jobs.py
```diff
@@ -21,7 +21,7 @@ def disable_github_org_checking(monkeypatch):
 
 
 # Basic smoketest to test the full execution path
-def test_create_or_update_jobs(tmp_work_dir):
+def test_create_or_update_jobs(tmp_work_dir, db):
```
madwort, line 24: `db` seems unrelated?

madwort (changes requested):
looks good, just a few nits

tests/test_create_or_update_jobs.py
bloodearnest: We now query the db for job_request metadata, where as before we just used the `job` instance that. Given that, this change avoid using the default db and gives it its own isolated db for hygiene.

jobrunner/tracing.py
bloodearnest: It's a fair question.

I think given that we're happy for any such data to live in honeycomb eventually, I think it's ok?

It makes configuring local runs to emit to honeycomb a bit easier, is the main benefit.

madwort: yep ok

madwort (approved):

PR #483, opened 2022-10-06
Fix: switch to BatchSpanProcessor for senting telemtry

bloodearnest:
This can queue and retry when there are network outages.

Max queue size can be controlled via setting the OTEL_BSP_MAX_QUEUE_SIZE
env var.

Our tests still use SimpleSpanProcessor for synchronous sending.

Fixes: 481


jobrunner/tracing.py
```diff
@@ -17,19 +17,22 @@
...
+    # Note: BatchSpanProcessor is configured via env vars:
+    # https://opentelemetry-python.readthedocs.io/en/latest/sdk/trace.export.html#opentelemetry.sdk.trace.export.BatchSpanProcessor
     provider.add_span_processor(processor(exporter))
 
 
-def setup_default_tracing():
+def setup_default_tracing(processor=None):
     """Inspect environmenet variables and set up exporters accordingly.
 
     We use the SimpleSpanProcessor by default, which is synchronous. The
```
madwort, line 30: docs need updating

madwort (changes requested):
yep easy change, just needs the comment changing to reflect the code

bloodearnest:
good catch, fixed

jobrunner/tracing.py
```diff
@@ -17,19 +17,22 @@
 trace.set_tracer_provider(provider)
 
 
-def add_exporter(exporter, processor=SimpleSpanProcessor):
-    # we default to SimpleSpanProcessor so that it's synchronous for tests
+def add_exporter(exporter, processor=BatchSpanProcessor):
+    """Utility method to add an exporter.
+
+    We use the BatchSpanProcessor by default, which is the defasult for
```
madwort, line 23:
```suggestion
    We use the BatchSpanProcessor by default, which is the default for
```

madwort (approved):
good, just one more nit

PR #490, opened 2022-10-12
fix: optimise loop for synchronous prepare/finalize

bloodearnest:
The job loop assumes all operations are async. But for the local
executor this isn't true - both `prepare()` and `finalize()` are
synchronous.

This leads head-of-line blocking when a new job request arrives, in that
all jobs that can run are synchronously prepared in one pass of the
loop. This blocks those jobs from executing until the loop goes round
again, which can be a long time (e.g 30+mins), as the file copying for
n jobs can take a while.

This change allows an executor to declare it has certain transitions
that are synchronous, and if the loop detects this, it will immeadiately
run the the handler function again to move it to the next state.

For the local executor, we declare that both PREPARING and FINALIZING
are synchronous.

### DIFF
--- a/jobrunner/executors/local.py
+++ b/jobrunner/executors/local.py
@@ -87,6 +87,8 @@ def workspace_is_archived(workspace):
 class LocalDockerAPI(ExecutorAPI):
     """ExecutorAPI implementation using local docker service."""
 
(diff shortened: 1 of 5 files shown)


evansd:
This is all actually quite neat, isn't it? I'd be happy with this as a solution.

evansd (approved):

PR #496, opened 2022-10-21
Fix broken import statement

lucyb:
In the process of writing a test, I also noticed that the call to mark_job_as_failed was failing in database.py because there was no enum value for the "killed by admin" status code, so this has also been added into the enum.


jobrunner/models.py
```diff
@@ -66,6 +66,7 @@ class StatusCode(Enum):
     CANCELLED_BY_USER = "cancelled_by_user"
     UNMATCHED_PATTERNS = "unmatched_patterns"
     INTERNAL_ERROR = "internal_error"
+    KILLED_BY_ADMIN = "Killed by admin"
```
bloodearnest, line 69: oh nice. Can you add this is another final state that all jobs can go to in the mermaid diagram in `DEVELOPERS.md`? It should be c/p of INTERNAL_ERROR bits, basically

jobrunner/cli/kill_job.py
```diff
@@ -15,12 +15,12 @@ def main(partial_job_ids, cleanup=False):
         # If the job has been previously killed we don't want to overwrite the
         # timestamps here
         if job.state in (State.PENDING, State.RUNNING):
-            mark_job_as_failed(job, "Killed by admin")
+            mark_job_as_failed(job, StatusCode.KILLED_BY_ADMIN)
```
bloodearnest, line 18:
```suggestion
            mark_job_as_failed(job, StatusCode.KILLED_BY_ADMIN, "An OpenSAFELY admin manually killed this job", error=True)
```

I've realised that mark_job_as_failed should implicitly set error=True, but it deson't atm, so you'll need to do this to get an accurate error state.

And we should always give the user a helpful english message

bloodearnest (approved):
LGTM, thanks

Left a couple of suggestions

PR #505, opened 2022-11-02
Set test coverage target

lucyb:
This will cause the tests to fail if the code coverage is under this percentage. It's initially been set to match the existing coverage amount, although the hope is that this will increase over time.

Coverage is only run when running `just test` or `just docker/test`. This is because other `just test*` options only run a subset of the tests, so we can't get a useful result from the coverage tool. However, since `just docker/test` is run in CI, it will always run for every code change.

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -85,4 +85,6 @@ services:
         service: dev
     container_name: job-runner-test
     # override command
-    command: /opt/venv/bin/pytest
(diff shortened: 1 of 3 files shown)


bloodearnest (approved):
Yay!

lucyb:
> Yay!

Thank you. I started to write the ticket for this work yesterday, but then realised it would be easier to do it than write about it

PR #513, opened 2022-11-08
feat: send tick traces in stats thread

### LINKED ISSUE #501: Add more granular job traces
The current job traces only really work for historic state. Whilst they do send an `ENTER $STATE` message, which is useful when looking at an individual state, it's not much use when asking the question "what is the state *right now*?".

Instead, following the guidelines [here](https://www.honeycomb.io/blog/ask-miss-o11y-long-running-requests), we should add a "tick" trace every ~60s or so.

This trick trace would start when the last last tick trace ended, and run till the current time. It would record a span denoting the current state for every job currently not in a final state - every span would have the same duration, the "tick" duration.

This would give us up date date information on the current state of all active jobs, at a 60s granularity.

The old job trace is still useful in the context of looking at a single job (And we could clean it up by dropping `ENTER $STATE` entries from the job trace). 

But the dashboards can use the tick traces with the CONCURRENCY visualisation to track actual current load.

bloodearnest:
Fixes #501

Hopefuly this should give us more up-to-date current job state.

### DIFF
--- a/jobrunner/record_stats.py
+++ b/jobrunner/record_stats.py
@@ -8,7 +8,10 @@ import sqlite3
 import sys
 import time
 
-from jobrunner import config
+from opentelemetry import trace
+
+from jobrunner import config, models, tracing
+from jobrunner.lib import database
 from jobrunner.lib.docker_stats import (
     get_container_stats,
     get_volume_and_container_sizes,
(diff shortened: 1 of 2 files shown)


jobrunner/record_stats.py
```diff
@@ -34,7 +37,11 @@ def main():
         return
     log.info(f"Logging system stats to: {database_file}")
     connection = get_database_connection(database_file)
+    last_run = None
```
lucyb, line 40: It looks like last_run isn't being set to something other than None, so record_tick_trace isn't going to get called?

lucyb (changes requested):

jobrunner/record_stats.py
bloodearnest: Thanks, good catch. A missing initialisation clause, fix incoming

PR #514, opened 2022-11-09
 fix: kill_job preserves job logs

### LINKED ISSUE #511: kill_job should copy docker container logs like a normal exit.
kill_job cli command when run with --cleanup will remove the docker containers and volumes, which is usually what we want.

But we should preserve the logs first, like normal job failure.

bloodearnest:
Fixes #511

### DIFF
--- a/jobrunner/cli/kill_job.py
+++ b/jobrunner/cli/kill_job.py
@@ -3,17 +3,20 @@ Ops utility for killing jobs and cleaning up containers and volumes
 """
 import argparse
 
-from jobrunner.executors.local import container_name, docker, volume_api
-from jobrunner.lib.database import find_where
+from jobrunner.executors import local
+from jobrunner.lib import database, docker
 from jobrunner.models import Job, State, StatusCode
-from jobrunner.run import mark_job_as_failed
+from jobrunner.run import job_to_job_definition, mark_job_as_failed
 
 
 def main(partial_job_ids, cleanup=False):
     jobs = get_jobs(partial_job_ids)
+    #
+    #
     for job in jobs:
         # If the job has been previously killed we don't want to overwrite the
         # timestamps here
+        container = local.container_name(job)
         if job.state in (State.PENDING, State.RUNNING):
             mark_job_as_failed(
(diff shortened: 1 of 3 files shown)


jobrunner/cli/kill_job.py
```diff
@@ -3,35 +3,51 @@
...
-        docker.kill(container_name(job))
+        docker.kill(container)
+
+        # save the logs
+        container_metadata = docker.container_inspect(
+            container, none_if_not_exists=True
+        )
+        # this will write the logs and metadata for this job, but with no
+        # output copying as we pass empty dict
+        local.persist_outputs(
```
evansd, line 35: I don't think we can use `persist_outputs` here because it modifies the workspace by updating log files and we need to leave those untouched. It's a similar to the issue we had in #284.

We might need to refactor `persist_outputs` so we can call this bit with invoking the rest of it:
https://github.com/opensafely-core/job-runner/blob/f9bed192b9a6ab44efd5fc956047b7f16c68acfd/jobrunner/executors/local.py#L325-L329

evansd (changes requested):

jobrunner/cli/kill_job.py
bloodearnest: OH, right, that makes sense.

I'll do the refactor.

bloodearnest: which log files do we need to leave untouched? I assume its the `$workspace/metadata/$action.log`? 

Why does this need to remain untouched?  Feels like it should be the log of latest run, even if that run was killed by an admin?

bloodearnest: That reference bug/PR fixes the pre-executor refactor, so I don't think that applies anymore.

AFAICT, `persist_outputs` no longer has any affect on metadata about outputs at all. It just writes and copies files around.

bloodearnest: I have addressed this issue I think, if you can take another look

evansd (approved):
:+1: Yep, looks good

PR #527, opened 2022-12-05
fix: retry on docker inpsect timeouts

### LINKED ISSUE #453: Gracefully handle `docker inspect` timeout
We periodically see the docker service get stuck responding to `docker inspect` here:

https://github.com/opensafely-core/job-runner/blob/master/jobrunner/executors/local.py#L188

This command timesout after the default of 300s, and then fails the job with internal error.

This is a good proportion (maybe around 50%?) of "Internal Errors" that the user and tech-support see, and it shouldn't be. It also means the main run loop of job-server is blocked for 300s, which is also bad. 

This particular `inspect` call is something we do to every job on every loop, and failing with a timeout it is not fatal (unlike say failing to get outputs, or find the container).

I think we should:

1) set a much shorted timeout, say 30s 
2) If it times out, *log it*, but raise an error that will not result in a job failure with internal error, but rather, the job will be skipped this run of the loop, and retried on the next run (as per normal for non-fatal errors).

This should allow us to ride out when docker has a hitch without failing a bunch of jobs.

However, this could result in a docker hang up getting jobs stuck. But such a case seems fairly catastrophic anyway, and would presumably result in many other failures. We could potentially handle this (e.g. tracking the number of timeouts and hard erroring if they reach a threshold, or similar schemes), but I'd be tempted to try it first.

bloodearnest:
Fixes #453

This adds a tuneable config for how many retries we allow before
failure.

### DIFF
--- a/jobrunner/executors/local.py
+++ b/jobrunner/executors/local.py
@@ -12,6 +12,7 @@ from jobrunner import config
 from jobrunner.executors.volumes import copy_file, get_volume_api
 from jobrunner.job_executor import (
     ExecutorAPI,
+    ExecutorRetry,
     ExecutorState,
     JobDefinition,
     JobResults,
@@ -187,9 +188,15 @@ class LocalDockerAPI(ExecutorAPI):
 
     def get_status(self, job):
         name = container_name(job)
-        job_running = docker.container_inspect(
-            name, "State.Running", none_if_not_exists=True
-        )
+        try:
+            job_running = docker.container_inspect(
+                name,
+                "State.Running",
+                none_if_not_exists=True,
+                timeout=10,
+            )
+        except docker.DockerTimeoutError:
+            raise ExecutorRetry("timed out inspecting container {name}")
 
         if job_running is None:
             # no volume for this job found
--- a/jobrunner/run.py
+++ b/jobrunner/run.py
@@ -14,6 +14,7 @@ from jobrunner import config, tracing
 from jobrunner.executors import get_executor_api
 from jobrunner.job_executor import (
     ExecutorAPI,
+    ExecutorRetry,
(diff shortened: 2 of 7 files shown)


jobrunner/executors/local.py
```diff
@@ -187,9 +188,15 @@ def cleanup(self, job):
...
         name = container_name(job)
-        job_running = docker.container_inspect(
-            name, "State.Running", none_if_not_exists=True
-        )
+        try:
+            job_running = docker.container_inspect(
+                name,
+                "State.Running",
+                none_if_not_exists=True,
+                timeout=10,
```
lucyb, line 196: Why 10?

jobrunner/run.py
```diff
@@ -188,7 +196,19 @@ def handle_job(job, api, mode=None, paused=None):
         )
         return
 
-    initial_status = api.get_status(definition)
+    try:
```
lucyb, line 199: Not one for this change, but `handle_job` is getting quite long. It took a while for me to read through and work out what was happening here. Would be worth thinking about how this function could be refactored in future.

lucyb (approved):
Looks good.

evansd (approved):
+1 I think this will be a big win in terms of operational stability under load. Thanks for taking it on.

jobrunner/run.py
bloodearnest: That's exactly what my re-thinking scheduling doc suggests :)

Basically, split the logic between two loops/functions, one that handles PENDING and one that handles RUNNING.

bloodearnest: But also, large functions are often unavoidable in schedulers, IME, as its essentially a state transition engine, and those often don't decompose well.

PR #530, opened 2022-12-12
loop metrics

bloodearnest:
- get_spans() test helper can now filter by tracer name
- attempt to reducing chance of flaky oomkiller test
- chore: add loop tracing

### DIFF
--- a/jobrunner/run.py
+++ b/jobrunner/run.py
@@ -10,6 +10,8 @@ import sys
 import time
 from typing import Optional
 
+from opentelemetry import trace
+
 from jobrunner import config, tracing
 from jobrunner.executors import get_executor_api
 from jobrunner.job_executor import (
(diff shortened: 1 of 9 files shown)


jobrunner/run.py
```diff
@@ -150,6 +153,29 @@ def handle_single_job(job, api):
         raise
 
 
+def trace_handle_job(job, api, mode, paused):
```
lucyb, line 156: Just a thought, but would this be better written as a decorator that could be applied to `handle_job`?

I've been thinking that the tracer code is normally a good candidate for a decorator, but it's the first time where I've thought it could be nicely applied.

lucyb (approved):
Looks good. One thought, but doesn't actually need a change.

I'm really interested to see what the loop tracing will show. I feel that the current "job", "tick" and "loop" traces could be better named, but don't have any suggestions for how to improve it, I'll probably keep thinking on it though.

jobrunner/run.py
bloodearnest: The thing is, it's specific to the handle_job functionality. So a decorator would not be reusable, in which case, a function is similar.

The only reason its a function is a) we have two call sites, due to the synchronous logic and b) adding it handle_job would have meant indenting everything in a large function to be in a with block, which would have been a bad diff and just a bit awkward.

PR #541, opened 2022-12-23
fix: change JobStatus.timestamp to ns precision

bloodearnest:
Using seconds meant that the  truncation to nearest second made it
possible for telemetry timestamps to move backwards in time which causes
all kinds of complexity.

This changes the semantics of JobStatus.timestamp to be integer
nanoseconds, which should mean it never goes backwards. To support this
for the PREPARED state, we write the ns timestamp into as the contents
of the .opensafely-timestamp, so we can read them out later.

Note, this doesn't change the find new files logic, which still uses the
filesystem mtime metadata for finding new files.

Also, the parsing of timestamps from docker was just completely busted,
so this fixed that and added more tests.

### DIFF
--- a/jobrunner/executors/local.py
+++ b/jobrunner/executors/local.py
@@ -19,7 +19,7 @@ from jobrunner.job_executor import (
     JobStatus,
     Privacy,
 )
-from jobrunner.lib import docker
+from jobrunner.lib import datestr_to_ns_timestamp, docker
 from jobrunner.lib.git import checkout_commit
 from jobrunner.lib.path_utils import list_dir_with_ignore_patterns
 from jobrunner.lib.string_utils import tabulate
@@ -55,14 +55,6 @@ def get_medium_privacy_workspace(workspace):
     if config.MEDIUM_PRIVACY_WORKSPACES_DIR:
         return config.MEDIUM_PRIVACY_WORKSPACES_DIR / workspace
     else:
-        return None
-
-
-def timestamp_from_iso(iso):
-    """Attempt to convert iso formatted date to unix timestamp."""
-    try:
-        return int(datetime.fromisoformat(iso))
-    except Exception:
         return None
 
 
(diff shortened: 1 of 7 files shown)


jobrunner/executors/local.py
```diff
@@ -325,14 +317,13 @@ def finalize_job(job):
...
         message=message,
-        timestamp=int(time.time()),
+        timestamp=time.time_ns(),
         action_version=labels.get("org.opencontainers.image.version", "unknown"),
         action_revision=labels.get("org.opencontainers.image.revision", "unknown"),
         action_created=labels.get("org.opencontainers.image.created", "unknown"),
         base_revision=labels.get("org.opensafely.base.vcs-ref", "unknown"),
         base_created=labels.get("org.opencontainers.base.build-date", "unknown"),
     )
-    job.completed_at = int(time.time())
```
bloodearnest, line 335: This was left over from previous code - JobDefinition has no completed_at field.

jobrunner/executors/volumes.py
```diff
@@ -108,18 +113,23 @@ def copy_from_volume(job, src, dst, timeout=None):
...
-    def file_timestamp(job, path, timeout=None):
+    def read_timestamp(job, path, timeout=None):
         abs_path = host_volume_path(job) / path
         try:
-            stat = abs_path.stat()
+            contents = abs_path.read_text()
+            if contents:
+                return int(contents)
+            else:
+                # linx host filesystem provides untruncated timestamps
```
madwort, line 126:
```suggestion
                # linux host filesystem provides untruncated timestamps
```

jobrunner/job_executor.py
```diff
@@ -49,7 +49,7 @@ class ExecutorState(Enum):
 class JobStatus:
     state: ExecutorState
     message: Optional[str] = None
-    timestamp: int = None
+    timestamp: int = None  # timestamp this JobStatus occored, in integer nanoseconds
```
madwort, line 52:
```suggestion
    timestamp: int = None  # timestamp this JobStatus occurred, in integer nanoseconds
```

jobrunner/lib/docker.py
```diff
@@ -181,47 +181,74 @@ def copy_to_volume(volume_name, source, dest, timeout=None):
...
+            return int(output)
+        except ValueError:
+            # could not convert to integer
+            logger.debug(
+                f"Could not parse int from {volume_name}:{path}: {response.stdout.strip()}"
+            )
 
-def file_timestamp(volume_name, path, timeout=None):
+    # either output was "" or we couldn't parse it as integer
+    # fallback to filesystem metadata, to as support older volumes, and just be
```
madwort, line 220:
```suggestion
    # fallback to filesystem metadata, to support older volumes, and just be
```

madwort (approved):

PR #548, opened 2023-01-06
fix: don't error the job on ExecutorRetry

### LINKED ISSUE #546: Revisit docker timeout retry handling in job-runner
When docker server is timing out, its usually for a good while, and the current retry count then error approach is not avoiding the INTERNAL_ERRORS we'd hoped to with this change.

Much discussion in this thread https://bennettoxford.slack.com/archives/C03ELD5FEAU/p1671702901142859

Ideally, we don't want to fail the job at all, but keep it waiting until docker recovers. But without an INTERNAL_ERROR, we don't have any alerting that its happening.

One options is to remove the retry count/errorring, and instead ensure the timeout is tracked in the tick traces we send to honeycomb. Then we can add an alert like "where count(tick spans with attribute docker_timeout=True) > threshold"

bloodearnest:
Instead, record that it happened, but don't fail the job. The idea is
that we can then be alerted to it happening, but when it's over, the job
continues.

Fixes #546

### DIFF
--- a/jobrunner/executors/local.py
+++ b/jobrunner/executors/local.py
@@ -194,16 +194,18 @@ class LocalDockerAPI(ExecutorAPI):
         RESULTS.pop(job.id, None)
         return JobStatus(ExecutorState.UNKNOWN)
 
-    def get_status(self, job):
+    def get_status(self, job, timeout=10):
         name = container_name(job)
         try:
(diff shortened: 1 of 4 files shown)


bloodearnest:
bad branch

PR #550, opened 2023-01-06
trace async state changes

bloodearnest:
- ci: add some sanity assertions to tracing timestamp code.
- fix: add missings state trace spans with timings

### DIFF
--- a/jobrunner/lib/__init__.py
+++ b/jobrunner/lib/__init__.py
@@ -1,4 +1,6 @@
+import functools
 import secrets
+import warnings
 from contextlib import contextmanager
 from datetime import datetime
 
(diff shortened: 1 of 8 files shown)


jobrunner/lib/__init__.py
```diff
@@ -57,3 +59,31 @@ def datestr_to_ns_timestamp(datestr):
     ts += ns
 
     return ts
+
+
+def warn_assertions(f):
+    """Helper decorator to catch assertions erros and emit as warnings.
```
madwort, line 65:
```suggestion
    """Helper decorator to catch assertions errors and emit as warnings.
```

jobrunner/tracing.py
```diff
@@ -48,6 +49,7 @@ def setup_default_tracing():
         add_exporter(ConsoleSpanExporter())
 
 
+@warn_assertions
```
madwort, line 52: stupid question: do we need `warn_assertions` here, if we have it on `record_job_span`?

tests/test_run.py
```diff
@@ -9,6 +11,85 @@
...
 
+@pytest.mark.parametrize(
+    "synchronous_transitions",
+    [
+        [],
+        [ExecutorState.PREPARING, ExecutorState.FINALIZING],
+    ],
+)
+def test_handle_job_full_execution(synchronous_transitions, db, freezer):
+    # move to a second boundary so rounding doesn't affect our times.
```
madwort, line 22: wondering if there's a way of wording this that doesn't make me initially wonder where the first boundary is 😆

bloodearnest: I'm sure there is :)

bloodearnest:
```suggestion
    # move to a whole second boundary for easier timestamp maths
```

PR #551, opened 2023-01-11
fix: stop job traces picking up loop trace as parent

bloodearnest:
This just means giving it an explicit empty context.

### DIFF
--- a/tests/test_tracing.py
+++ b/tests/test_tracing.py
@@ -110,7 +110,7 @@ def test_trace_attributes_missing(db, monkeypatch):
     )
 
 
-def test_initialise_trace(db):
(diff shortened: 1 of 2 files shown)


tests/test_tracing.py
```diff
@@ -110,7 +110,7 @@ def test_trace_attributes_missing(db, monkeypatch):
     )
 
 
-def test_initialise_trace(db):
+def test_initialise_trace1(db):
```
madwort, line 113:
```suggestion
def test_initialise_trace(db):
```
looks like a wip that accidentally made it thru

madwort (approved):

PR #552, opened 2023-01-12
fix timings

bloodearnest:
- fix: ensure int timestamp in negative duration handling
- refactor: subsume set_state into set_code
- feat: make run loop handle synchronous operations

### DIFF
--- a/jobrunner/run.py
+++ b/jobrunner/run.py
@@ -22,6 +22,7 @@ from jobrunner.job_executor import (
     Privacy,
     Study,
 )
+from jobrunner.lib import ns_timestamp_to_datetime
 from jobrunner.lib.database import find_where, select_values, update
 from jobrunner.lib.log_utils import configure_logging, set_log_context
 from jobrunner.models import FINAL_STATUS_CODES, Job, State, StatusCode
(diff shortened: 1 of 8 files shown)


tests/test_run.py
```diff
@@ -90,6 +82,80 @@ def test_handle_job_full_execution(synchronous_transitions, db, freezer):
...
 
 
+def test_handle_job_full_execution_synchronous(db, freezer):
+    # move to a whole second boundary for easier timestamp maths
+    freezer.move_to("2022-01-01T12:34:56")
+
+    api = StubExecutorAPI()
+    api.synchronous_transitions = [ExecutorState.PREPARING, ExecutorState.FINALIZING]
+
+    start = int(time.time() * 1e9)
```
madwort, line 92: any reason not to use `time_ns`?
```suggestion
    start = int(time.time_ns())
```

jobrunner/run.py
```diff
@@ -550,20 +533,42 @@ def set_code(job, code, message, error=None, results=None, timestamp_ns=None, **
...
-
-    if job.status_code_updated_at > timestamp_ns:
-        # we somehow have a negative duration, which honeycomb does funny things with.
-        # This can happen in tests, where things are fast, but we've seen it in production too.
-        log.warning(
-            f"negative state duration, clamping to 1ms ({job.status_code_updated_at} > {timestamp_ns})"
-        )
-        timestamp_ns = job.status_code_updated_at + 1e6  # set duration to 1ms
-        timestamp_s = int(timestamp_ns // 1e9)
+        timestamp_s = int(timestamp_ns / 1e9)
```
madwort, line 536: double-checking this should be `/` and not `//` 
```suggestion
        timestamp_s = int(timestamp_ns // 1e9)
```

madwort (approved):

tests/test_run.py
bloodearnest: nope, I'm slowly moveing to time_ns(), I didn't know about it when I originally wrote this.

jobrunner/run.py
bloodearnest: yep, agreed

PR #554, opened 2023-01-16
chore: get the correct volume api for a job

### LINKED ISSUE #522: Finish and rollout local volume api in TPP
The copying of files in and out of jobs workspace is slow. Also, it currently blocking, which means that in blocks the loop for a long time. 

In #433 we implemented a volume_api abstraction as part of shore leave, and added a bind mount implementation that is much faster, but only works on linux. The reason why its faster is its just coping on the host, not via docker cp, which a) goes over the network and b) uses the docker api, which can fail under load.

This was implemented as a run time config - we're using one api or the other.

However, when coming to try roll it out in prod, we realise that we needed to handle both at the same time, as jobs that were started with the old api will need use the same api to finalise them.

To have a smooth rollout, we need to select the api to use per job, rather than at boot time. That will allow use to switch to the bind mount api and still correctly handle in-flight jobs. And vice versa for rolling back.

bloodearnest:
This means that a job that was PREPARED with one volume API will be
finalized with the same volume API. This will facilitate switching it on
bind mount API in production.

Fixes #522

### DIFF
--- a/jobrunner/executors/volumes.py
+++ b/jobrunner/executors/volumes.py
@@ -64,9 +64,13 @@ class DockerVolumeAPI:
         return docker.find_newer_files(docker_volume_name(job), path)
 
 
-def host_volume_path(job):
+def host_volume_path(job, create=True):
     path = config.HIGH_PRIVACY_VOLUME_DIR / job.id
-    path.parent.mkdir(parents=True, exist_ok=True)
+    if create:
+        try:
+            path.parent.mkdir(parents=True, exist_ok=True)
+        except PermissionError:
+            raise Exception(f"Could not create {path.parent} due to permissions error")
     return path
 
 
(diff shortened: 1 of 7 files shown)


jobrunner/executors/volumes.py
```diff
@@ -154,7 +160,18 @@ def find_newer_files(job, reference):
...
+
+DEFAULT_VOLUME_API = default_volume_api()
+
+
+def get_volume_api(job):
+    for api in [BindMountVolumeAPI, DockerVolumeAPI]:
+        if api.volume_exists(job):
+            return api
+
+    return DEFAULT_VOLUME_API
```
evansd, line 177: Clever! So in theory we can just switch the default, new jobs start using the new API and any existing jobs continue with the old one.

evansd (approved):

jobrunner/executors/volumes.py
bloodearnest: Zactly

PR #556, opened 2023-01-19
feat: instrument assert_new_jobs_created

madwort:
* experimental instrumentation to explore what this block is doing in production
 * to be reverted if trace volumes get out of hand and/or investigation is complete

### DIFF
--- a/jobrunner/create_or_update_jobs.py
+++ b/jobrunner/create_or_update_jobs.py
@@ -283,7 +283,6 @@ def assert_new_jobs_created(new_jobs, current_jobs):
     tracer = trace.get_tracer("new_jobs")
     with tracer.start_as_current_span("assert_new_jobs_created") as span:
 
-        span = trace.get_current_span()
         if new_jobs:
             span.set_attribute("new_jobs", True)
             # want to have access to the job id that is causing these new jobs
--- a/jobrunner/create_or_update_jobs.py
+++ b/jobrunner/create_or_update_jobs.py
@@ -10,6 +10,7 @@ import logging
 import re
 import time
 
+from opentelemetry import trace
 from pipeline import RUN_ALL_COMMAND, ProjectValidationError, load_pipeline
 
 from jobrunner import config, tracing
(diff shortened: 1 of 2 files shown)


jobrunner/create_or_update_jobs.py
```diff
@@ -279,33 +280,61 @@ def job_should_be_rerun(job_request, job):
...
 def assert_new_jobs_created(new_jobs, current_jobs):
-    if new_jobs:
-        return
-
-    pending = [job.action for job in current_jobs if job.state == State.PENDING]
-    running = [job.action for job in current_jobs if job.state == State.RUNNING]
-    succeeded = [job.action for job in current_jobs if job.state == State.SUCCEEDED]
-    failed = [job.action for job in current_jobs if job.state == State.FAILED]
+    tracer = trace.get_tracer("new_jobs")
+    with tracer.start_as_current_span("assert_new_jobs_created"):
```
bloodearnest, line 284:
```suggestion
    with tracer.start_as_current_span("assert_new_jobs_created") as span:
```

jobrunner/create_or_update_jobs.py
```diff
@@ -279,33 +280,61 @@ def job_should_be_rerun(job_request, job):
...
+
+        span = trace.get_current_span()
```
bloodearnest, line 286:
```suggestion

```

jobrunner/create_or_update_jobs.py
```diff
@@ -279,33 +280,61 @@ def job_should_be_rerun(job_request, job):
...
+        if new_jobs:
+            span.set_attribute("new_jobs", True)
```
bloodearnest, line 288:
```suggestion

```

bloodearnest: This is needed - the fact we're using tracer called new_jobs is enough, scope = new_jobs will find this.

jobrunner/create_or_update_jobs.py
```diff
@@ -279,33 +280,61 @@ def job_should_be_rerun(job_request, job):
...
+            # want to have access to the job id that is causing these new jobs
+            # span.set_attribute("job_id", job.id)
+            span.set_attribute(
+                "new_job_actions", ", ".join(f"{job.action}" for job in new_jobs)
```
bloodearnest, line 292:
```suggestion
                "actions", ", ".join(f"{job.action}" for job in new_jobs)
```
We already new these are new jobs, we can just call them actions?

bloodearnest (approved):
lgtm!

Issue #535, opened 2022-12-15
Add Honeycomb metrics to containers on the backend


lucyb:
@bloodearnest to add more details into this ticket

bloodearnest:
We already collect metrics periodically in a separate thread, and write them to a sqllite db

In the same thread, we also send the tick traces that tell us the current state of play.

We need to join these up, and add any current metrics for a running job to the tick trace spans.


Caveat: we currently use the `docker stats` api to get metrics. If possible, it would be nice to avoid using this, as it's not cheap. There are ways to get this information without hitting the docker api

madwort:
can we add "create links from job server to the new metrics" to this ticket?

bloodearnest:
Done in https://github.com/opensafely-core/job-server/pull/2530

PR #558, opened 2023-01-19
feat: track memory usage in tick traces.

bloodearnest:
This replaces the old logging to a local sqlite db with adding that
memory and cpu info to the tick traces.

Volume/container disk sizes are much trickier, so have been removed for
now. We'll add them back in once we've switched to BindMount API in
prod.


bloodearnest:
Fixes #535

madwort (approved):
excited to see this in production!

PR #559, opened 2023-01-23
feat: record the high water mark for cpu/memory for a job

bloodearnest:
We don't do anything with it right now, but maybe in future we will,
e.g. displaying it to users, or using it to schedule jobs better.

There's a bit of wrinkle here though: we're kinda of breaking the
scheduler/executor layering a bit, by storing these values directly in
the job-runner db.

Ideally, the executor would be responsible for tracking resource usage,
both for live metrics and for high-water-marks. But the current executor
was not designed with this in mind.

So this takes the easy option, and makes the current record_stats thread
handle it. With the caveat that its specific to the local docker
executor.

### DIFF
--- a/jobrunner/models.py
+++ b/jobrunner/models.py
@@ -158,6 +158,9 @@ class Job:
             completed_at INT,
             trace_context TEXT,
             status_code_updated_at INT,
+            -- resource high water marks --
+            cpu_hwm INT,
+            memory_hwm INT,
 
             PRIMARY KEY (id)
         );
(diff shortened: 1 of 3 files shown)


madwort (approved):

bloodearnest:
Am planning to abandon this approach, i.e. storing stuff in job-runners. It's icky.

A better approach would be to forward the data from the otel-collector on dokku4 to *somewhere* we can persist and query it.  That could be job-server, but probably better to use either a SaaS (I'm trying grafana cloud) or COTS component (e.g. jaeger)

PR #564, opened 2023-01-27
fix: more careful tempfile handling

bloodearnest:
This has been tested to work on windows.


jobrunner/executors/volumes.py
```diff
@@ -49,11 +49,14 @@ def delete_volume(job):
...
-            f.write(str(time.time_ns()))
-            f.flush()
+        try:
+            f = tempfile.NamedTemporaryFile(delete=False)
+            f.close()
             p = Path(f.name)
+            p.write_text(str(time.time_ns()))
             docker.copy_to_volume(docker_volume_name(job), p, path, timeout)
+        finally:
+            p.unlink(missing_ok=True)
```
evansd, line 59: If something fails on either of these lines (which isn't impossible) then `p` won't be set:
```py
f = tempfile.NamedTemporaryFile(delete=False)
f.close()
```

Is it worth setting `p = None` before entering the block and then having an `if p is not None` guard on the `finally` clause?

evansd (approved):

jobrunner/executors/volumes.py
bloodearnest: ok, I've avoided relying on p at all.

also, I vaguely recall os.remove works better in windows, so I switched to that

evansd (approved):
:+1: That looks pretty solid

PR #576, opened 2023-02-17
feat: preserve logs of user cancelled jobs [VERY WIP]

madwort:


tests/factories.py
```diff
@@ -226,6 +226,14 @@ def finalize(self, definition):
             expected = ExecutorState.FINALIZING
         return self.do_transition(definition, ExecutorState.EXECUTED, expected)
 
+    def finalize_job(self, definition):
```
madwort, line 229: note to self: don't do this!

PR #579, opened 2023-02-22
fix: run as non-root user with BindMountVolumeAPI

bloodearnest:
Current, delete_volume() doesn't work properly with the
BindMountVolumeAPI. Because the containers are run as root, files they
create are owned by root, and this prevents jobrunner from deleting
them, and we currently manually have to delete them.

This change runs jobs with the jobrunner uid/gid when using
BindMountVolumeAPI, which solves the problem.

I have tested that all our current images work when run as non-root with
some sample studies, so I think this is good to go.

### DIFF
--- a/jobrunner/config.py
+++ b/jobrunner/config.py
@@ -260,3 +260,9 @@ HIGH_PRIVACY_VOLUME_DIR = HIGH_PRIVACY_STORAGE_BASE / "volumes"
 # needs to point to the path to the HIGH_PRIVACY_VOLUME_DIR from the *hosts*
 # perspective, as that's what docker will be looking for.
 DOCKER_HOST_VOLUME_DIR = os.environ.get("DOCKER_HOST_VOLUME_DIR")
+
+# These are currently only used with the BindMountVolumeAPI.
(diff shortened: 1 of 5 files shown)


ghickman (approved):

jobrunner/config.py
```diff
@@ -255,7 +255,10 @@ def parse_job_resource_weights(config_file):
 # BindMountVolumeAPI config
 #
 # used to store directories to be mounted into jobs with the BindMountVolumeAPI
-HIGH_PRIVACY_VOLUME_DIR = HIGH_PRIVACY_STORAGE_BASE / "volumes"
+HIGH_PRIVACY_VOLUME_DIR = os.environ.get(
+    "HIGH_PRIVACY_VOLUME_DIR",
+    HIGH_PRIVACY_STORAGE_BASE / "volumes",
+)
```
evansd, line 261: It would be worth wrapping this in `Path()` otherwise you'll get a `Path` as the default and then a string if the env var is set which is liable to cause problems.

evansd (approved):
It's so good we've finally got here. Just one small suggestion below.

PR #588, opened 2023-03-10
chore: enhance tests

madwort:
* clarify some states
* rename some misnamed vars
* add some more scenarios

### DIFF
--- a/tests/test_local_executor.py
+++ b/tests/test_local_executor.py
@@ -385,7 +385,8 @@ def test_finalize_failed_137(docker_cleanup, job, tmp_work_dir, volume_api):
     # slightly strange that this is EXECUTED and not ERROR
     wait_for_state(api, job, ExecutorState.EXECUTED)
 
-    # This does run finalize, because the cancelled bit is not set
+    # In contrast to test_running_job_terminated_finalized , finalize would be run
+    # if a job is externally killed rather than being terminated
     status = api.finalize(job)
     assert status.state == ExecutorState.FINALIZED
 
--- a/tests/test_local_executor.py
+++ b/tests/test_local_executor.py
@@ -456,7 +456,7 @@ def test_pending_job_terminated_not_finalized(docker_cleanup, job, tmp_work_dir)
 
 
 @pytest.mark.needs_docker
-def test_running_job_cancelled_finalized(docker_cleanup, job, tmp_work_dir):
+def test_running_job_terminated_finalized(docker_cleanup, job, tmp_work_dir):
     job.args = ["sleep", "101"]
 
     api = local.LocalDockerAPI()
--- a/tests/test_local_executor.py
+++ b/tests/test_local_executor.py
@@ -382,7 +382,8 @@ def test_finalize_failed_137(docker_cleanup, job, tmp_work_dir, volume_api):
     # impersonate an admin
     docker.kill(local.container_name(job))
 
-    # slightly strange that this is EXECUTED and not ERROR
+    # status EXECUTED is in contrast to test_running_job_terminated_finalized,
+    # which produces status ERROR in similar circumstances.
     wait_for_state(api, job, ExecutorState.EXECUTED)
 
     # In contrast to test_running_job_terminated_finalized , finalize would be run
--- a/tests/test_local_executor.py
+++ b/tests/test_local_executor.py
@@ -20,6 +20,7 @@ def volume_api(request, monkeypatch):
 
 @pytest.fixture
 def job(request, test_repo):
+    # TODO: this is a "JobDefinition", not a "Job"
     """Basic simple action with no inputs as base for testing."""
     if "needs_docker" in list(m.name for m in request.node.iter_markers()):
         ensure_docker_images_present("busybox")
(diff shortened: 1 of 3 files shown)


tests/test_local_executor.py
```diff
@@ -360,8 +382,10 @@ def test_finalize_failed_137(docker_cleanup, job, tmp_work_dir, volume_api):
     # impersonate an admin
     docker.kill(local.container_name(job))
 
+    # slightly strange that this is EXECUTED and not ERROR
     wait_for_state(api, job, ExecutorState.EXECUTED)
```
bloodearnest, line 386: ExecutoreState.ERROR means something went wrong with the **executor**, not with the job. From the executor's POV, the job has finished with an exit code, just like normal.

tests/test_local_executor.py
```diff
@@ -360,8 +382,10 @@ def test_finalize_failed_137(docker_cleanup, job, tmp_work_dir, volume_api):
...
 
+    # This does run finalize, because the cancelled bit is not set
```
bloodearnest, line 388: This comment seems a bit contextless in this test case?

tests/test_local_executor.py
```diff
@@ -401,6 +427,68 @@ def test_finalize_failed_oomkilled(docker_cleanup, job, tmp_work_dir):
         == "Ran out of memory (limit for this job was 0.01GB)"
     )
 
+    assert workspace_log_file_exists(job)
+
+
+@pytest.mark.needs_docker
+def test_pending_job_cancelled_not_finalized(docker_cleanup, job, tmp_work_dir):
```
bloodearnest, line 434:
```suggestion
def test_pending_job_terminated_not_finalized(docker_cleanup, job, tmp_work_dir):
```

Cancelled is a run.py concept, not an executor concept, which only has `terminated()`

tests/test_local_executor.py
```diff
@@ -401,6 +427,68 @@ def test_finalize_failed_oomkilled(docker_cleanup, job, tmp_work_dir):
...
+    assert status.state == ExecutorState.UNKNOWN
+    assert api.get_status(job).state == ExecutorState.UNKNOWN
+
+    assert job.id not in local.RESULTS
+
+    assert not workspace_log_file_exists(job)
+
+
+@pytest.mark.needs_docker
+def test_running_job_cancelled_finalized(docker_cleanup, job, tmp_work_dir):
```
bloodearnest, line 459:
```suggestion
def test_running_job_terminated_finalized(docker_cleanup, job, tmp_work_dir):
```

bloodearnest (approved):
LGTM, good catch on the missing test.

I think there might be a slight layering confusion between the scheduler (run.py/handle_job) and the executor, might be worth talking through?

Also, this doesn't add any features, so the commit/PR should be `chore:` or `refactor:`, to avoid needlessly bumping the minor or patch release numbers.

tests/test_local_executor.py
madwort: ah, yes, you're right, I wanted to differentiate from https://github.com/opensafely-core/job-runner/blob/master/tests/test_local_executor.py#L383 but now I look again at that, that's `kill` not `terminate`

tests/test_local_executor.py
madwort: as with the other conversation, was trying to make explicit why this test (`kill`) is different from the other test (now `terminate`), maybe a reword would make that clearer.

madwort:
```suggestion
    # In contrast to test_running_job_terminated_finalized , finalize would be run
    # if a job is externally killed rather than being terminated
```

tests/test_local_executor.py
madwort: but compare with test_running_job_terminated_finalized - a running job that is terminated via `terminate` puts us into ExecutorState.ERROR *not* EXECUTED. https://github.com/opensafely-core/job-runner/pull/588/files#diff-c0a1e0861f681ae3cbc247887b135901340f6501175217aa4903ef7f7af1068bR473 
Why are these two scenarios different?

bloodearnest: Hmm. 

One is the result of docker kill, which simulate a process ending asynchronously external to the API. I.e. we'll only be notified of it next time we call `get_status()`.

The other is the result of an explicit api.terminate() call, which blocks on termination, and we know is done. 

However, I think the current implementation of terminate() may be incorrect, or at the very least misleading.

`terminate()` currently explicitly returns ERROR, but if you ask it just after with api.get_status(), it will depend on what state it was when it was terminated. If it was running, it will be EXECUTED. If it was any other state I think the state will be unchanged.

I suspect terminate() returning ERROR is incorrect, as its not an error really. But it's never really been an issue before, as we generally follow a terminate with cleanup() so its all going way.

Possibly, terminate() should return `self.get_status(job)`, if it's going to be 100% accurate? The only reason to return ERROR would be if the docker kill failed, and that's handled by exceptions anyway. Not sure how much it matters, but would avoid the misleading ERROR. 

It may be we want to reshape some of the terminate/cleanup API as part of this work to ensure finalization anyway.

tests/test_local_executor.py
```diff
@@ -360,8 +382,10 @@ def test_finalize_failed_137(docker_cleanup, job, tmp_work_dir, volume_api):
     # impersonate an admin
     docker.kill(local.container_name(job))
 
+    # slightly strange that this is EXECUTED and not ERROR
```
madwort, line 385:
```suggestion
    # status EXECUTED is in contrast to test_running_job_terminated_finalized, 
    # which produces status ERROR in similar circumstances.
```

PR #589, opened 2023-03-10
feat: preseve logs of user cancelled jobs

madwort:
* make the use of ExecutorStatus.ERROR more consistent between different Executors
* use ExecutorStatus.ERROR in finalize to determine whether execution finished, and therefore whether we should just preserve logs
* there are a couple of reasons to implement this using ExecutorStatus rather than passing a parameter to the ExecutorAPI `finalize()` call, including:
  * jobs whose docker process is killed externally will currently run the finalize process, they currently attempt to store outputs & I think they should instead do the same as a process killed through the cancel process
  * adding parameters to the ExecutorAPI means reworking the [metaprogramming in the LoggingExecutor](https://github.com/opensafely-core/job-runner/blob/002939feb4260734fe3ff47a6fa13e5bd77702be/jobrunner/executors/logging.py#L40) in a very unaesthetic way

### DIFF
--- a/tests/test_local_executor.py
+++ b/tests/test_local_executor.py
@@ -416,9 +416,7 @@ def test_finalize_failed_137(docker_cleanup, job, tmp_work_dir, volume_api):
     # impersonate an admin
     docker.kill(local.container_name(job))
 
-    # status EXECUTED is in contrast to test_running_job_terminated_finalized,
-    # which produces status ERROR in similar circumstances.
-    wait_for_state(api, job, ExecutorState.EXECUTED)
+    wait_for_state(api, job, ExecutorState.ERROR)
 
     # In contrast to test_running_job_terminated_finalized , finalize would be run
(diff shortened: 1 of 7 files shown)


tests/test_local_executor.py
```diff
@@ -507,23 +505,21 @@ def test_running_job_terminated_finalized(docker_cleanup, job, tmp_work_dir):
 
     status = api.terminate(job)
     assert status.state == ExecutorState.ERROR
-    assert api.get_status(job).state == ExecutorState.EXECUTED
+    assert api.get_status(job).state == ExecutorState.ERROR
```
madwort, line 508: if you really don't want to use `ExecutorState.ERROR` in this way, I can probably resurrect the previous iteration that uses the parameterised finalize function (and we just put up with the implementation annoyances)

madwort:
reworked with a tweaked approach in a new PR

PR #591, opened 2023-03-15
chore: integration test for bindmountapi

madwort:
The integration tests currently only run using one of the APIs that we support. We should run the tests on both APIs, which will hopefully catch more bugs more quickly.


madwort:
weird, tests pass on macos (and are def using BindMountAPI), but I just noticed tests fail on linux on CI. will fix.

bloodearnest (approved):

PR #605, opened 2023-04-19
feat: Alias ehrql to databduilder

rebkwok:
This is a temporary measure until the work to rename databuilder to ehrql is complete. It means that users can write ehrql:v0 in their project.yaml and opensafely exec commands now

See also https://github.com/opensafely-core/opensafely-cli/pull/201 (which depends on this PR, and updated vendored code)

### DIFF
--- a/jobrunner/actions.py
+++ b/jobrunner/actions.py
@@ -6,6 +6,7 @@ from typing import Dict, List
 from pipeline.exceptions import ProjectValidationError
 from pipeline.outputs import get_output_dirs
 
+from jobrunner import config
 from jobrunner.lib.path_utils import ensure_unix_path
 
 from .extractors import is_extraction_command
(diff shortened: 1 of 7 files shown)


rebkwok:
> Could we publish the docker image under both names during the transition period? It would save needing the aliasing logic.

So that's what I was suggesting in the docs meeting this morning, but @inglesp said we could easily alias it in jobrunner.  It hasn't turned out to be that easy, so I was wondering about publishing the image as ehrql instead. It does mean that we'd have to remember to pull both on the server any time there's an update.

evansd:
Great minds! :smile: 

> It does mean that we'd have to remember to pull both on the server any time there's an update.

We could probably hack something in here to save us having to remember:
https://github.com/opensafely-core/job-runner/blob/851a5334d1248e346274d5309ca8dcc0152f1e99/scripts/update-docker-image.sh

And longer term we want to move to automatic pulls anyway.

PR #610, opened 2023-05-23
fix: use correct extension for workspace archive

madwort:
* file extension was not updated when compression choice was changed in https://github.com/opensafely-core/job-runner/pull/350
* production files seem to have the correct extension, for unknown reason


madwort:
tests fail because of stata licence

PR #612, opened 2023-06-16
fix: correct bugs in archive script

madwort:
* some simple fixes that must have somehow got lost when this script was created & first used
* it's possible that there are some old archive files that contain gzipped data but have the xz extension - I guess we'll find out if/when we need them. If we encounter a "corrupt" xz file, we should try to remember to experiment with changing the extension to gz...!

### DIFF
--- a/scripts/archive.sh
+++ b/scripts/archive.sh
@@ -3,7 +3,7 @@ set -euo pipefail
 
 workspace=$1
 workspace_dir=$HIGH_PRIVACY_STORAGE_BASE/workspaces/$workspace
-archive=$HIGH_PRIVACY_STORAGE_BASE/archives/$workspace.tar.xz
+archive=/archive/archived_workspaces/$workspace.tar.gz
 tmp_archive=$archive.tmp
 
 if ! test -d "$workspace_dir"; then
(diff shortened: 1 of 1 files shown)


scripts/archive.sh
```diff
@@ -3,7 +3,7 @@ set -euo pipefail
 
 workspace=$1
 workspace_dir=$HIGH_PRIVACY_STORAGE_BASE/workspaces/$workspace
-archive=$HIGH_PRIVACY_STORAGE_BASE/archives/$workspace.tar.xz
+archive=/archive/archived_workspaces/$workspace.tar.gz
```
bloodearnest, line 6: Hmm, I think we should probably chance the mount point from /archive to /src/high_privacy/archives, so it's all under one dir, but that can be a different thing

scripts/archive.sh
```diff
@@ -44,5 +44,5 @@ fi
 echo
 
 rm -r "$workspace_dir"
-# on TPP windows this needs to be docker 
+# on TPP windows this needs to be docker
```
bloodearnest, line 47: I think these two lines can go now, as they we for running it pre-vm migration

bloodearnest (approved):

scripts/archive.sh
madwort: https://github.com/opensafely-core/sysadmin/issues/116

scripts/archive.sh
madwort: https://github.com/opensafely-core/job-runner/pull/614

PR #618, opened 2023-07-14
Allow db access for `ehrql generate-measures`

evansd:
Unlike Cohort Extractor, ehrQL retrieves the results and calculates the measures in a single step and so requires database access to run.

We also clean up unused code relating to long-since vanished Data Builder behaviour. In particular, this allows us to remove `is_extraction_command` entirely which was a potential footgun as it looked like the function for deciding whether a command should have database access, but that wasn't its role.


inglesp (approved):
Nice cleanup too

PR #621, opened 2023-07-14
chore: make explicit some state vars that are ExecutorStates

madwort:
this one feels a bit flat-footed / overly-verbose, but it’s the kind of slightly-more-stuff-that-you-have-to-hold-in-your-head that made it a lot slower for me getting started with this codebase…


ghickman (approved):
Given the complexity we've ended up with from having two systems, this is a great change 👍

PR #640, opened 2023-08-25
Configure required checks

ghickman:
This uses [alls-green](https://github.com/re-actors/alls-green) to set up a job that depends on previous jobs, capturing our matrixed jobs too, so we can configure one required check in the repo settings.


madwort (approved):
yes please

PR #643, opened 2023-08-28
Remove left-over logic from switch to finalizing everything

### LINKED ISSUE #634: Missing docker metrics for some jobs
We emit tick traces every ~10s for jobs running on a backend.

When a job is EXECUTING, we grab stats from docker and add cpu/memory data to those ticks, which we can graph to view individual job's resource usage.

However, sometimes we don't have this data, e.g. https://ui.honeycomb.io/bennett-institute-for-applied-data-science/environments/production/datasets/jobrunner/result/5krnc7PFBjA?omitMissingValues&useStackedGraphs

More evidence: https://ui.honeycomb.io/bennett-institute-for-applied-data-science/environments/production/datasets/jobrunner/result/q45zK21YZEy

We should investigate why some jobs have this data missing.

Relevant code: https://github.com/opensafely-core/job-runner/blob/main/jobrunner/lib/docker_stats.py#L14

⌚ Timebox to half a day

bloodearnest:
ExecutorState.ERROR can be confusing. It is meant to indicate that the
*executor* has errored - a job exiting with an exit code is expected
behaviour, not an *executor* error.

As such, `get_status()` should not be returning ExecutorState.ERROR when
a job is OOM killed - that's business as usual, and should be handled by
`finalise()` as normal.

When we refactored job-runner recently for everything to go via
`finalise()`, I think this behaviour was mistakenly preserved from the
old way of doing things. It resulted in job-runner treating OOM kills
as INTERNAL_ERRORS, and obscuring the message to the users and
needlessly paging teck-support.

Now we only handle the OOM case in one place, I de-factored the
functions that were previously shared between the two locations

Fixes #634


jobrunner/run.py
```diff
@@ -82,6 +82,10 @@ def handle_jobs(api: Optional[ExecutorAPI]):
                 # workspace. This gives a fairer allocation of capacity among
                 # workspaces.
                 running_for_workspace[job.workspace],
+                # prioritise db jobs over cpu jobs, as they a) have their own
+                # limit and b) the system usually benefits from getting these
+                # done before cpu jobs
+                0 if job.requires_db else 1,
```
madwort, line 88: this change seems like it's been pulled in accidentally from another commit?

jobrunner/job_executor.py
```diff
@@ -37,13 +37,21 @@ class JobDefinition:
 
 
 class ExecutorState(Enum):
+    # Job is currently preparing to run: creating volumes,copying files, etc
     PREPARING = "preparing"
+    # Job volume is prepared and ready to run
     PREPARED = "prepared"
+    # Job currently executing
     EXECUTING = "executing"
+    # Job process has finished executing, and has an error code
```
madwort, line 46: `error code` -> `exit code`?

tests/test_local_executor.py
```diff
@@ -422,11 +422,7 @@ def test_finalize_failed_137(docker_cleanup, job_definition, tmp_work_dir, volum
     # impersonate an admin
     docker.kill(local.container_name(job_definition))
 
-    wait_for_state(api, job_definition, ExecutorState.ERROR)
-    assert (
-        api.get_status(job_definition).message
-        == "Job either ran out of memory or was killed by an admin"
-    )
+    wait_for_state(api, job_definition, ExecutorState.EXECUTED)
```
madwort, line 425: yep, this path is not on this diagram https://github.com/opensafely-core/job-runner/blob/main/docs/local_executor_run.md but the `the job finishes with an error` has `ExecutorState.EXECUTED`

madwort (approved):
good fix, a couple of nits to address though

jobrunner/run.py
bloodearnest: Ah, yes, will drop thanks

tests/test_local_executor.py
bloodearnest: I don't understand this comment I'm afraid? Is the diagram wrong now?

madwort: nope all fine, this is just a by the way

madwort: the diagram says this state should have been EXECUTED

bloodearnest:
deployed

PR #648, opened 2023-09-06
Do not raise INTERNAL_ERROR for job resubmission

bloodearnest:
There is nothing we can do here, and no error to deal with


tests/test_create_or_update_jobs.py
```diff
@@ -16,7 +16,7 @@
 from jobrunner.lib.database import find_one, update_where
 from jobrunner.models import Job, JobRequest, State, StatusCode
 from tests.conftest import get_trace
-from tests.factories import job_request_factory_raw
+from tests.factories import job_request_factory, job_request_factory_raw
```
madwort, line 19: linter says no
```
F401 'tests.factories.job_request_factory' imported but unused
```

```suggestion
from tests.factories import job_request_factory_raw
```

madwort (approved):

PR #652, opened 2023-09-11
fix: amend broken urls to moved playbook

madwort:


madwort:
CI failed with unrelated intermittent problem

lucyb (approved):

PR #662, opened 2023-09-28
level 4 limits

bloodearnest:
- Use correct privacy strings in tests
- Limit level 4 file size


evansd (approved):
This looks good to me. In particular, I like the fact that the max size is passed on the JobDefinition so we have the option to customise on a per-job basis further down the road. And I also like the fact that we have a mapping from excluded files to arbitrary error messages so we can introduce new reasons for excluding files later.

bloodearnest:
> This looks good to me. In particular, I like the fact that the max size is passed on the JobDefinition so we have the option to customise on a per-job basis further down the road. 

Very intentional, I see that coming...

>And I also like the fact that we have a mapping from excluded files to arbitrary error messages so we can introduce new reasons for excluding files later.

It's almost like I could see the future... :rofl:

PR #666, opened 2023-10-11
more l4 limits

bloodearnest:
- Rename leve4 file size config
- Add more l4 file limits


evansd (approved):
Looks good. Made a couple of tweaks to the message text but nothing else.

bloodearnest:
> Looks good. Made a couple of tweaks to the message text but nothing else.

Good catch, thanks!

PR #676, opened 2023-11-07
Check for out-of-date codelists and fail the job request if any actions require the database

rebkwok:
Receive the `codelists_ok` flag that job-server now passes with a job request, and use it to fail the entire job request if any of its actions require the database.

I added an `is_database_action` to the Action object in the pipeline library, but job-runner already has a method for checking for actions requiring the database, so I'm just using that.


jobrunner/create_or_update_jobs.py
```diff
@@ -299,6 +304,16 @@ def assert_new_jobs_created(job_request, new_jobs, current_jobs):
...
 
 
+def assert_codelists_ok(job_request, new_jobs):
+    if job_request.codelists_ok:
+        return True
+    for job in new_jobs:
+        # Codelists are out of date; fail the entire job request if any job
+        # requires database access
+        if job.requires_db:
+            raise JobRequestError("Codelists are out of date")
```
evansd, line 314: Is it worth adding any more detail here? I guess the user will have the warning message about out-of-date codelists already at the top of the page, won't they?

evansd (approved):
It's nice how little code this took.

Obviously it will need to hold off being deployed while the job-server side of things is finished off.

jobrunner/create_or_update_jobs.py
rebkwok: My thought was that they'd probably have already seen the out of date codelist message and ignored it - or at least would have seen it enough times before to know what this meant.  But I've assumed that before and been wrong.
Since it fails on a specific job, and we assume that's a dependency that they didn't explicitly request (otherwise job-server would have refused to create the job request in the first place) it could say something like "Codelists are out of date (required by Job X)"?

PR #679, opened 2023-11-15
Fix add_job cli command

bloodearnest:
This was not previously covered in tests, and it broke with the codelist
checking changes.  It's used in the backed-server test suite, which is
how I noticed.

If fixed, and added test coverage, which allowed me to bump the coverge
minimum a bit.

### DIFF
--- a/jobrunner/cli/add_job.py
+++ b/jobrunner/cli/add_job.py
@@ -5,6 +5,7 @@ job-server
 import argparse
 import dataclasses
 import pprint
+import sys
 import textwrap
 from pathlib import Path
 from urllib.parse import urlparse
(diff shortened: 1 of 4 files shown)


rebkwok (commented):
Changes look fine to me, but it would probably be a good idea to add a test for this, since it's still not covered. I added the new codelists check flag that broke this, and I don;t think there's anything here that would prevent that happening again in future

bloodearnest:
> Changes look fine to me, but it would probably be a good idea to add a test for this, since it's still not covered. I added the new codelists check flag that broke this, and I don;t think there's anything here that would prevent that happening again in future

Indeed, I started with such a tests, but forgot to include it in the PR :facepalm:

rebkwok (approved):

PR #687, opened 2023-12-11
Remove automated pull of `databuilder` alias

evansd:
Pulling `ehrql:v1` should not attempt to pull `databuilder:v1` because there is no such image.

### DIFF
--- a/scripts/update-docker-image.sh
+++ b/scripts/update-docker-image.sh
@@ -5,24 +5,11 @@ if test "${1:-}" == ""; then
   echo "usage: $0 image[:tag]"
   exit 1
 fi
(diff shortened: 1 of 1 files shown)


rebkwok (approved):

rebkwok:
Are we still supporting databuilder in jobrunner? Can we, if the databuilder image no longer exists - or does it just alias to ehrql?  `config.py` and `commands.py` still have databuilder references

evansd:
There is no `databuilder:v1` image, but the `databuilder:v0` image still exists and will still work (I think OpenPROMPT are using it – or at least there were until quite recently).

PR #691, opened 2023-12-13
Collect, store and send aggregate metrics, take 2

bloodearnest:
This PR contains the 3 reverted commits https://github.com/opensafely-core/job-runner/pull/686, and then an additional commit to fix the use of a global variable to cache the connection string in multithreaded application.  This is known to be Bad Juju, and I regret the error.

The sync, run and stats thread *all* have a db connection to the metrics
db, so this cause sqlite to barf.

This was not caught in the test suite because we do not currently have
a functional test that runs the full multithreaded service and tries to
run a job with it.  This is in theory possible (it's what the
a backend-server tests for the test backend do), but its also hard to
set up and slow and requires lxc currently, so I have not done that
here.

Once this lands, I will test it does fix the multithreaded issue by
using the backend-server functional tests to fully exercise it.

The commit message for the fix is:

This removes the use of a process global, and adds support for readonly
connections to help reduce contention on the metrics db with
multithreaded access.

Adding readonly mode, and the initial PR feedback, led to a re-working
of the connection management. Rather than re-use
`lib.database.get_connection()`, we instead implement our own version.
This gives us control, and allows us to efficiently assure appropriate
set up as well as cleanly handle readonly connections.

One small chnage with this - we cannot use an in-memory db in testing
for the metrics db, as `mode=memory` is not compatible with setting
`mode=ro`. Each test still gets their own db, but its on disk.

### DIFF
--- a/jobrunner/record_stats.py
+++ b/jobrunner/record_stats.py
@@ -3,8 +3,10 @@ Super crude docker/system stats logger
 """
 import json
 import logging
+import sqlite3
 import subprocess
 import sys
+import threading
 import time
 from collections import defaultdict
 
@@ -28,24 +30,63 @@ CREATE TABLE IF NOT EXISTS jobs (
 )
 """
 
-_conn = None
+CONNECTION_CACHE = threading.local()
 
 
-def ensure_metrics_db():
-    global _conn
-    _conn = database.get_connection(config.METRICS_FILE)
-    _conn.execute("PRAGMA journal_mode = WAL")
-    _conn.execute(DDL)
+def get_connection(readonly=True):
+    db_file = config.METRICS_FILE
+
+    # developer check against using memory dbs, which cannot be used with this
+    # function, as we need to set mode ourselves
+    assert not db_file.startswith(
+        "file:"
+    ), "urls not supported for metrics db - must be path"
+
+    if readonly:
+        db = f"file:{db_file}?mode=ro"
+    else:
+        db = f"file:{db_file}?mode=rwc"
+
+    cache = CONNECTION_CACHE.__dict__
+    if db not in cache:
+        try:
+            conn = sqlite3.connect(db, uri=True)
+        except sqlite3.OperationalError as exc:
+            # if its readonly, we cannot create file, so fail gracefully.
+            # Caller should check for conn being None.
+            if readonly and "unable to open" in str(exc).lower():
+                return None
+            raise
+
+        # manual transactions
+        conn.isolation_level = None
+        # Support dict-like access to rows
+        conn.row_factory = sqlite3.Row
+
+        if not readonly:
+            conn.execute("PRAGMA journal_mode = WAL")
+            conn.execute(DDL)
+
+        cache[db] = conn
+
+    return cache[db]
 
 
 def read_job_metrics(job_id, **metrics):
-    if _conn is None:
-        ensure_metrics_db()
+    conn = get_connection(readonly=True)
 
-    raw_metrics = _conn.execute(
-        "SELECT metrics FROM jobs WHERE id = ?",
-        (job_id,),
-    ).fetchone()
+    raw_metrics = None
+
+    if conn is not None:
+        try:
+            raw_metrics = conn.execute(
+                "SELECT metrics FROM jobs WHERE id = ?",
+                (job_id,),
+            ).fetchone()
+        except sqlite3.OperationalError as exc:
+            if "no such table" not in str(exc).lower():
+                raise
+
     if raw_metrics is None:
         metrics = {}
     else:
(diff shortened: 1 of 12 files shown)


jobrunner/record_stats.py
```diff
@@ -28,24 +28,29 @@
...
+        raw_metrics = (
+            get_connection()
+            .execute(
+                "SELECT metrics FROM jobs WHERE id = ?",
+                (job_id,),
+            )
+            .fetchone()
+        )
+    except database.CannotOpenReadonlyDBFile:
+        raw_metrics = None
```
evansd, line 52: There's another possible failure mode here in that the database file could exist but not have the relevant tables created yet. Rather than raising `CannotOpenReadonlyDBFile` from `get_connection()` I'd be more inclined to catch general "SQLite barfed" errors here i.e. `read_job_metrics` is a best-effort function that just returns None if it can't read for any reason.

jobrunner/lib/database.py
```diff
@@ -168,13 +172,28 @@ def get_connection(filename=None):
...
+            log.info("applying initial DB connection setup")
+            try:
+                conn.execute("BEGIN")
+                for statement in initial:
+                    conn.execute(statement)
+            except Exception:
+                conn.rollback()
+                raise
+            else:
+                conn.commit()
```
evansd, line 196: I don't quite understand how this works. This is going to run the `initial` statements on every connection which are going to error once the tables exist. But I can't see what catches these errors.

jobrunner/record_stats.py
```diff
@@ -54,11 +59,8 @@ def read_job_metrics(job_id, **metrics):
 
 
 def write_job_metrics(job_id, metrics):
-    if _conn is None:
-        ensure_metrics_db()
-
     raw_metrics = json.dumps(metrics)
-    _conn.execute(
+    get_connection(readonly=False).execute(
```
evansd, line 63: Rather than passing `initial` to get connection and having it execute every time would another approach be to wrap this write in a `try/except` which catches the table not existing (probably have to do string matching on the error text) and then creates them and retries.

evansd (commented):
Sorry, I should have spotted the issue with that global because I had exactly the same problem myself a while back which is why `get_connection()` now uses a thread-local.

I'm not quite convinced the changes here are the right approach though. There's a test failure with:
```
sqlite3.OperationalError: cannot change into wal mode from within a transaction
```

But I think that indicates that `get_connection()` isn't the right place to do this set up.

jobrunner/lib/database.py
```diff
@@ -168,13 +172,28 @@ def get_connection(filename=None):
     # Looks icky but is documented `threading.local` usage
     cache = CONNECTION_CACHE.__dict__
     if filename not in cache:
-        conn = sqlite3.connect(filename)
+        try:
+            conn = sqlite3.connect(filename)
+        except sqlite3.OperationalError:
+            raise CannotOpenReadonlyDBFile(filename)
```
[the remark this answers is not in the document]
bloodearnest: I did consider checking the exception message.  But it might be better to let the exception propagate up and handle it at a higher level, as you allude to below.

jobrunner/lib/database.py
bloodearnest: It only runs if the connection is not already in the cache.

And the DDL is explicitly idempotent, by have IF NOT EXISTS.

jobrunner/record_stats.py
bloodearnest: That is an option. 

The reason I did it in get_connection is that is only place that knows if it is a new connection or not.  But I think that adding the readonly connections muddies those waters a bit.

bloodearnest:
Ok, I've completely reworking this, after we reverted the old.

This PR now includes the 3 reverted commits, and then 1 additional commit, which changes the connection management to a) avoid process global and b) support readonly mode

jobrunner/record_stats.py
```diff
@@ -17,12 +21,98 @@
...
+
+CONNECTION_CACHE = threading.local()
+
+
+def get_connection(readonly=True):
+    db_file = config.METRICS_FILE
+
+    # developer check against using memory dbs, which cannot be used with this
+    # function, as we need to set mode ourselves
+    assert not db_file.startswith(
```
evansd, line 41: The default `METRICS_FILE` value is a `pathlib.Path` instance and so won't have a `startswith` method. Cast to `str` maybe? Or mabye assert that it's a `pathlib.Path` rather than a `str`?

evansd (approved):
Nice, this makes sense. Small `str`/`Path` type confusion which will need fixing though.

jobrunner/record_stats.py
bloodearnest: Good catch, fixed.

PR #700, opened 2024-01-03
Add StatusCode.STALE_CODELISTS

bloodearnest:
Previously, the handling of a stale codelists meant an INTERNAL_ERROR
was raised, which needlessly alerted tech-support.

This change adds a new StatusCode to track this specific error. This
allows us to to avoid re-using INTERNAL_ERROR, and also make it easier
to measure stale codelist frequency.


madwort (approved):

jobrunner/create_or_update_jobs.py
```diff
@@ -337,10 +341,17 @@ def create_failed_job(job_request, exception):
...
-        code = StatusCode.INTERNAL_ERROR
-        status_message = f"{type(exception).__name__}: {exception}"
+
+        if isinstance(exception, StaleCodelistError):
+            code = StatusCode.STALE_CODELISTS
+            status_message = str(exception)
+        else:
+            code = StatusCode.INTERNAL_ERROR
+            status_message = f"{type(exception).__name__}: {exception}"
+
```
evansd, line 351: Might be slightly more readable to collapse this down to:
```python
if isinstance(exception, NothingToDoError):
    ...
elif isinstance(exception, StaleCodelistError):
    ...
else:
    ...
```

It would mean repeating `state = State.FAILED` on two branches, but I think the control flow is easier to read than nesting it.

evansd (approved):

jobrunner/create_or_update_jobs.py
bloodearnest: It's a bit more than `state`, but I've made it work.

madwort (approved):

PR #708, opened 2024-02-08
Reset cumulative metrics after job restart

### LINKED ISSUE #706: Fix job metrics for restarts
When a job is restarted (db maintenance or reboot), we should wipe any recorded stats, as they will break the maths.

Probably, we should this initialise all stats to zero just before we execute.

We don't have a good way to model re-executions of jobs (as its the same job), so the execution metrics will represent the latest execution *only*, not any previous ones.

This is what we want from user perspective, in many ways, but not from a system utilisation perspective.

Some more context https://bennettoxford.slack.com/archives/C069YDR4NCA/p1706606452556949

evansd:
This involves tracking the container ID associated with each job and resetting the metrics if we spot that the ID has changed.

Closes #706


bloodearnest (approved):
Nice pragmatic fix

Issue #709, opened 2024-02-27
TPP Job raised INTERNAL_ERROR


iaindillingham:
Honeycomb reports that a TPP Job raised an INTERNAL_ERROR.[^1] The affected jobs in Job Request [22288][] are:

* [measures_short_opioid][]
* [measures_strong_opioid][]

I've scanned the logs on L3 (both are 53,466 lines long 🤔), but can't determine why the jobs failed.

[^1]: https://bennettoxford.slack.com/archives/C0270Q313H7/p1708990931394309

[22288]: https://jobs.opensafely.org/opioid-prescribing-trends-and-changes-during-covid-19/waiting-list/22288/
[measures_short_opioid]: https://jobs.opensafely.org/opioid-prescribing-trends-and-changes-during-covid-19/waiting-list/22288/bigqnrugwvzdut6a/
[measures_strong_opioid]: https://jobs.opensafely.org/opioid-prescribing-trends-and-changes-during-covid-19/waiting-list/22288/fe5wkte2b67impkd/

evansd:
Interesting, looks like the state machine hit something unexpected:
```
Traceback (most recent call last):
  File "/home/opensafely/jobrunner/code/jobrunner/run.py", line 153, in handle_single_job
    synchronous_transition = trace_handle_job(job, api, mode, paused)
  File "/home/opensafely/jobrunner/code/jobrunner/run.py", line 190, in trace_handle_job
    synchronous_transition = handle_job(job, api, mode, paused)
  File "/home/opensafely/jobrunner/code/jobrunner/run.py", line 232, in handle_job
    api.terminate(job_definition)
  File "/home/opensafely/jobrunner/code/jobrunner/executors/logging.py", line 41, in wrapper
    status = method(job_definition)
  File "/home/opensafely/jobrunner/code/jobrunner/executors/local.py", line 208, in terminate
    assert current_status.state in [
AssertionError
```

From Honeycomb logs:
 * https://ui.honeycomb.io/bennett-institute-for-applied-data-science/environments/production/datasets/jobrunner/result/27nx99m1iwM/trace/JEnqHNDcTmP?fields[]=s_name&fields[]=s_serviceName&span=54097357b81a1faa
 * https://ui.honeycomb.io/bennett-institute-for-applied-data-science/environments/production/datasets/jobrunner/result/27nx99m1iwM/trace/D9YC7zZQmes?fields[]=s_name&fields[]=s_serviceName&span=27011caeedd272ee

[...]

PR #710, opened 2024-02-28
failing test for db maintenance mode terminate error

madwort:
* as seen in https://github.com/opensafely-core/job-runner/issues/709
* nb. this is a naive way of doing this & would be very slow - we could probably find a better way to test this, but it's important that the command executing in the container completes successfully (rather than being killed & returning an error)


madwort:
We're not prioritising fixing this, and this test is quite slow, so closing this ticket for now - it can be resurrected later.

PR #716, opened 2024-03-19
Tweak otel configuration

bloodearnest:
Use a root resource, with base information in, rather than repeating it
on all spans.

Also, use `job-runner` as default value for OTEL_SERVICE_NAME, as that
will help simplify things in the backend config


bloodearnest:
I give up

PR #718, opened 2024-03-20
Test the setup_default_tracing()

bloodearnest:
This resulted in some fix to the configuration too,
as consuming OTEL_SERVICE_NAME was not being done correctly

It does involve some messy reaching into otel internals though

### DIFF
--- a/tests/test_tracing.py
+++ b/tests/test_tracing.py
@@ -1,11 +1,66 @@
 import os
 import time
 
+import opentelemetry.exporter.otlp.proto.http.trace_exporter
 from opentelemetry import trace
+from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
+from opentelemetry.sdk.trace.export import ConsoleSpanExporter
 
 from jobrunner import config, models, tracing
 from tests.conftest import get_trace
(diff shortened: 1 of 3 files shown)


tests/test_tracing.py
```diff
@@ -1,13 +1,68 @@
...
+    processor = provider._active_span_processor._span_processors[0]
+    assert isinstance(processor.span_exporter, ConsoleSpanExporter)
+
+
+def test_setup_default_tracing_otlp_defaults(monkeypatch):
+    env = {"OTEL_EXPORTER_OTLP_HEADERS": "foo=bar"}
+    monkeypatch.setattr(os, "environ", env)
+    monkeypatch.setattr(
+        opentelemetry.exporter.otlp.proto.http.trace_exporter, "environ", env
+    )
```
rebkwok, line 35: just curious - why doesn't monkeypatching `os` also change it here? Is it just because `opentelemetry.exporter.otlp.proto.http.trace_exporter` does `from os import environ` rather than just `import os`, or is there something else going on?

rebkwok (approved):

tests/test_tracing.py
bloodearnest: That is exactly what is going on, sadly.

PR #724, opened 2024-04-24
l4 manifest

bloodearnest:
- **Write output metadata to manifest.json**
- **Add a cli command to backfill manifests for all workspaces**


jobrunner/cli/manifests.py
```diff
@@ -0,0 +1,94 @@
...
+Ops utility for backfilling manifest.json files from db
+"""
+import argparse
+
+from jobrunner.executors import local
+from jobrunner.lib import database
+from jobrunner.models import Job
+
+
+def main(workspaces=None):
```
evansd, line 11: nit: unused arg

jobrunner/cli/manifests.py
```diff
@@ -0,0 +1,94 @@
...
+    ]
+
+    n_workspaces = len(workspaces)
+    for i, workspace in enumerate(workspaces):
+        print(f"workspace {i+1}/{n_workspaces}: {workspace}")
+
+        level4_dir = local.get_medium_privacy_workspace(workspace)
+
+        sentinel = level4_dir / ".manifest-backfill"
+        if sentinel.exists():
```
evansd, line 24: Why do you need a sentinal file? Isn't the `manifest.json` itself the sentinal?

jobrunner/cli/manifests.py
```diff
@@ -0,0 +1,94 @@
...
+            print(" - already done, skipping")
+            continue
+
+        write_manifest(workspace)
```
evansd, line 28: There's a potential race-condition here if the main job-runner loop is itself in the process of updating the manifest. Is it feasible to stop job-runner while the backfill runs? Do you have a guess for how long it will take?

evansd (approved):
Looks good.

One more general question I have is under what circumstances old outputs get deleted? It used to be that if you ran `action_x` and it produced a different set of outputs to last time then any previous outputs would get deleted (and removed from the manifest). But I think that general logic has changed at some point? In any case, as long as the manifest reflects what's actually on disk then I think that's OK.

jobrunner/cli/manifests.py
bloodearnest: Because manifest.json does still exist, it much neutered form, so that osrelease knows what workspace this directory is for.

jobrunner/cli/manifests.py
bloodearnest: That was my plan, given how busy we are. I don't expect it take more than a couple of hours at most.

jobrunner/cli/manifests.py
evansd: Ahha! OK, fine

bloodearnest:
> Looks good.
> 
> One more general question I have is under what circumstances old outputs get deleted? It used to be that if you ran `action_x` and it produced a different set of outputs to last time then any previous outputs would get deleted (and removed from the manifest). But I think that general logic has changed at some point? In any case, as long as the manifest reflects what's actually on disk then I think that's OK.

I think that is still done? https://github.com/opensafely-core/job-runner/blob/main/jobrunner/run.py#L384

It's just done in the scheduler, not the executor (which has no knowledge of any outputs except the current job's outputs)

evansd:
> It's just done in the scheduler, not the executor (which has no knowledge of any outputs except the current job's outputs)

Ah yeah, I see. Doesn't that mean though that these don't get removed from the manifest and so that will end up with stale outputs in it which don't actually exist on disk?

bloodearnest:
> > It's just done in the scheduler, not the executor (which has no knowledge of any outputs except the current job's outputs)
> 
> Ah yeah, I see. Doesn't that mean though that these don't get removed from the manifest and so that will end up with stale outputs in it which don't actually exist on disk?

Oh rats, yes it will.

I think we'll need to make the executor's `API.delete_files()` update the manifest too?

PR #727, opened 2024-04-30
Prevent large CSV files from being copied to level 4

### LINKED ISSUE #726: Prevent large CSV files from being copied to level 4
We've been asked to ensure that CSV files with more than 5000 rows cannot be copied into level 4.  See discussion in [this thread](https://bennettoxford.slack.com/archives/C069YDR4NCA/p1714056557505449).

To do this, we should update `check_l4_file`:

 https://github.com/opensafely-core/job-runner/blob/ba22eb52538d18000516cd373e40196a0345f189/jobrunner/executors/local.py#L626-L668

If a user exceeds the limit, the error message should tell the user to contact tech-support.  We should then add a record to the tech-support routing table that in the first instance says "paste the issue in #team-ops and connect the researcher with Alex".

rebkwok:
Fixes #726 

Accompanying [docs PR](https://github.com/opensafely/documentation/pull/1501)

### DIFF
--- a/jobrunner/executors/local.py
+++ b/jobrunner/executors/local.py
@@ -511,15 +511,16 @@ def persist_outputs(job_definition, outputs, job_metadata):
         if level == "moderately_sensitive"
     ]
 
+    csv_metadata = {}
     # check any L4 files are vaild
     for filename in l4_files:
-        ok, job_msg, file_msg = check_l4_file(
+        ok, job_msg, file_msg, csv_counts = check_l4_file(
             job_definition, filename, sizes[filename], workspace_dir
         )
         if not ok:
             excluded_job_msgs[filename] = job_msg
             excluded_file_msgs[filename] = file_msg
-
+        csv_metadata[filename] = csv_counts
     medium_privacy_dir = get_medium_privacy_workspace(job_definition.workspace)
 
     # local run currently does not have a level 4 directory, so exit early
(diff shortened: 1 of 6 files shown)


jobrunner/executors/local.py
```diff
@@ -622,6 +622,17 @@ def get_output_metadata(
...
 
+MAX_CSV_ROWS_MSG = """
+The file:
+
+{filename}
+
+contained {row_count} rows, which is above the limit for moderately_sensitive files of
+{limit} rows.
+
+As such, it has *not* been copied to Level 4 storage.
```
inglesp, line 644: Please could you add a message about contacting tech-support?  (See [Slack](https://bennettoxford.slack.com/archives/C069YDR4NCA/p1714560739506289?thread_ts=1714552008.581379&cid=C069YDR4NCA).)

jobrunner/job_executor.py
```diff
@@ -36,6 +36,7 @@ class JobDefinition:
     cpu_count: str = None  # number of CPUs to be allocated
     memory_limit: str = None  # memory limit to apply
     level4_max_filesize: int = 16 * 1024 * 1024
+    level4_max_csv_rows: int = 5000
```
inglesp, line 39: Do we need a default here?  I'm a little bit wary of having two places that look like they define what the limit is.  If somebody were to change this, then there'd be no effect since it gets set in `job_to_job_definition`.

inglesp (approved):

jobrunner/executors/local.py
rebkwok: I had a message about contacting tech-support, and I took it out because of the conversation on the issue (about using the same process as files rejected for other reasons).

jobrunner/job_executor.py
rebkwok: No, I think we can remove it. Your comment would also apply to the `level4_max_filesize` on the line above.

bloodearnest: The reason there is a default for filesize was mainly testing convenience, so you didn't have to specify it every time you instantiate a job_definition directly (which many executor tests do, as they can't use `job_to_job_defintion` in most cases). So, not having a default will probably require updating a bunch of test cases/factories

An alternative might be to default to `None`, and update the consuming code to handle that case, perhaps?

rebkwok: It only involved updating one fixture and one function (in `test_local_executor.py` and `test_logging_executor.py` respectively), so i think this is fine.

PR #735, opened 2024-06-14
Add repo back in to manifest.json

rebkwok:
Include repo at the top level of manifest.json (used by Airlock)


rebkwok:
Closing as https://github.com/opensafely-core/airlock/pull/391 now always gets the repo from outputs, so no need to include it at the manifest root (for now at least)

PR #737, opened 2024-08-19
Test under python 3.10

bloodearnest:
Whilst we are still not running job-server inside docker, we need
job-runner to work under the system python. Currently, TPP is on
20.04/pyh38, andw e want to upgrade to 22.04/py3.10

job-runner does work on 3.10, as its bundled in opensafely-cli, and
that's been running on 3.10 fine for anyone on 22.04 for ages.

But this formalises it in CI.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,7 +26,7 @@ jobs:
   test-job:
     strategy:
       matrix:
-        os: [ubuntu-20.04, windows-2019, macos-12]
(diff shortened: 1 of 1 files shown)


rebkwok (approved):
Should it also test on ubuntu-22.04 instead of 20.04? Or is the plan to change this after tpp is upgraded?

bloodearnest:
> Should it also test on ubuntu-22.04 instead of 20.04? Or is the plan to change this after tpp is upgraded?

That's a good idea, will add.

PR #740, opened 2024-09-25
Full job id matching

### LINKED ISSUE #738: `kill_job` cli command should not require confirmation when full job ID is specified
The `kill_job` cli command was originally designed to be used with partial fragments of job IDs, typed by hand into the console. It therefore requires confirmation that the matching job is the intended one:
https://github.com/opensafely-core/job-runner/blob/a4e9cb4a1ff57f9751b970c4f6530d95755ff6bd/jobrunner/cli/kill_job.py#L57-L77

However we've recently had cause to use this with a large number of jobs where the full IDs were retrieved from the database and then fed in using `xargs` and we had to manually disable the confirmation check in order for this to be usable.

We should instead update the logic so that where the supplied ID is a full match for the job ID no confirmation is required.

Relevant Slack conversation:
https://bennettoxford.slack.com/archives/C02GL3A9THD/p1726823128853219

madwort:
* fix #738

### DIFF
--- a/jobrunner/cli/kill_job.py
+++ b/jobrunner/cli/kill_job.py
@@ -55,26 +55,25 @@ def get_jobs(partial_job_ids):
     jobs = []
     need_confirmation = False
     for partial_job_id in partial_job_ids:
-        # look for full matches
-        full_matches = database.find_where(Job, id=partial_job_id)
-        if len(full_matches) == 1:
-            jobs.append(full_matches[0])
+        # look for partial matches
+        partial_matches = database.find_where(Job, id__like=f"%{partial_job_id}%")
+        if len(partial_matches) == 0:
+            raise RuntimeError(f"No jobs found matching '{partial_job_id}'")
+        elif len(partial_matches) > 1:
+            print(f"Multiple jobs found matching '{partial_job_id}':")
+            for i, job in enumerate(partial_matches, start=1):
+                print(f"  {i}: {job.slug}")
+            print()
+            index = int(input("Enter number: "))
+            assert 0 < index <= len(partial_matches)
+            jobs.append(partial_matches[index - 1])
         else:
-            # look for partial matches
-            partial_matches = database.find_where(Job, id__like=f"%{partial_job_id}%")
-            if len(partial_matches) == 0:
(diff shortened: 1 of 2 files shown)


jobrunner/cli/kill_job.py
```diff
@@ -55,20 +55,26 @@ def get_jobs(partial_job_ids):
...
+                print(f"Multiple jobs found matching '{partial_job_id}':")
+                for i, job in enumerate(partial_matches, start=1):
+                    print(f"  {i}: {job.slug}")
+                print()
+                index = int(input("Enter number: "))
+                assert 0 < index <= len(partial_matches)
+                jobs.append(partial_matches[index - 1])
+            else:
+                need_confirmation = True
+                jobs.append(partial_matches[0])
```
evansd, line 77: There's a simpler implementation here which doesn't involve adding a separate database query. You can just add a conditional around the line which sets the `need_confirmation` flag to only do so if there isn't an exact match e.g.
```py
# We only need confirmation if the supplied job ID doesn't exactly match the found job
if job.id != partial_job_id:
    need_confirmation = True
jobs.append(partial_matches[0])
```

In practical terms there's not much difference (the performance difference is basically irrelevant here), but it reduces the depth of nesting and I think is therefore a bit easier to understand.

evansd (approved):
This looks correct to me, and really nice to see the test coverage improving here as well. Great job!

I've pointed out a possible simplification, but it's not a big deal and I'm happy for this to go as is.

When you come to merge this, instead of merging directly could you click the down arrow and use the "Squash and merge" option?
image

We generally use the "Create a merge commit" option only when the commits have been specifically crafted to make sense on their own. And when the commits just reflect the development workflow we use "Squash and merge" which means they end of as a single commit in the project history.

jobrunner/cli/kill_job.py
madwort: ah good idea, `job` isn't defined yet at that point, but this would work:
```
# We only need confirmation if the supplied job ID doesn't exactly match the found job
job = partial_matches[0]
if job.id != partial_job_id:
    need_confirmation = True
jobs.append(job)
```

madwort:
Nb. my gpg key had expired whilst I was working on this, hence `Unverified` commits.

PR #742, opened 2024-09-27
Fix docker image to actually work

bloodearnest:
The job-runner docker image is in a half working state. A couple of attempts have previously been made to figure out how to get it to be able to use the host's docker and file system to run jobs, but they were messy and also didn't fully work.

However, in docker 4.26.0, released Dec '23, they added support for [`--group-add`](https://docs.docker.com/desktop/release-notes/#4260), which made the permissions much simpler.  

Now we are on 22.04, we have a more up to date docker, and docker-compose v2, so this is simpler change. Using this new features, this change cleans up some of the old approaches.
We use a simple trick to make bind mounts work - the bind mounts must use the same path inside the container as on the host.  This avoids us needing to add new configuration options and logic, and seems a reasonable constraint.

- **Switch to using --mount instead of --volume**
- **Remove docker-over-ssh experiment support**
- **Move to explicit 22.04 base image**
- **Fix docker image permissions**
- **Fix docker configuration**
- **Add functional test to github workflows**

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -1,6 +1,8 @@
 # note: this docker-compose file is intended to define the *building* of
 # job-runner images, and running them locally in dev, not for running them in
 # production
+name: job-runner
+
 services:
   prod:
     # image name, both locally and public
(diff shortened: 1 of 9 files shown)


jobrunner/executors/local.py
```diff
@@ -175,7 +175,9 @@ def execute(self, job_definition):
                 label=LABEL,
                 labels=get_job_labels(job_definition),
                 extra_args=extra_args,
+                volume_type=volume_api.volume_type,
```
evansd, line 178: Rather than threading an extra arg through I think it would be worth considering making `volume` a more structured object i.e. dataclass or namedtuple with `source`, `target` and `type`.

docker/docker-compose.yaml
```diff
@@ -23,12 +23,19 @@ services:
         - GITREF
     # use dockers builitin PID daemon
     init: true
-    # paths relative to docker-compose.yaml
+    # run as the the current user to preserve file permissions
+    user: "${DEV_USERID:-1000}:${DEV_GROUPID:-1000}"
+    # run with additional group docker (the *hosts* gid for docker)
+    group_add: ["${DOCKER_HOST_GROUPID}"]
```
evansd, line 31: This is a nice solution :+1:

docker/docker-compose.yaml
```diff
@@ -29,24 +31,34 @@ services:
...
+      - ../workdir:/workdir
       # used to avoid ugly uid/gid lookup failures, but not strictly required
       - /etc/group:/etc/group:ro
       - /etc/passwd:/etc/passwd:ro
+      # these paths must be absolute on the host, and must be mount to the
+      # identical absolute paths inside the container. This allows for
+      # job-runner to use the same value for both its own file access, and for
+      # what to pass to the host docker as a bindmount source directory, which
+      # needs to be on the host. 
+      # We can potentially fix this by adding an explicit bind mount source configs, if we want to
```
evansd, line 46: Just requiring them to be identical seems like a pretty reasonable solution to me.

evansd (approved):
Always nice to see things getting simpler and more powerful at the same time!

Was a nice, well-explained series of commits as well. Could maybe do with editing the auto-generated PR title and description just to give more context for these changes.

jobrunner/executors/local.py
bloodearnest: That is cleaner, but also potentially a non-trivial amount of work, as I think doing it properly like will likely mean refactoring the VolumeAPI and resulting tests?

I have two days left to try get this ready for production before our next initiative starts, or else I don't know when we'll get change to work on it again. And I am still working on the backend-server plumbing to consume this image.

So, is worth doing here and now? Or something for later?

evansd: Oh yeah, totally ignorable for now! Sorry if I didn't make that clear. Just think it's always worth highlighting stuff when you happen to notice it.

bloodearnest: :heavy_check_mark: https://github.com/opensafely-core/job-runner/issues/748

PR #743, opened 2024-09-27
Set MAX_DB_WORKER limit in `local_run`

evansd:
We allow the user to set the number of concurrent jobs using the `--concurrency` argument and use this to update the worker limit, but we didn't update the database worker limit.

The distinction is meaningless for local jobs in any case because these jobs are just generating dummy data, but it meant that jobs ran with less parallelism than expected. It also lead to "Waiting on available database workers" messages appearing in the log output which caused confusion and anxiety.


evansd:
After this is merged and a new PyPI package published we'll need to update the vendored version in `opensafely-cli` and release a new version of that.

evansd:
I think failures are related to the issue Simon is battling at the moment, see:
https://bennettoxford.slack.com/archives/C069YDR4NCA/p1727436683675039

Once a fix for that is out I'll rebase this PR.

bloodearnest (approved):

PR #744, opened 2024-09-27
Upgrade to pyyaml 6.0.1

madwort:
* hoping this will fix the compiling from source issue that was preventing @Providence-o from installing pyyaml locally on Apple Silicon
* https://github.com/yaml/pyyaml/issues/736
* use macos-14 GHA runner


madwort:
looks like we have the same failure as https://github.com/opensafely-core/job-runner/pull/743#issuecomment-2379183134

bloodearnest (approved):

PR #747, opened 2024-09-30
Increase test coverage requirement to 84%

madwort:
* write a few simple tests to cover more of the codebase
* ignore some uninteresting parts of the codebase
* increase the target to 84

### DIFF
--- a/tests/lib/test_database.py
+++ b/tests/lib/test_database.py
@@ -17,6 +17,7 @@ from jobrunner.lib.database import (
     migrate_db,
     query_params_to_sql,
     select_values,
+    transaction,
     update,
     upsert,
 )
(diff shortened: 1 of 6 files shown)


tests/lib/test_database.py
```diff
@@ -56,6 +58,26 @@ def test_update_excluding_a_field(tmp_work_dir):
     assert j.commit == "commit-of-glory"
 
 
+def test_exists_where(tmp_work_dir):
+    insert(Job(id="foo123", state=State.PENDING))
+    insert(Job(id="foo124", state=State.RUNNING))
+    insert(Job(id="foo125", state=State.FAILED))
+    values = exists_where(Job, state__in=[State.PENDING, State.FAILED])
```
madwort, line 125: `values` is a weird choice here!

PR #754, opened 2024-10-08
Bump to pipeline 2024.10.08.124104

alarthast:
Make version 4 of pipeline available for choosing. 
It ceases supporting cohort extractor actions and does not use an `expectations` section in the `project.yaml` file.


rebkwok (approved):

alarthast:
Added `feat:` to commit message to tag new version

PR #757, opened 2024-10-11
Add a dedicated manual maintenance mode flag

bloodearnest:
Currently, manually enabling database maintenance mode is awkward, as by
default it will just get switch off again by the regular automatic
check. So you have to remember to stop that first, which is awkward,
then restart it, for more awkwardness

Instead, this change adds an explicit flag check for
`manual-db-maintenance` - if that is set, we skip the automated check
altogether.

This will be added to the jobrunner justfile in backend-server as
a documented command in a separate PR.

### DIFF
--- a/jobrunner/service.py
+++ b/jobrunner/service.py
@@ -93,6 +93,14 @@ def maintenance_mode():
     """Check if the db is currently in maintenance mode, and set flags as appropriate."""
     # This did not seem big enough to warrant splitting into a separate module.
     log.info("checking if db undergoing maintenance...")
+
+    # manually setting this flag bypasses the automaticaly check
+    manual_db_mode = get_flag_value("manual-db-maintenance")
+    if manual_db_mode:
+        log.info("manually set db mode: db-maintenance")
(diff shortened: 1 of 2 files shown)


evansd (approved):
:+1 I can definitely think of previous situations where this would have saved time and confusion.

As an aside (not a comment on this PR) we're use the string `"db-maintenance"` an awful lot in the code. It would seem worth replacing with a constant, given how easy I find it to misspell "maintenance".

bloodearnest:
> :+1 I can definitely think of previous situations where this would have saved time and confusion.
> 
> As an aside (not a comment on this PR) we're use the string `"db-maintenance"` an awful lot in the code. It would seem worth replacing with a constant, given how easy I find it to misspell "maintenance".

Yeah, I think we should probably change the name of the flag from `mode=db-maintenance` to `db-maintenance=on`. Originally, `mode` was intended to be multiple modes (e.g. `paused`), but that ended up needing to be a different thing.

I will add a constant for now

PR #758, opened 2024-10-11
fix: Bump pipeline version

alarthast:
- Part of https://github.com/opensafely-core/pipeline/issues/344
- More informative error message when parsing `project.yaml` files with empty actions
- Contains code to allow optional use of fast YAML parser, needs separate PR to install requirements

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ classifiers = [
 ]
 requires-python = ">=3.8"
 dependencies = [
-  "opensafely-pipeline @ git+https://github.com/opensafely-core/pipeline@v2024.10.08.124104",
+  "opensafely-pipeline @ git+https://github.com/opensafely-core/pipeline@v2024.10.11.170331",
   "ruyaml",
   "requests",
   "opentelemetry-exporter-otlp-proto-http",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -14,7 +14,7 @@ classifiers = [
 ]
 requires-python = ">=3.8"
 dependencies = [
-  "opensafely-pipeline @ git+https://github.com/opensafely-core/pipeline@v2024.10.08.124104",
+  "opensafely-pipeline[fastparser] @ git+https://github.com/opensafely-core/pipeline@v2024.10.11.170331",
```
evansd, line 17: Ah sorry, confusingly job-runner _can't_ depend on the `fastparser` extra because it ends up vendored inside opensafely-cli and we can't vendor compiled dependencies. This is part of the Gordian knot we trying to cut by switching to `uv`.

alarthast: Got it, I will switch it back :)

bloodearnest: We can't hard depend on it for the above reason, but we could add it to the job-runner-dependencies so we get the benefit in job-runner in prod.

evansd: Ah yeah, I hadn't thought of that.

evansd (approved):

PR #770, opened 2025-01-22
update pipeline

bloodearnest:
fix: update to latest version of pipeline.

Local tests were failing for me due to a pip-tools issue, so I took the opportunity to upgrade all the dev dependencies (1st commit). This created some black/flake churn (2nd commit), and then finally the actual update, which created its own churn (3rd commit)


madwort (dismissed):

requirements.dev.txt
```diff
@@ -1,151 +1,159 @@
...
+attrs==24.3.0
+    # via hypothesis
+black==24.8.0
     # via -r requirements.dev.in
+build==1.2.2.post1
+    # via pip-tools
 certifi==2020.11.8
     # via
-    #   -c requirements.prod.txt
+    #   -c /home/wavy/bennett/job-runner/requirements.prod.txt
```
madwort, line 15: this is very mildly irritating but not really worth the manual fixup

bloodearnest: I think this is a consequence of updating the pip version, and AFAICT, there's not much we can do about it, other than maybe post-processing the .txt files to remove the path?

inglesp (approved):

Issue #771, opened 2025-01-22
Write more error information into logs


bloodearnest:
If a job fails with unmatched outputs, we do not show this information in a the log file. 

Where as, if it had invalid L4 outputs, we do show this information.

It would be useful to users to see this in the log files, or else they have an exit_code 0 with no information in airlock, only in job-server.

https://github.com/opensafely-core/job-runner/blob/main/jobrunner/executors/local.py#L767

User thread: https://bennettoxford.slack.com/archives/C01D7H9LYKB/p1737563740332699?thread_ts=1737553869.538369&cid=C01D7H9LYKB

PR #776, opened 2025-02-10
Added hint and status message to output log for 'unmatched output' errors

Providence-o:
#771

### DIFF
--- a/jobrunner/executors/local.py
+++ b/jobrunner/executors/local.py
@@ -395,9 +395,26 @@ def finalize_job(job_definition):
 
     # First get the user-friendly message for known database exit codes, for jobs
     # that have db access
-    message = None
+    unmatched_hint = None
 
-    if exit_code == 137 and job_definition.cancelled:
+    if exit_code == 0 and unmatched_patterns == [] and outputs is not None:
+        message = "Completed successfully"
+
+    elif (
+        exit_code == 0
+        and unmatched_patterns is not None
+        and unmatched_outputs is not None
+    ):
+        message = "\n  No outputs found matching patterns:\n - {}".format(
+            "\n   - ".join(unmatched_patterns)
+        )
+        unmatched_hint = (
+            "\n  Did you mean to match one of these files instead?\n - {}".format(
+                "\n   - ".join(unmatched_outputs)
+            )
+        )
+
+    elif exit_code == 137 and job_definition.cancelled:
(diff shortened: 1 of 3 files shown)


Providence-o:
I've added `unmatched_hint = ""` to non-related conditional statements because without this it was causing the below error:

> **UnboundLocalError: local variable 'unmatched_hint' referenced before assignment**

jobrunner/executors/local.py
```diff
@@ -395,22 +395,41 @@ def finalize_job(job_definition):
 
     # First get the user-friendly message for known database exit codes, for jobs
     # that have db access
-    message = None
```
rebkwok, line 398: If you add a `unmatched_hint = None` here, you won't need to declare it in each branch of the if/elif/else below. In `write_log_file()` a job_metadata entry only gets written out if it exists, so default of None should be OK.

rebkwok (approved):
One suggestion about the default for `unmatched_hint`, otherwise lgtm.

jobrunner/executors/local.py
```diff
@@ -395,9 +395,26 @@ def finalize_job(job_definition):
...
+    unmatched_hint = None
 
-    if exit_code == 137 and job_definition.cancelled:
+    if exit_code == 0 and unmatched_patterns == [] and outputs is not None:
+        message = "Completed successfully"
+
+    elif (
+        exit_code == 0
+        and unmatched_patterns is not None
+        and unmatched_outputs is not None
```
bloodearnest, line 406: What about if there are unmatched patterns, but no unmatched outputs at all? I think we should still trigger this branch in that case?

Having unmatched outputs is not an error - only having unmatched patterns is an error, basically.

jobrunner/executors/local.py
```diff
@@ -395,9 +395,26 @@ def finalize_job(job_definition):
...
+    ):
+        message = "\n  No outputs found matching patterns:\n - {}".format(
+            "\n   - ".join(unmatched_patterns)
+        )
+        unmatched_hint = (
```
bloodearnest, line 411: I was wondering if it was worth having a separate `unmatched_hint` variable for this, or whether it might be simpler to just append this to the message if we have unmatched_outputs.

However, doing it this way means we don't send file names up to job-server, which is the currently policy, even if its probably going to change some point soon.

So on balance, I do think that a separate hint variable is worth it, nice work.

rebkwok (approved):

PR #777, opened 2025-02-11
Added new condition and test for unmatched pattern and output error

Providence-o:
Part of #776

### DIFF
--- a/jobrunner/executors/local.py
+++ b/jobrunner/executors/local.py
@@ -397,14 +397,10 @@ def finalize_job(job_definition):
     # that have db access
     unmatched_hint = None
 
-    if exit_code == 0 and unmatched_patterns == [] and outputs is not None:
+    if exit_code == 0 and outputs and not unmatched_patterns:
         message = "Completed successfully"
 
-    elif (
-        exit_code == 0
-        and unmatched_patterns is not None
-        and unmatched_outputs is not None
-    ):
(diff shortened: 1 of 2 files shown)


jobrunner/executors/local.py
```diff
@@ -397,14 +397,10 @@ def finalize_job(job_definition):
...
-    if exit_code == 0 and unmatched_patterns == [] and outputs is not None:
+    if exit_code == 0 and outputs and not unmatched_patterns:
         message = "Completed successfully"
 
-    elif (
-        exit_code == 0
-        and unmatched_patterns is not None
-        and unmatched_outputs is not None
-    ):
+    elif exit_code == 0 and unmatched_patterns and unmatched_outputs:
```
bloodearnest, line 403: So, I don't think this is quite right.

I think you should nest this, rather than handling it a separate if clauses
```
...
elif exit_code == 0 and unmatched_patterns:
    message = "No outputs found matching patterns:\n - {}".format(
       "\n   - ".join(unmatched_patterns)
    )
    if unmatched_outputs:
        hint = "Did you mean to match one of these files instead?\n - {}".format(
            "\n   - ".join(unmatched_patterns)
        )
elif ...
```

Using more complex separate if clauses is usually more brittle than nesting in this kind of case.

Does that make sense?

Providence-o: Yes it does, thank you!

bloodearnest (approved):

Issue #779, opened 2025-02-12
Do not error on unexpected state in terminate()


madwort:
[This function](https://github.com/opensafely-core/job-runner/blob/74b3cbf4f7d2b8faf766b2b6edb0c6ad98c59e16/jobrunner/executors/local.py#L204) is erroring occasionally in production on the assert

madwort:
> there’s only a [limited palette of states](https://github.com/opensafely-core/job-runner/blob/9ce393fe5178e3e1565f0046a05783800ea4e5f5/jobrunner/job_executor.py#L45-L61) which could have failed that assert, I’d be hugely surprised if it were in a prepare/preparing state, so that leaves executed, finalizing/finalized, error
the job [ran for 4h53m](https://ui.honeycomb.io/bennett-institute-for-applied-data-science/environments/production/datasets/jobrunner/result/rtK4qavVBHy/trace/zQgcZyE7NhW?fields[]=s_name&fields[]=s_serviceName&source=query&span=5f1a31e31bab1443) before this error happened
actions with the same ran actionname before & after have completed successfully [in less time](https://ui.honeycomb.io/bennett-institute-for-applied-data-science/environments/production/datasets/jobrunner/board-query/BvJk5LN1Lrc/result/rtK4qavVBHy?vs=hideCompare&tab=traces)
my guess is that we saw a race condition, in which the job was in in the state executing from the view of the loop, but had completed by the time that we [refresh the status in terminate()](https://github.com/opensafely-core/job-runner/blob/9ce393fe5178e3e1565f0046a05783800ea4e5f5/jobrunner/executors/local.py#L205) & was then in the state executed (& would have gone to finalizing in the next round of the run loop)
I can’t prove any of this definitively, but I think adding if executed: [log.info](http://log.info/)(); return before the assert would sort it

PR #780, opened 2025-02-12
Gracefully handle suspected race condition

madwort:
hopefully fixes https://github.com/opensafely-core/job-runner/issues/779


evansd (approved):
Great, yep. Makes sense to me.

PR #781, opened 2025-02-12
update CI to use 22.04

bloodearnest:


.github/workflows/tests.yml
```diff
@@ -27,7 +27,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-20.04, ubuntu-22.04, windows-2019, macos-14]
+        os: [ubuntu-22.04, ubuntu-22.04, windows-2019, macos-14]
```
madwort, line 30:
```suggestion
        os: [ubuntu-22.04, windows-2019, macos-14]
```

.github/workflows/tests.yml
```diff
@@ -52,7 +52,7 @@ jobs:
       - uses: extractions/setup-just@69d82fb0233557aec017ef13706851d0694e0f1d
 
       - name: Run actual tests on ${{ matrix.os }}
-        if: ${{ matrix.os == 'ubuntu-20.04' || matrix.os == 'ubuntu-22.04' }}
+        if: ${{ matrix.os == 'ubuntu-22.04' || matrix.os == 'ubuntu-22.04' }}
```
madwort, line 55:
```suggestion
        if: ${{ matrix.os == 'ubuntu-22.04' }}
```

.github/workflows/tests.yml
bloodearnest: ah, good catch

Will manually fix, rather than take suggestion, or else you can't review

madwort: or...

```suggestion
        os: [ubuntu-22.04, ubuntu-24.04, windows-2019, macos-14]
```

PR #783, opened 2025-02-18
Update .txt file if pipeline has changed

bloodearnest:
The current implementation just updates the .in file, so doesn't actually get tested :facepalm:

### DIFF
--- a/justfile
+++ b/justfile
@@ -97,7 +97,7 @@ upgrade env package="": virtualenv
     test -z "{{ package }}" || opts="--upgrade-package {{ package }}"
     FORCE=true "{{ just_executable() }}" requirements-{{ env }} $opts
 
-upgrade-pipeline:
+upgrade-pipeline: && requirements-prod
     ./scripts/upgrade-pipeline.sh pyproject.toml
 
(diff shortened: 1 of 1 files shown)


rebkwok (approved):
Out of interest, would the update-depenencies action have worked as expected if the two just commands in the workflow were in the other order? ie. if we changed it [here](https://github.com/opensafely-core/job-runner/blob/main/.github/workflows/update-dependencies.yml#L29) to:

```
just upgrade-pipeline
just upgrade dev 
``` 
(Makes sense for `just upgrade-pipeline` to do all the required things if it's run on its own though)

bloodearnest:
> Out of interest, would the update-depenencies action have worked as expected if the two just commands in the workflow were in the other order? ie. if we changed it [here](https://github.com/opensafely-core/job-runner/blob/main/.github/workflows/update-dependencies.yml#L29) to:
> 
> ```
> just upgrade-pipeline
> just upgrade dev 
> ```

I don't think so, as upgrade dev wouldn't update prod.txt?

We don't auto update any other prod deps except pipeline, due the awkward way they have to be vendored into opensafely-cli.

PR #786, opened 2025-03-11
fix: do not handle modes if job is to be cancelled.

### LINKED ISSUE #766: Cancelled database jobs aren't cancelled while in db-maintenance mode
In one sense this isn't a problem as the jobs are cancelled as soon as we come out of maintenance mode. But as well as a source of user confusion this potentially makes our stats/dashboard a bit misleading as we continue to consider the jobs active until the end of maintenance mode.

We could address this with an extra case in this branch (if a job isn't running but is cancelled them move it immediately to failed):
https://github.com/opensafely-core/job-runner/blob/4fba7431acf3eaab8b6cc88d0c018f4497a5b091/jobrunner/run.py#L228-L241

But I'm wary of adding even more complexity to the state manipulation code here. Maybe there's a more principled way of refactoring things here to get the behaviour we want?

Slack thread:
https://bennettoxford.slack.com/archives/C069YDR4NCA/p1733311815069519

bloodearnest:
The db-maintenance and pause modes have special handling at the top of
the handle_job method that eagerly handle those options.

This introduced a bug that job that are to be be cancelled do not get
cancelled if those modes affect them. This means that user's db jobs are
not cancelled when in db-maintenance mode, for example.

This quick fix simply protects the eager mode handling to apply only to
jobs that are not about to be cancelled, which will let the cancellation
logic proceed as normal.

This is probably not the ideal solution, but a) this is affecting users
right now with the extended db-maintenance mode and b) we are about to
completely refactor all the things, so we can live with a less than
ideal solution

Closes #766

### DIFF
--- a/jobrunner/run.py
+++ b/jobrunner/run.py
@@ -215,31 +215,33 @@ def handle_job(job, api, mode=None, paused=None):
     synchronous_transitions = getattr(api, "synchronous_transitions", [])
     is_synchronous = False
 
-    # handle special modes before considering executor state, as they ignore it
-    if paused:
-        if job.state == State.PENDING:
-            # do not start the job, keep it pending
+    # only consider these modes if we are not about to cancel the job
+    if not job_definition.cancelled:
+        # handle special modes before considering executor state, as they ignore it
+        if paused:
+            if job.state == State.PENDING:
+                # do not start the job, keep it pending
+                set_code(
+                    job,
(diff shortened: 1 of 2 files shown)


evansd (approved):
Excellent, thanks!

Don't forget to add `Closes #766` to the PR description so we get the ticket linkage.

PR #788, opened 2025-03-13
Post update-dependencies PRs to slack

rebkwok:


.github/workflows/update-dependencies.yml
```diff
@@ -13,19 +13,30 @@ jobs:
...
         token: ${{ steps.generate-token.outputs.token }}
         # note: we don't currently auto update external production dependencies,
         # as vendoring them into job-runner is messy
         update_command: |
           just upgrade dev
           just upgrade-pipeline
+
+    - name: Notify slack of PR
+      if: ${{ steps.update.outputs.pull-request-operation != 'none' }}
+      uses: slackapi/slack-github-action@v2.0.0
```
bloodearnest, line 36: I think our policy is to link to the full sha of an external action? Dependabot will still update it we add a comment:
So: `slackapi/slack-github-action@485a9d42d3a73031f12ec201c457e2162c45d02d  # v2.0.0`

I won't suggest, as then you'll need another review, as this is job-runner.
Have quickly done PR for airlock: https://github.com/opensafely-core/airlock/pull/901

bloodearnest (dismissed):
Small suggestion, will re-review straight away

.github/workflows/update-dependencies.yml
bloodearnest: And ehrql: https://github.com/opensafely-core/ehrql/pull/2417

rebkwok: slackapi is a [Marketplace verified creator](https://github.com/marketplace?query=slackapi&type=actions) so possibly it's also ok with a version?  Our org-level actions permissions include any marketplace verified creator actions as well as github-created ones

bloodearnest (approved):

Issue #804, opened 2025-03-31
Remove all code associated with `local_run` and set up proper functional test harness


evansd:
Now that the job-runner codebase is no longer intended to do double duty as the part of the local CLI tooling we can remove the `local_run` command and all the associated configuration and branching logic which went along with it. This simplification and clarification should make the upcoming refactoring work easier.

This will involve adding some new machinery to run end-to-end tests. Because the features of `local_run` are very similar to what you'd want from a test harness, parts of the test suite currently use `local_run` to exercise the code and so these will need updating.

PR #812, opened 2025-03-31
Remove DockerVolumeAPI

### LINKED ISSUE #811: Remove `DockerVolumeAPI`
As we no longer need to support running on Windows, we can remove all code relating to the `DockerVolumeAPI` class.  This will reduce the amount of code we have to deal with, and will make the test suite much faster.

This means that there will be a single volume API class (`BindMountVolumnAPI`), and there are number of places in the code that expect a list of volume API classes.  Unless it's going to be very easy, we should not change that code for now, but save it for later refactoring efforts.

inglesp:
Fixes #811.

Further easy refactoring` is possible (specifically: dropping the `BindMountVolumeAPI` namespace and moving all static methods to the module) but this is likely to cause merge conflicts with work on #804.


bloodearnest (approved):
This make sense to me, and is one dark corner we won't need to worry about anymore

PR #813, opened 2025-03-31
Remove `local_run` command and associated code

### LINKED ISSUE #804: Remove all code associated with `local_run` and set up proper functional test harness

evansd:
I've removed all explicit references to `local_run` and any code which is now obviously redundant. I'm almost certain there are further simplifications which could be made but I think this will do for now.

Running `pytest` locally this knocks about a third off the test suite run time (from 230s to 166s).

Closes #804


bloodearnest (approved):
noice

PR #815, opened 2025-04-01
Split test_run into agent and controller tests

bloodearnest:
This should help clarify responsibilities doing the actual split


bloodearnest:
OK, we've actually gain 18 lines, I assume from duplicated imports.

As this is job-runner, you'll need to re-approve

PR #821, opened 2025-04-02
Inital Task API using db table

bloodearnest:
Initial implementation of a Task API, using a new db table to communicate
between agent and controller.

This is currently only excercised by tests, but should act as a foundation for
splitting things up.

### DIFF
--- a/jobrunner/agent/task_api.py
+++ b/jobrunner/agent/task_api.py
@@ -1,0 +1,44 @@
+from jobrunner.lib import database  # cheating!
+from jobrunner.models import Task as ControllerTask  # cheating!
+from jobrunner.schema import AgentTask, TaskStage
+
+
+def get_active_tasks() -> list[AgentTask]:
+    """Get a list of active tasks from the controller"""
+    # cheating for now - should be HTTP API call
+    return [
+        AgentTask(
+            id=t.id,
+            type=t.type,
+            definition=t.definition,
+            created_at=t.created_at,
+        )
+        for t in database.find_where(ControllerTask, active=True)
+    ]
+
+
+def update_controller(
+    task: AgentTask,
+    stage: TaskStage,
+    timestamp: int,
+    results: dict = None,
+):
+    """Update the controller with the current state of the task.
+
+    stage: the current stage of the task from the agent's perspective
+    timestamp: ns timestamp of current stage was acheived (for accurate telemetry)
+    results: optional dictionary of completed results of this task, expected to be immutable
+    """
+    # cheating for now - should be HTTP API call, which ends up updating the
+    # controller db, but we just update the db directly
+    #
+    # this currently just updates the task table, and lets the main controller
+    # loop update the jobs table as needed, and handle a completed task. In the
+    # future, we may want the HTTP handler to do both, so that the main loop
+    # does not need to handle agent updates and completed jobs at all.  But all
+    # we have currently is the loop, so we'll do that logic there for step 1
+    db_task = database.find_one(ControllerTask, id=task.id)
+    db_task.agent_stage = stage
+    db_task.agent_stage_ns = timestamp
+    db_task.agent_results = results
+    database.update(db_task)
--- a/jobrunner/agent/task_api.py
+++ b/jobrunner/agent/task_api.py
@@ -22,6 +22,7 @@ def update_controller(
     stage: TaskStage,
     timestamp: int,
     results: dict = None,
+    complete: bool = False,
 ):
     """Update the controller with the current state of the task.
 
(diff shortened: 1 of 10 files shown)


jobrunner/schema.py
```diff
@@ -0,0 +1,43 @@
...
+
+from jobrunner.job_executor import ExecutorState
+
+
+class TaskType(Enum):
+    RUNJOB = "runjob"
+    # TODO: delete job
+
+
+class TaskStage(Enum):
```
madwort, line 12: should this be called ExecutorState for the time being?

jobrunner/schema.py
```diff
@@ -0,0 +1,43 @@
...
+    controller, and is effectively immutable.
+
+    The agent's view of a task's current state is computed on demand from
+    disk/docker state
+    """
+
+    id: str  # noqa: A003
+    type: TaskType  # noqa: A003
+    definition: dict
+    created_at: int = None
```
madwort, line 43: should we also store `State` & `StatusCode` here?

bloodearnest: I don't think so, as the State will never change from the agent's POV (it's always RUNNING), and StatusCode is the controller's responsibility, agent should never worry about it.

bloodearnest: The presences of agent_results is the primary indicated that a task has finished:

https://github.com/opensafely-core/job-runner/pull/821/files#diff-c71376d8e6de22a033c81b0636cc27f102c41485bc40d738fcaca92176d61a01R419

Additionally, I do not think is not going to be possible in step 1 to make that TaskStage opaque to the controller, as it will need to set the relevant StatusCode in the controller DB in order to maintain the current behaviour.

Once we are free to change things, there may be a way to make them opaque, perhaps.

jobrunner/agent/task_api.py
```diff
@@ -0,0 +1,44 @@
...
+    task: AgentTask,
+    stage: TaskStage,
+    timestamp: int,
+    results: dict = None,
+):
+    """Update the controller with the current state of the task.
+
+    stage: the current stage of the task from the agent's perspective
+    timestamp: ns timestamp of current stage was acheived (for accurate telemetry)
+    results: optional dictionary of completed results of this task, expected to be immutable
```
evansd, line 30: Ah, this is an interesting difference in how I was thinking about things. In my head, I didn't have a specially, distinguished final results dict. Instead, the agent just sends up JSON blobs from the task, at any point the task wants to emit them, the details of which are determined by the task itself. Then it's up the controller to decide that a task is done.

I do think we'll have a need for some kind of mutable data that's more than just the `stage` enum so that tasks can provide more complex information about what they're currently up to. Whether that's in addition to the `stage` or subsumes it, I'm less sure.

We should probably discuss all this directly rather than on the PR.

evansd (approved):
Solid start. I'd be happy to merge as is and then iterate on the details.

jobrunner/schema.py
bloodearnest: I don't think so, as we need to add a delete file task for step 1, i.e. soon

jobrunner/agent/task_api.py
bloodearnest: So, some stuff will be the same for all tasks e.g. a stage value,a dn the timestamp in happened.

Whereas the results json format will be different per task type.

I get that we may want to send more that stage up as general task metadata, but I don't think we should mix that with task-specific results, I'd prefer to keep them separate.

We can add a completed boolean flag easily enough.

jobrunner/agent/task_api.py
```diff
@@ -0,0 +1,44 @@
...
+    ]
+
+
+def update_controller(
+    task: AgentTask,
+    stage: TaskStage,
+    timestamp: int,
+    results: dict = None,
+):
+    """Update the controller with the current state of the task.
```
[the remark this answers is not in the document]
bloodearnest: Not the timestamp comes from things on disk - the timestamp of files, the docker container metadata, etc

jobrunner/schema.py
```diff
@@ -0,0 +1,43 @@
...
+
+    # RUNJOB stages
+    PREPARING = ExecutorState.PREPARING.value
+    PREPARED = ExecutorState.PREPARED.value
+    EXECUTING = ExecutorState.EXECUTING.value
+    EXECUTED = ExecutorState.EXECUTED.value
+    FINALIZING = ExecutorState.FINALIZING.value
+    # Final stages
+    FINALIZED = ExecutorState.FINALIZED.value
+    ERROR = ExecutorState.ERROR.value
```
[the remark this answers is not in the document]
bloodearnest: No, it's a parallel state enum for the agent parts of the job process. We already have this in ExectutorState, but this is meant to be broader that just RUNJOB stages

PR #825, opened 2025-04-03
Get coverage to 100%

### LINKED ISSUE #802: Get coverage to 100%
* Set coverage target to 100%
* add `pragma: no cover` to all missing lines until it's happy
* add coverage by writing tests if important/easy/other reason

rebkwok:
Fixes #802 
- Adds some extra tests and skips to get coverage to 100%
- Removes a few functions that are no longer used (previously only used in the removed local_run or the DockerVolumeAPI)
- Skips all remaining uncovered lines in run.py - I didn't look at adding any extra tests here because I think this is about to be split up. We'll want to revisit though (see #828)


madwort:
one of the commits is titled 'Remove some used functions' and should be 'unused' - can be fixed up when you merge

PR #829, opened 2025-04-04
Build & push docker images for dev branch

madwort:
* created a separate workflow, so @inglesp 's comment no longer applies


inglesp (commented):
It looks like we're now only building off the dev branch.  I'm keen that can still merge into main, and for that to be deployed.  Is that possible?  Sorry if I've been unclear here.

madwort:
oh, the ticket said:

> We should automatically deploy from the dev branch to the test backend

we can do something else if you prefer!

inglesp (approved):

PR #833, opened 2025-04-08
Task API tweaks

bloodearnest:
- no timestamps - agents gonna do its own telemetry, at least for now
- add backend field, as controller needs to support multiple
- get rid of TaskStage - the controller doesn't care, just use a string
- renamed completed_at to finished_at, as semantically different
   from completed flag


evansd (approved):
Great. Thanks for getting this in place so quickly.

PR #835, opened 2025-04-09
Initial implementation of controller loop

### LINKED ISSUE #826: Initial Controller split
The goal is to split the parts of run.py loop and handle_job function out into a new controller specific loop.

 - a loop that handles all PENDING job logic
   - handle all the reasons why a job is not ready to run or has failed before it even
     started
   - when a job is ready to run, insert RUNJOB task into task queue
 - a loop that handles all completed tasks
   - calculate final state and update the job db
   - mark the task as complete

Spike code: https://github.com/opensafely-core/job-runner/blob/api-spike/jobrunner/controller.py  

Notes:

 - suggest code lives in jobrunner/controller/
 - task api is at jobrunner/controller/task_api.py
 - tests already split out in tests/test_controller.py 
 - punt on making the delete_files api a task  - just preserve the sync code current 
   code for now.
 - tests should be able to fake the agent doing work by just calling 
   `agent.task_api.update_controller` in the test to set the desired task state for 
   tests
 - hopefully drop all use of the StubExecutorAPI in tests?
 - may be don't worry about emitting telemetry initially, but go back after and handle 
   non completed task updates in the task loop?

evansd:
This is currently a draft: there are no tests and no guarantees that the code even runs.

To hopefully make the changes easier to follow this PR
 * starts with an exact copy of the existing run file;
 * removes a load of unused stuff;
 * redefines the `handle_job()` function, making liberal use of stub-functions with no implementation;
 * attempts to define the stub functions;
 * makes (some of) the changes necessary elsewhere to make this all work

There are TODOs sprinkled liberally throughout, particular regarding support for multiple backends.

Will eventually close #826


bloodearnest (approved):

bloodearnest (commented):
Left a comment in slack that I will repeat here:

I wonder if pragma no cover is worth it here, in this specific context.
We're merging to dev, so there's no passing tests (i.e. 100% coverage) required.
Leaving them out could help us rapidly cover missing code.
Putting them in means we may forget to remove them, so end up with missing coverage.
So I say we have a period over the next week or so where we are living with failing coverage, and work towards actual 100%

PR #836, opened 2025-04-09
Skeleton minimal mobbed agent

bloodearnest:
This is a somewhat meandering set of commits that explore and end up with a minimally working agent with a single test

This was done via mobbing over multiple days with multiple drivers, so some commits are a bit messy as state was passed around.

### DIFF
--- a/jobrunner/agent/main.py
+++ b/jobrunner/agent/main.py
@@ -1,0 +1,758 @@
+"""
+Script which polls the database for active (i.e. non-terminated) jobs, takes
+the appropriate action for each job depending on its current state, and then
+updates its state as appropriate.
+"""
+
+import collections
+import datetime
+import logging
+import os
+import sys
+import time
+
+from opentelemetry import trace
+
+from jobrunner import config, tracing
+from jobrunner.executors import get_executor_api
+from jobrunner.job_executor import (
+    ExecutorAPI,
+    ExecutorRetry,
+    ExecutorState,
+    JobDefinition,
+    Privacy,
+    Study,
+)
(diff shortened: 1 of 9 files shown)


jobrunner/agent/main.py
```diff
@@ -0,0 +1,241 @@
...
+
+def handle_tasks(api: ExecutorAPI | None):
+    active_tasks = task_api.get_active_tasks()
+
+    handled_tasks = []
+
+    for task in active_tasks:
+        # `set_log_context` ensures that all log messages triggered anywhere
+        # further down the stack will have `job` set on them
+        with set_log_context(task=task.id):
```
rebkwok, line 56: Do we mean task id or job id now? When we wrote this, we were assuming they were the same

jobrunner/agent/main.py
```diff
@@ -0,0 +1,241 @@
...
+
+# we do not control the transition from these states, the executor does
+def handle_single_task(task, api):
+    """The top level handler for a task.
+
+    Mainly exists to wrap the task handling in an exception handler.
+    """
+    # we re-read the flags before considering each task, so make sure they apply
+    # as soon as possible when set.
+    mode = get_flag_value("mode")
```
rebkwok, line 72: We don't care about mode anymore, do we?

jobrunner/agent/main.py
bloodearnest: Good point.

bloodearnest: Hmm, actually, at this point, we don't know if the task is job related or not

So I think task.id is correct.

In handle_run_job_task, we can *add* the job id to the log context

PR #839, opened 2025-04-10
add telemetry for the agent loop and runjob tasks

rebkwok:
Add tracing for the agent loop.

Sets common attributes from the task (which later may not be job-related), and then attributes from the job. I didn't add all job attributes, and this is a job definition rather than a job, so it doesn't have quite the same things as the main jobrunner.tracing module has for a job.  We may need to revisit what's actually useful to include.
  
On each update_controller call, we also add the final job status and results/error attributes if there are any.

### DIFF
--- a/tests/agent/test_agent.py
+++ b/tests/agent/test_agent.py
@@ -2,6 +2,7 @@ from jobrunner.agent import main
 from jobrunner.controller import task_api as controller_task_api
 from jobrunner.job_executor import ExecutorState
 from tests.agent.stubs import StubExecutorAPI
+from tests.conftest import get_trace
 
 
 def test_handle_job_full_execution(db, freezer):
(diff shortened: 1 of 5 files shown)


tests/agent/test_agent.py
```diff
@@ -40,9 +41,70 @@ def finalize(job_id):
...
+            # results included on the final span
+            "unmatched_patterns",
+            "image_id",
+            "unmatched_outputs",
+            "message",
+            "exit_code",
+        }
+    )
+
+    assert spans[0].attributes["initial_job_status"] == "UNKNOWN"
```
bloodearnest, line 102: What do you think about this test just testing the correct statuses, and not all the attributes, and move the testing of attributes into a test_tracing.py module?

bloodearnest (approved):
LGTM, just one suggestion

tests/agent/test_agent.py
evansd: The various tracing assertions in the controller tests I was working on also felt a bit weird, but it didn't seem the right time to be worrying about them and so I left them as they were all passing.

PR #846, opened 2025-04-11
Switch service.py to using split agent/controller rather than run.py

bloodearnest:
We may want to rethink our naming a bit in future, too many mains!

Note that running both agent and controller in threads is a temporary
thing - they will be moving to being independant processes soon.

This is just a quick path to being able to run them both simulataneously
using the existing machinery.

Tested manually with add job that it works

### DIFF
--- a/jobrunner/service.py
+++ b/jobrunner/service.py
@@ -6,7 +6,9 @@ import logging
 import threading
 import time
 
-from jobrunner import config, record_stats, run, sync, tracing
+from jobrunner import config, record_stats, sync, tracing
+from jobrunner.agent.main import main as agent_main
+from jobrunner.controller.main import main as controller_main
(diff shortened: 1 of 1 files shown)


jobrunner/service.py
```diff
@@ -28,7 +30,7 @@ def start_thread(target, name):
 def main():
     """Run the main run loop after starting the sync loop in a thread."""
     # extra space to align with other thread's "sync" label.
```
rebkwok, line 32: I know it's only temporary, but can we update the docstring and comment to match the change?

rebkwok (approved):

jobrunner/service.py
bloodearnest: ah, yeah, good catch

PR #852, opened 2025-04-11
Implement agent handle_cancel_job_task

rebkwok:
Implements cancelling jobs in the agent. Note that the controller only includes job id in the CANCELJOB task definition, which means that if we need to write job logs, there will be some information that we don't have. We could always have the controller send this too, but I haven't done that yet and have left a TODO.

Testing locally is a bit tricky, this is what I did:

1) Edit `main()` in `agent/main.py` so that the agent only runs one loop at a time:
```
def main(exit_callback=lambda _: True):
    ...
```
2) Add a job with `just add-job`
```
just add-job https://github.com/opensafely/test-age-distribution generate_dataset -f
```
Make a note of the job id

3)  `just run-controller`
You should see the job get picked up and log that it's executing. Leave it running for the rest of the steps.

4) In another terminal, `just run-agent`
This will run one loop and move the runjob task into PREPARED
Repeat to move it into EXECUTING

5) In a shell, cancel the job:
```
In [1]: from jobrunner.models import Job
In [2]: from jobrunner.lib.database import update_where
In [3]: update_where(Job, {"cancelled": True}, id=<job id>)
```
You should see the controller register that the job has been cancelled by user.

6) `just run-agent`
You should see the agent pick up the CANCELJOB task

### DIFF
--- a/jobrunner/agent/main.py
+++ b/jobrunner/agent/main.py
@@ -110,6 +110,9 @@ def handle_cancel_job_task(task, api):
     # a dummy JobDefinition to use with the executor API. job ID is all we
     # need to find out the current status and do the actions required to
     # cancel and clean up
+    # TODO: finalize() writes job logs, and will be missing some expected information
+    # if only job_id is passed (from job definition fields, it expects to write
+    # job id, job_request_id, created_at, database_name and commit)
     job = JobDefinition.from_job_id(task.definition["job_id"], cancelled=True)
 
     job_status = api.get_status(job)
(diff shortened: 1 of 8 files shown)


bloodearnest (approved):
Looks good

I think the controller should send the full job_definition in canceljob.

I would be keen to get rid of JobDefinition.cancelled, have left a suggestion of how we might do so, but happy to land as is.

rebkwok:
I have made the controller send the full job definition instead of just job_id, and I've updated the `case _`, but I'm going to leave getting rid of the job definition `cancelled` flag for another PR (have made #860)

PR #862, opened 2025-04-15
Remove cancelled flag on JobDefinition

### LINKED ISSUE #860: Get rid of the JobDefinition.cancelled flag
See [comments here](https://github.com/opensafely-core/job-runner/pull/852#discussion_r2041993800)

rebkwok:
Fixes #860 

- removes the cancelled flag from JobDefinition
- adds a cancelled arg to the executor api `get_status()` and `finalize()`, True when these methods are called from a canceljob task in the agent
-  persists the cancelled status in the job metadata so that subsequent calls to get_status can know if a job was already cancelled
- this means that we now call finalize() on all cancelled jobs, even if they haven't started, so that we can persist the metadata, including the created_at/cancelled status.

### DIFF
--- a/jobrunner/agent/main.py
+++ b/jobrunner/agent/main.py
@@ -132,10 +132,10 @@ def handle_cancel_job_task(task, api):
                 new_status = api.terminate(job)
                 update_controller(task, new_status)
                 # call finalize to write the job logs
-                final_status = api.finalize(job)
+                final_status = api.finalize(job, cancelled=True)
             case ExecutorState.EXECUTED | ExecutorState.ERROR:
                 # job has finished or errored, call finalize to write the job logs
-                final_status = api.finalize(job)
+                final_status = api.finalize(job, cancelled=True)
             case _:  # pragma: no cover
                 raise InvalidTransition(
                     f"unexpected state of job {job.id}: {job_status.state}"
--- a/jobrunner/agent/main.py
+++ b/jobrunner/agent/main.py
@@ -117,27 +117,35 @@ def handle_cancel_job_task(task, api):
     span = trace.get_current_span()
     span.set_attributes({"id": job.id, "initial_job_status": job_status.state.name})
 
+    # tell the controller what stage we're at now
+    update_controller(task, job_status)
+
     with set_log_context(job_definition=job):
         match job_status.state:
-            case ExecutorState.UNKNOWN | ExecutorState.FINALIZED:
-                # The job hasn't started or has already finished
+            case ExecutorState.FINALIZED:
+                # The job has already finished and been finalized, nothing to do here
                 final_status = job_status
-            case ExecutorState.PREPARED:
-                # Nb. no need to actually run finalize() in this case.
-                final_status = JobStatus(ExecutorState.FINALIZED)
+            case (
+                ExecutorState.UNKNOWN
+                | ExecutorState.PREPARED
+                | ExecutorState.EXECUTED
+                | ExecutorState.ERROR
+            ):
+                # States wehere we need to run finalize()
+                # If the job hasn't started; we run finalize() to record metadata, including
+                # its cancelled state. If it's finished or errored, finalize() will also write the job logs
+                final_status = api.finalize(job, cancelled=True)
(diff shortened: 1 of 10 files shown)


jobrunner/executors/local.py
```diff
@@ -553,6 +567,12 @@ def write_job_logs(
             )
 
 
+def write_job_metadata(job_definition, job_metadata):
```
bloodearnest, line 567: Worth maybe yoinking this up to being near to read_job_metadata, as they are inverses of each other? Keep the path logic shared between them closer together.

jobrunner/executors/local.py
```diff
@@ -279,12 +283,15 @@ def get_status(self, job_definition, timeout=15):
         metrics = record_stats.read_job_metrics(job_definition.id)
 
         if container is None:  # container doesn't exist
-            if job_definition.cancelled:
+            # cancelled=True indicates that we are in the process of cancelling this
+            # job. If we're not, the job may have been previously cancelled; look up
+            # its cancelled status in job metadata, if it exists
+            if cancelled or (self.get_metadata(job_definition) or {}).get("cancelled"):
```
bloodearnest, line 289: Do you think that we could avoid having to pass cancelled to get_status if

a) in finalize_job, we add a cancelled key to job_metadata, and set to to value of the cancelled variable we passed to finalize_job?
b) in get_status, we read the metadata earlier, and use metadata["cancelled"] to detect cancellation?

This also permanently persists the metadata that this job was cancelled

wydt?

bloodearnest (approved):
Looks good. The clean up in finalize_job about how handling missing metadata is similar to changes @Providence-o and I were making for persisting errors, will be useful thanks.

Let a suggesting about persisting the fact a job is cancelled, which I think means we can tell that automatically in get_status, and possibly don't need to pass cancelled flag there?

jobrunner/executors/local.py
rebkwok: I don't know that we can; we call get_status() as one of the first things in `handle_canceljob_task`; the status we get back there depends on whether we're cancelled or not. When we're at the start of `handle_canceljob_task`, we haven't written any metadata yet, because we haven't actually cancelled it yet, so we haven't called `finalize`

rebkwok: And it's already doing (a)

bloodearnest: Hmm, yeah, I see the problem.

So, I think it's possible that we do not need get_status correctly differentiate between PREPARED and UNKNOWN in the face of being cancelled anymore, as we're calling it from an entirely separate task handler. The only reason I think it currently exists is to set a different message, which is no longer the agents responsibility.

So, I think we could possibly remove all cancellation special casing in get_status. That will still lead to UNKNOWN or PREPARED being returned, which are handled appropriately already? There is no message to set.

In fact, it might be worth getting rid of JobStatus.message altogether?

jobrunner/agent/main.py
```diff
@@ -106,41 +106,46 @@ def trace_handle_task(task, api):
...
                 new_status = api.terminate(job)
                 update_controller(task, new_status)
                 # call finalize to write the job logs
-                final_status = api.finalize(job)
-            case ExecutorState.EXECUTED | ExecutorState.ERROR:
-                # job has finished or errored, call finalize to write the job logs
-                final_status = api.finalize(job)
+                final_status = api.finalize(job, cancelled=True)
             case _:  # pragma: no cover
                 raise InvalidTransition(
```
bloodearnest, line 144: Another potential use of `assert False` here?

jobrunner/agent/main.py
```diff
@@ -106,41 +106,46 @@ def trace_handle_task(task, api):
...
                     f"unexpected state of job {job.id}: {job_status.state}"
                 )
 
+        # Clean up based on the starting job state
         if job_status.state in [
```
bloodearnest, line 149: I am not sure what the states check is doing here. api.cleanup is idempotent, and we never want to leave things left around in the cancel case, so I wonder if its worth just getting rid of the if?

bloodearnest:
doh too late

Issue #861, opened 2025-04-15
Get executor to write errors to disk


Providence-o:
The agent needs to to persist errors to disk and send to the controller as part of the results. This will also involve updating `get_status` to read `error.json` files.

PR #866, opened 2025-04-16
Get errors to persist to disk and update controller with results 

Providence-o:
Fixes: #861

The original plan has been to persist the error information in a separate error.json file. However, this proved to be awkward, as it required plumbing a bunch of extra error related methods that mirrored the metadata methods, e.g. read/write_job_error, get_error, etc.

Instead, we chose to embed the error information in the already persisted metadata.json that is written by ``finalize_job`

This involved making finalize_job work on a job in any state, as an error can happen at any times, and storing as much info as it has available.

Along the way, we refactored away the get_results/get_metadata executor API calls, and instead just return the metadata as part of every get_status() call. This simplified a bunch of things, as we already were loading and checking for metadata.json in get_status anyway.

There's a lot of commits, but they are all fairly small and consistent - we fixed a few smaller issues along the way.

### DIFF
--- a/jobrunner/executors/local.py
+++ b/jobrunner/executors/local.py
@@ -70,7 +70,7 @@ def read_job_metadata(job_definition):
     if path:
         return json.loads(path.read_text())
 
-    return None
+    return {}
 
 
 def write_job_metadata(job_definition, job_metadata):
--- a/jobrunner/executors/local.py
+++ b/jobrunner/executors/local.py
@@ -234,7 +234,7 @@ class LocalDockerAPI(ExecutorAPI):
             return JobStatus(ExecutorState.ERROR, f"failed to finalize job: {exc}")
 
         # this api is synchronous, so we are now FINALIZED
-        return JobStatus(ExecutorState.FINALIZED)
+        return self.get_status(job_definition)
 
     def terminate(self, job_definition):
         current_status = self.get_status(job_definition)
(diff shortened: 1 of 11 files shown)


jobrunner/executors/local.py
```diff
@@ -318,7 +318,7 @@ def get_status(self, job_definition, timeout=15, cancelled=False):
                         metrics=metrics,
                         results=job_metadata,
                     )
-                else:
+                else:  # pragma: no cover
```
inglesp, line 321: Why can't this be tested?

tests/agent/test_agent.py
```diff
@@ -159,32 +159,65 @@ def check_env(definition):
...
         main.handle_single_task(task, api)
 
-    mock_update_controller.assert_called_with(
-        task=task,
-        stage=ExecutorState.ERROR.value,
-        results={},
-        complete=True,
+    assert mock_update_controller.call_count == 1
+    task, stage, results, complete = mock_update_controller.call_args[0]
+    assert task == task
```
inglesp, line 176: This is never going to fail!

tests/agent/test_agent.py
```diff
@@ -159,32 +159,65 @@ def check_env(definition):
...
+    assert stage == ExecutorState.ERROR.value
+    assert results["error"]["exception"] == "Exception"
+    assert results["error"]["message"] == "foo"
+    assert "traceback" in results["error"]
+    assert complete == complete
```
inglesp, line 181: And nor is this.

tests/agent/test_agent.py
```diff
@@ -159,32 +159,65 @@ def check_env(definition):
...
+
+    # canceljob handler always calls update first
+    assert mock_update_controller.call_count == 2
+    task, stage, results, complete = mock_update_controller.call_args[0]
+    assert task == task
+    assert stage == ExecutorState.ERROR.value
+    assert results["error"]["exception"] == "Exception"
+    assert results["error"]["message"] == "foo"
+    assert "traceback" in results["error"]
+    assert complete == complete
```
inglesp, line 214: More tautologous assertions.

tests/agent/test_agent.py
bloodearnest: bah

rebkwok (approved):
Left a couple of minor comments. It might be worth adding an integration test (or adding to the existing one) to test that the controller correctly handles a job that ends up in the ERROR state

PR #869, opened 2025-04-16
Delete all the old things

### LINKED ISSUE #868: Remove old run.py code

rebkwok:
Fixes #868 
Delete run.py and associated tests
Delete some other unused bits of code
Add a few more tests for missing coverage

The few lines still not covered should be dealt with by #867


bloodearnest (approved):

rebkwok:
I think we probably can remove LoggingExecutor/RecordingExecutor, if they were only for the multiple executors. I wonder if this is one of those cases where we should avoid trying to improve job-runner in this phase though.
I might see if I can do it in <30 mins and abandon otherwise...

PR #880, opened 2025-04-22
Add backend to job model

### LINKED ISSUE #879: add a `backend` attribute to Job
and use it to set Task.backend when the controller creates tasks

rebkwok:
Fixes #879 

- adds a backend attribute to Job
- and to JobRequest (not a db object, but it's convenient to include it as a representation of the job-server job request)
- job-server already includes the backend in the job request response, so we can use it now when creating jobs in `sync`
- Use the new job.backend to set the task's backend in the controller loop
- Add a cli command to add the backend to existing jobs based on their SavedJobRequest, or if not available, the config.BACKEND; this is intended to run once only per backend before we move the controller and the jobs DB out of the backends and merge them

Note this deliberately doesn't do any filtering of jobs by backend yet (for determining dependencies etc). It doesn't need it while the db is still backend-specific and it will need the backend for existing jobs populated first.

### DIFF
--- a/jobrunner/cli/add_backend_to_job.py
+++ b/jobrunner/cli/add_backend_to_job.py
@@ -1,5 +1,5 @@
 from jobrunner import config
-from jobrunner.lib.database import find_all, find_where, update
+from jobrunner.lib.database import find_where, update
 from jobrunner.models import Job, SavedJobRequest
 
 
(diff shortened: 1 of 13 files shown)


jobrunner/cli/add_backend_to_job.py
```diff
@@ -0,0 +1,36 @@
...
+
+def main():
+    """
+    Command to add missing backend attribute to jobs. We expect this to only
+    run once per backend, prior to moving the controller out of the backend
+    """
+    jobs = find_all(Job)
+    jobs_missing_backend = find_where(Job, backend=None)
+    if len(jobs) > len(jobs_missing_backend):
+        print("This command has already been run, please confirm you want to re-run:")
```
inglesp, line 14: Is this the case?  Once we've merged this PR, won't we start adding backends to jobs?  So unless we run this command immediately after merging and before we pick up any new jobs from job-server, there will be some jobs with backends, and so `jobs_missing_backends` won't be the same as `jobs`.  Or have I misunderstood?

inglesp (approved):

jobrunner/cli/add_backend_to_job.py
rebkwok: Hmm, yes, that is true. I just wanted for it to warn you if you might be re-running it with a jobs db that's not inside a backend (re-running it on an in-backend db won't do anything bad). I can re-word the warning; once it's been run on all the separate in-backend dbs, we can get rid of it anyway.

rebkwok: I can skip the comparison actually, and just print a warning and require confirmation

PR #896, opened 2025-04-23
Split config between agent and controller

### LINKED ISSUE #886: Split config between agent and controller

rebkwok:
Fixes #886

Splits config into agent, controller and common

`common` is intended to contain only config that both agent and controller need, and which isn't only used in the context of a job (i.e. can't be retrieved from a task definition's env by the agent). 

I've tried to limit this PR to just splitting the config and updating import as they are in order to get tests to pass, so in some places agent code is importing controller config or vice versa; I plan to address those in later PRs


jobrunner/config/controller.py
```diff
@@ -0,0 +1,135 @@
...
+POLL_INTERVAL = float(os.environ.get("POLL_INTERVAL", "5"))
+
+
+# TODO Add a BACKENDS config and validate each job is from a known backend
+# We'll also use the BACKENDS config for looping through BACKENDS in sync
+BACKENDS = []
+
+truthy = ("true", "1", "yes")
+
+# TODO: Both need this; the controller can pass to the agent in the task
```
rebkwok, line 32: My next PR will address this TODO and make this an agent-only concern

PR #898, opened 2025-04-24
Move dummy data backend config to agent only

rebkwok:
- Make `USING_DUMMY_DATA_BACKEND` agent-only config
  The controller will soon become responsible for all backends so logic about whether the current backend is a dummy data one belongs with the agent only. Also removed references to the no-longer-active "expectation" backend
- Moves database URL validation to the agent and just validates database name is one of the allowed ones ("default" or "allow_t1oo") in the controller
- Removes code that just handled modifiying dummy data args for cohort-extractor actions

### DIFF
--- a/dotenv-sample
+++ b/dotenv-sample
@@ -1,5 +1,6 @@
 # The name of this backend
-BACKEND=expectations
+BACKEND=local
+USING_DUMMY_DATA_BACKEND=True
 
 # The endpoint to poll for jobs
 JOB_SERVER_ENDPOINT=https://jobs.opensafely.org/jobs/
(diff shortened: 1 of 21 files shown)


dotenv-sample
```diff
@@ -1,5 +1,6 @@
 # The name of this backend
-BACKEND=expectations
+BACKEND=local
```
inglesp, line 2: Should this be `BACKEND=dummy`?

inglesp (approved):

dotenv-sample
rebkwok: It can be anything really; for local envs it doesn't matter, and tests set it explicitly when they need to. In an imminent PR it's going to need to be one of a fixed list, because we'll need to know the backends to iterate over and fetch from job-server, so it may change then anyway

Issue #883, opened 2025-04-23
Remove JobStatus.message


bloodearnest:
This is no longer used for anything meaningful - it is not sent to the controller. We set it in various place, but only time it is used is in finalize_job, where it is written to metadata.json, and serves as some agent provided detail about the final status of job. We no longer need it to provide interim progress messages.

PR #902, opened 2025-04-24
Remove JobStatus.message

Providence-o:
Fixes: #883


jobrunner/executors/local.py
```diff
@@ -132,7 +132,6 @@ def prepare(self, job_definition):
             if workspace_is_archived(job_definition.workspace):
                 return JobStatus(
                     ExecutorState.ERROR,
-                    f"Workspace {job_definition.workspace} has been archived. Contact the OpenSAFELY tech team to resolve",
```
inglesp, line 137: We should add a log message noting that the workspace has been archived.

jobrunner/executors/local.py
```diff
@@ -213,9 +211,9 @@ def execute(self, job_definition):
                 volume_type=volumes.volume_type,
             )
 
-        except Exception as exc:  # pragma: no cover
+        except Exception:  # pragma: no cover
```
inglesp, line 214: We should log this exception.

jobrunner/executors/local.py
```diff
@@ -235,8 +233,10 @@ def finalize(self, job_definition, cancelled=False, error=None):
 
         try:
             finalize_job(job_definition, cancelled, error=error)
-        except LocalDockerError as exc:  # pragma: no cover
-            return JobStatus(ExecutorState.ERROR, f"failed to finalize job: {exc}")
+        except LocalDockerError:  # pragma: no cover
```
inglesp, line 236: We should log this exception.

inglesp (changes requested):
Looks good.  One small change please: every time we return a `JobStatus` with `ExecutorState.ERROR`, we should make sure that we've logged the exception.  We already do this in some cases and I think I've made a note for the places where we don't do this, but please double check.

Providence-o:
We don't raise an exception here, should a message still be logged? `local.py, def get_status, line 301`
```
if job_metadata.get("error"):
    return JobStatus(
        ExecutorState.ERROR,
        timestamp_ns=job_metadata["timestamp_ns"],
        metrics=metrics,
        results=job_metadata,
    )
```

inglesp (approved):

PR #912, opened 2025-04-25
Don't fail jobs on transient errors

### LINKED ISSUE #908: Treat "database locked" exceptions as transient errors, rather than job failures
Related discussion:
https://bennettoxford.slack.com/archives/C069YDR4NCA/p1745567684545229
https://docs.google.com/document/d/1X-_xOMvMYiPfm9CN9x5Mq735ySWlSGa0huXkFCoYBvk/

evansd:
For now the only errors we treat as transient are "database locked" errors but we may expand this in future.

Closes #908

### DIFF
--- a/jobrunner/agent/main.py
+++ b/jobrunner/agent/main.py
@@ -71,7 +71,11 @@ def handle_single_task(task, api):
         except Exception as exc:
             span.set_status(trace.Status(trace.StatusCode.ERROR, str(exc)))
             span.record_exception(exc)
-            if not is_transient_error(exc):
+            if error_type := get_transient_error_type(exc):
+                span.set_attributes(
+                    {"transient_error": True, "transient_error_type": error_type}
+                )
+            else:
                 mark_task_as_error(
                     api,
                     task,
@@ -86,10 +90,9 @@ def handle_single_task(task, api):
             raise
 
 
-def is_transient_error(exc):
+def get_transient_error_type(exc):
     if is_database_locked_error(exc):
-        return True
-    return False
+        return "db_locked"
 
 
 def handle_cancel_job_task(task, api):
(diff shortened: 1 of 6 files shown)


rebkwok (approved):

jobrunner/agent/main.py
```diff
@@ -74,11 +75,12 @@ def handle_single_task(task, api):
         except Exception as exc:
             span.set_status(trace.Status(trace.StatusCode.ERROR, str(exc)))
             span.record_exception(exc)
-            mark_task_as_error(
-                api,
-                task,
-                sys.exc_info(),
-            )
+            if not is_transient_error(exc):
```
bloodearnest, line 78: Is there a way we could tag the current span with an attribute to indicate non-fatal platform error?

Currently, we alert from INTERNAL_ERROR - db locked errors will rightly no longer trigger INTERNAL_ERROR with this change, so it would be good to beable to easily filter for such errors and alert on them?

It may be that setting the status to ERROR, is enough perhaps.  But maybe we want to tag db locked errors with a `db_locked=true` attribute, for easy finding?

bloodearnest: So, thinking about it, we could potentially have enough already.

We're marking the span as having status error, with the str(exc) as the message.

In theory, that should enough for us to filter on it and check for "database locked" in the error message in HC.

But it's a bit more involved, so another way would bet to get the current span get_current_span() inside is_transient_error, and add two attributes

1. transient_error=True (for all future transient errors)
2. db_locked=True (for specifically this error)

That will make it trivial for us to filter in HC.

Adding extra attributes are free, we should generally be bullish on adding them, I think.

evansd: Would it work to change `is_transient_error()` to `get_transient_error_type()` (returning `None` for non-transient errors) and then annotate the span inside error handler? That would mean having something like `transient_error_type=db_locked` rather than `db_locked=True`, but I'm assuming that's OK from the Honeycomb point-of-view.

bloodearnest: yeah, that's better

bloodearnest (approved):

PR #913, opened 2025-04-28
Dont return JobStatus from executor methods

inglesp:


tests/test_local_executor.py
```diff
@@ -581,12 +607,15 @@ def test_finalize_failed_oomkilled(docker_cleanup, job_definition, tmp_work_dir)
 
     api = local.LocalDockerAPI()
 
-    status = api.prepare(job_definition)
-    status = api.execute(job_definition)
+    api.prepare(job_definition)
+    status = api.get_status(job_definition)
+    api.execute(job_definition)
+    status = api.get_status(job_definition)
```
rebkwok: Do we need to call `get_status()` after `prepare` and `execute`, if we're not asserting anything about it until after the `finalize` call?

rebkwok (approved):

rebkwok (approved):

PR #916, opened 2025-04-28
Measure transactions and slow function

bloodearnest:
- **Add telemetry for transactions**
- **Add explicit telemetry for suspected slow function**

### DIFF
--- a/jobrunner/lib/database.py
+++ b/jobrunner/lib/database.py
@@ -9,6 +9,7 @@ surface area of this module is sufficiently small that swapping it out
 shouldn't be too large a job.
 """
 
+import contextlib
 import dataclasses
 import json
 import logging
@@ -16,6 +17,8 @@ import sqlite3
 import threading
 from enum import Enum
 from pathlib import Path
+
+from opentelemetry import trace
 
 from jobrunner.config import controller as config
 
@@ -25,6 +28,9 @@ log = logging.getLogger(__name__)
 CONNECTION_CACHE = threading.local()
 TABLES = {}
 MIGRATIONS = {}
+
+
+tracer = trace.get_tracer("db")
 
 
 def databaseclass(cls):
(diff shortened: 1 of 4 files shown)


jobrunner/lib/database.py
```diff
@@ -150,14 +156,18 @@ def select_values(itemclass, column, **query_params):
...
     # See: https://docs.python.org/3/library/sqlite3.html#using-the-connection-as-a-context-manager
     # We're relying here on the fact that because of the lru_cache,
     # `get_connection` actually returns the same connection instance every time
-    conn = get_connection()
-    conn.execute("BEGIN IMMEDIATE")
-    return conn
+    span_context = (
+        tracer.start_as_current_span("TRANSACTION") if trace else nullcontext()
+    )
+    with span_context:
```
evansd, line 167: I'm not sure this is going to work as is because the existing code doesn't define a context manager: it relies on the fact that the connection object [acts as a context manager](https://docs.python.org/3/library/sqlite3.html#how-to-use-the-connection-context-manager). So the trace code will make the transaction as finished as soon as this function returns, which is incorrect.

I think we might need to define an explicit context manager to make this work, unless there's another trick I'm missing.

jobrunner/queries.py
```diff
@@ -3,10 +3,15 @@
 from itertools import groupby
 from operator import attrgetter
 
+from opentelemetry import trace
+
 from jobrunner.lib.database import find_all, find_one, find_where, upsert
 from jobrunner.models import Flag, Job
 
 
+tracer = trace.get_tracer("db")
```
evansd, line 12: I'm guessing this belongs in the next commit as it's not used here (not that it really matters)?

jobrunner/lib/database.py
bloodearnest: Hmm, that is a good point.

I started with a @contextmanager, and then removed it, but yeah this is clearly not going to measure the right thing, good catch.

evansd (approved):

PR #919, opened 2025-04-28
Add wrapper to agent_main thread

### LINKED ISSUE #909: Add top level wrapper around agent.main()
We start various threads in service.py - the sync thread, the maintenance thread, the stats thread, and now the new agent thread.

The sync, maintenance and stats threads all have wrapper functions, who's purpose is to run the the thread's target function in a while loop, catching and logging any exceptions, then sleeping for a while.

However, when we added the agent thread, we neglected to add this wrapper, so if the agent raises and exception, the thread stops.  

We need to add an agent_wrapper function, similar to the other wrapper functions, that calls agent.main() and handles and logs any exceptions. Then start_thread should run the agent_wrapper instead of agent.main

There's an optional opportunity for some clean up here. The wrapper functions are all nearly identical. We could replace them with a generic thread_wrapper factory function, something like:

```
def thread_wrapper(name, func, loop_interval):
    def wrapper():
        while True:
            try:
                func()
            except Exception:
                log.exception(f"Exception in {name} thread")

            time.sleep(loop_interval)
    return wrapper

start_thread(thread_wrapper("agent", agent.main, config.JOB_LOOP_INTERVAL))
```

This should work for sync, maintenance, and agent threads, but may not for the stats thread, due the return semantics.

Providence-o:
Fixes #909

### DIFF
--- a/jobrunner/service.py
+++ b/jobrunner/service.py
@@ -46,40 +46,44 @@ def main():
 
     try:
         log.info("jobrunner.service started")
-        start_thread(sync_wrapper, "sync")
+
+        sleep_after_error = config.POLL_INTERVAL * 5
+        start_thread(thread_wrapper("sync", sync_wrapper, sleep_after_error), "sync")
         start_thread(record_stats_wrapper, "stat")
         if config.ENABLE_MAINTENANCE_MODE_THREAD:
-            start_thread(maintenance_wrapper, "mntn")
-        start_thread(agent_wrapper, "agnt")
+            start_thread(
+                thread_wrapper(
+                    "maintenance_mode",
+                    maintenance_mode,
+                    config.MAINTENANCE_POLL_INTERVAL,
+                ),
+                "mntn",
+            )
+        start_thread(thread_wrapper("agent", agent_main, config.POLL_INTERVAL), "agnt")
         controller_main()
     except KeyboardInterrupt:
         log.info("jobrunner.service stopped")
 
 
-def agent_wrapper():
-    """Wrap the agent_main call with logging context and an exception handler."""
-    while True:
-        try:
-            agent_main()
-        except Exception:
-            log.exception("Exception in agent thread")
-            time.sleep(config.POLL_INTERVAL)
+def thread_wrapper(name, func, loop_interval):
+    def wrapper():
+        while True:
+            try:
+                func()
+            except Exception:
+                log.exception(f"Exception in {name} thread")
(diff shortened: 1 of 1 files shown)


jobrunner/service.py
```diff
@@ -45,41 +45,45 @@ def main():
...
 
 
 def sync_wrapper():
-    """Wrap the sync call with logging context and an exception handler."""
-    # avoid busy retries on hard failure
     sleep_after_error = config.POLL_INTERVAL * 5
     while True:
         try:
             sync.main()
         except sync.SyncAPIError as e:
```
rebkwok, line 84: Have we lost this special handling of the SyncAPIError with the generic wrapper function?

rebkwok: Oh, no, we haven't, nvm

rebkwok: But I'm finding this wrapped wrapper quite hard to parse.  Do you think it's worth it for the amount of code we avoid repeating?

rebkwok: You could move the handling of the `SyncAPIError` up into the `thread_wrapper` (it's only raised in the sync loop, so although it'd catch it in any of the other uses of the `thread_wrapper`, it'd only actually happen for `sync`). And then `sync()` can just be wrapped in the `thread_wrapper` in the same way as the other two

rebkwok: (that would also mean you don't need to redefine `sleep_after_error`)

bloodearnest: I don't think we need to preserve the special handling, and log the full exception.

If we need to, we can catch SyncAPI error within the sync thread, perhaps.

Providence-o: @bloodearnest Is this different from @rebkwok's suggestion ? from what i understand, if  `SyncAPIError` is handled in `thread_wrapper` there wouldn't be a need for `sync_wrapper` function. 

```
def thread_wrapper(name, func, loop_interval):
    def wrapper():
        while True:
            try:
                func()
            except sync.SyncAPIError as e:
                log.error(e)
                time.sleep(loop_interval)
            except Exception:
                log.exception(f"Exception in {name} thread")
                time.sleep(loop_interval)

    return wrapper
```

Providence-o: > I think we could make it easier perhaps by creating the wrapper inside start_thread?

This would make it tricky to handle the `record_stats` thread

rebkwok (approved):

PR #921, opened 2025-04-28
Use GLOB rather than LIKE for partial ID matches

evansd:
SQLite's LIKE queries are case-insensitive [by default](https://www.sqlite.org/lang_expr.html#the_like_glob_regexp_match_and_extract_operators). This means that left-anchored prefix queries such as:
```
my_col LIKE 'some_prefix_%'
```
which you might expect to make use of an index on `my_col` will in fact always use a sequential scan.

SQLite also provides the GLOB operator which is very similar to LIKE but uses `*` and `?` instead of `%` and `_` and, more importantly, is case sensitive by default.

It is possible to change LIKE to be case sensitive using a pragma, but the docs very strongly [discourage this](https://www.sqlite.org/pragma.html#pragma_case_sensitive_like):
> This pragma is deprecated and exists for backwards compatibility only. New applications should avoid using this pragma. Older applications should discontinue use of this pragma at the earliest opportunity.

So switching to use GLOB seems like a better solution here.

Below is a SQLite session demonstrating the difference in query plans:
```
$ sqlite3

SQLite version 3.37.2 2022-01-06 13:25:41

sqlite> CREATE TABLE job (id TEXT PRIMARY KEY);

sqlite> EXPLAIN QUERY PLAN SELECT * FROM job WHERE id LIKE 'abc%';
QUERY PLAN
`--SCAN job

sqlite> EXPLAIN QUERY PLAN SELECT * FROM job WHERE id GLOB 'abc*';
QUERY PLAN
`--SEARCH job USING COVERING INDEX sqlite_autoindex_job_1 (id>? AND id<?)

sqlite> PRAGMA case_sensitive_like = true;
sqlite> EXPLAIN QUERY PLAN SELECT * FROM job WHERE id LIKE 'abc%';
QUERY PLAN
`--SEARCH job USING COVERING INDEX sqlite_autoindex_job_1 (id>? AND id<?)
```

### DIFF
--- a/jobrunner/cli/kill_job.py
+++ b/jobrunner/cli/kill_job.py
@@ -57,7 +57,7 @@ def get_jobs(partial_job_ids):
     need_confirmation = False
     for partial_job_id in partial_job_ids:
         # look for partial matches
-        partial_matches = database.find_where(Job, id__like=f"%{partial_job_id}%")
+        partial_matches = database.find_where(Job, id__glob=f"*{partial_job_id}*")
         if len(partial_matches) == 0:
             raise RuntimeError(f"No jobs found matching '{partial_job_id}'")
         elif len(partial_matches) > 1:
--- a/jobrunner/cli/retry_job.py
+++ b/jobrunner/cli/retry_job.py
@@ -41,7 +41,7 @@ def main(partial_job_id):
 
 
 def get_job(partial_job_id):
-    matches = find_where(Job, id__like=f"%{partial_job_id}%")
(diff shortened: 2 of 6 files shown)


rebkwok (approved):
Yet another sqlite gotcha I didn't know about.

There's a couple of other `id__like` in cli/retry_job.py and cli/kill_job.py; those commands are going to need to change, but is it worth changing the like to glob there just so we don't forget when we get around to looking at them?

PR #926, opened 2025-04-28
Per-backend flags

### LINKED ISSUE #922: Update `flags` (for paused and db-maintenance mode) to be per-backend

rebkwok:
Fixes #922

Update the flags table and queries to always filter by a backend.

Punts on how the db-maintenance flag will be set for now and just uses the agent's BACKEND setting.

Also adds a cli command to add the current backend to any existing flags, intended to be run while the controller is still inside the backend and removed later (along with the add_backend_to_job command)

### DIFF
--- a/jobrunner/sync.py
+++ b/jobrunner/sync.py
@@ -92,13 +92,13 @@ def api_request(method, path, *args, backend, headers=None, **kwargs):
 
     url = "{}/{}/".format(config.JOB_SERVER_ENDPOINT.rstrip("/"), path.strip("/"))
 
-    flags = {
-        f.id: {"v": f.value, "ts": f.timestamp_isoformat}
-        for f in queries.get_current_flags()
-    }
-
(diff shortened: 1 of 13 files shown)


jobrunner/sync.py
```diff
@@ -72,15 +72,15 @@ def sync():
     api_post("jobs", json=jobs_data)
 
 
-def api_get(*args, **kwargs):
-    return api_request("get", *args, **kwargs)
+def api_get(*args, backend=None, **kwargs):
```
inglesp, line 81: Why is backend an optional argument here?

inglesp (approved):

jobrunner/sync.py
rebkwok: Because I'd forgotten how to make it a required keyword arg. This branch had also got out of sync, so I've fixed it in the sync PR

PR #939, opened 2025-05-02
Robust metadata

### LINKED ISSUE #931: Gracefully handle missing keys in metadata.json
We have metadata.json files for all jobs but we have not consistently migrated old files as we have changed the metadata structure.  This is usually not a problem, but caused issues on deployment of #857 where we saw lots of `KeyError`s caused by new code expecting to find keys in metadata.json that were missing.

As such, we would like to gracefully handle missing keys in metadata.json.

This behaviour would be localised at the point where metadata is read, and not defensively spread out across the codebase.

bloodearnest:
Ensure that job metadata always has all expected keys present.

They maybe Falsey or have default values in some cases, like when reading
metadata.json generated by older version of job-runner, but they will always be
present.

The goal is to always be able to assume presence and use `metadata["key"]`
rather than `metadata.get(key)`, thus simplifying usage.

I chose to just deal with the expected issues we have with the migration to the split, in that only some things will be missing from old version of written metadata.json, which were only written on a successful finalize. I anticipate we'll make further changes to this later, I think we can do a better job of typing and defining job metadata with some refactoring

Fixes #931

### DIFF
--- a/jobrunner/executors/local.py
+++ b/jobrunner/executors/local.py
@@ -479,10 +479,13 @@ def finalize_job(job_definition, cancelled, error=None):
         base_created=labels.get("org.opencontainers.base.build-date", "unknown"),
     )
     job_metadata = get_job_metadata(
-        job_definition, outputs, container_metadata, results, cancelled=cancelled
+        job_definition,
+        outputs,
+        container_metadata,
+        results,
+        cancelled=cancelled,
+        error=error,
     )
-    if error:
-        job_metadata["error"] = error
 
     if cancelled or error:
         if container_metadata:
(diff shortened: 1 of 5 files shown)


jobrunner/executors/local.py
```diff
@@ -533,6 +537,22 @@ def get_job_metadata(
     return job_metadata
 
 
+# keys that may be missing in older metadata files, and their default empty values
+METADATA_DEFAULTS = {
+    "hint": None,
+    "unmatched_patterns": [],
+    "unmatched_outputs": [],
```
inglesp, line 548: Is this going to bit us in the bum like mutable default arguments?

That is, if we end up using these lists because the metadata for a job is missing one of these keys, is there any risk that we'll mutate the list for one job and then accidentally mutate the defaults too?

inglesp (approved):

jobrunner/executors/local.py
bloodearnest: Hmm, it could do. Empty tuple, so its always iterable?

evansd (approved):

PR #947, opened 2025-05-06
Always sleep in thread wrapper

### LINKED ISSUE #946: `thread_wrapper` should sleep after a clean exit
The `thread_wrapper()` function correctly sleeps after an exception from the inner function:
https://github.com/opensafely-core/job-runner/blob/fd6755b23ea7aa7a7307a616b80f18ceb12bc99f/jobrunner/service.py#L67-L79

But it does not sleep if the inner function returns cleanly with no exception.

For the `sync` and `agent` threads this makes no difference because they contain their own infinite loops and so will only ever terminate with an exception. But the `maintenance_mode()` function is one-shot and relies and being continually re-run by the wrapper thread. The above behaviour means that we run it much more frequently than we intend to.

bloodearnest:
Fixes #946, with some refactoring on the way

### DIFF
--- a/jobrunner/service.py
+++ b/jobrunner/service.py
@@ -48,7 +48,10 @@ def main():
 
         sleep_after_error = config.POLL_INTERVAL * 5
         start_thread(thread_wrapper("sync", sync.main, sleep_after_error), "sync")
-        start_thread(record_stats_wrapper, "stat")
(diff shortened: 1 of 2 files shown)


evansd (approved):
Some very pleasing drive-by refactoring here!

PR #950, opened 2025-05-07
Add a DBSTATUS task type

### LINKED ISSUE #889: Add a new TaskType which runs the database maintenance check
This can borrow from the existing code in the database maintenance thread/service to invoke Cohort Extractor correctly. However this won't do any updating of flags itself; it just reports its results via the task API.

evansd:
This will be used by the controller to check whether we are currently in database maintenance mode.

The [recent changes](https://github.com/opensafely-core/cohort-extractor/pull/1038) in Cohort Extractor means that the logic here can be simpler than that used in the existing check.

We also tighten up the network security here so that the status check runs with the same configuration as pipeline database job i.e. it can access the database but nothing else. This is obviously less critical as there is no user-supplied code involved, but it's good to restrict access to the minimum required.

Closes #889


jobrunner/agent/main.py
```diff
@@ -320,6 +323,68 @@ def inject_db_secrets(job):
         job.env["EMIS_ORGANISATION_HASH"] = config.EMIS_ORGANISATION_HASH
 
 
+def handle_simple_task(task_function, task):
+    """
+    A "simple" task function is one which takes keyword arguments as supplied in the
+    Task definition and returns a dictionary which will be reported under the `results`
+    key of the Task results. Any exceptions are caught and reported under the `error`
+    key of the taks results.
```
madwort, line 331:
```suggestion
    key of the task results.
```

PR #952, opened 2025-05-07
remove 39 backport

bloodearnest:
- **remove unneeded backport**
- **Use cached_property now we are on 3.10**

### DIFF
--- a/jobrunner/models.py
+++ b/jobrunner/models.py
@@ -15,7 +15,7 @@ import hashlib
 import secrets
 import shlex
 from enum import Enum
-from functools import total_ordering
+from functools import cached_property, total_ordering
 
 from jobrunner.lib.database import databaseclass, migration
 from jobrunner.lib.string_utils import slugify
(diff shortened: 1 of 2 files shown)


jobrunner/models.py
```diff
@@ -325,9 +325,7 @@ def started_at_isoformat(self):
     def completed_at_isoformat(self):
         return timestamp_to_isoformat(self.completed_at)
 
-    # On Python 3.8 we could use `functools.cached_property` here and avoid
-    # recomputing this every time
-    @property
+    @cached_property
     def slug(self):
```
evansd, line 329: I think we could actually get rid of `slug` altogether. I don't think we use it anywhere now.

bloodearnest: hilariously, its used in kill_job and retry_job.

But yeah, may do that as a follow on

bloodearnest:
sorry, had to rebase on main, so this need re-review :facepalm:

bloodearnest:
no it doesn't wat
sigh

PR #954, opened 2025-05-07
Regularly update the timestamps of running jobs

evansd:
This restores behaviour which we lost in the agent/controller split. It allows us to provide confirmation to the user that their job is actually still running.

See:
https://bennettoxford.slack.com/archives/C069YDR4NCA/p1746626176239809


rebkwok (approved):
Is it worth adding a test to confirm the timestamps are updated as expected? There are tests for the status code timing (I think I added them) but they only test the unchanged status code in the PENDING state. The RUNNING state is maybe covered in the integration test

PR #960, opened 2025-05-07
Split CLI module into agent and controller commands

### LINKED ISSUE #955: Reorganise CLI commands into agent/controller
Controller:
- add-job
- migrate
- prepare-for-reboot
- flags

Agent:
- add-backend-to-job (will be removed later)
- add-backend-to-flags (will be removed later)

rebkwok:
Fixes #955


jobrunner/cli/add_backend_to_flags.py
```diff
@@ -1,30 +1,5 @@
...
-            return
-    else:
-        print("All flags have a backend assigned; nothing to do")
-
-    for flag in flags_missing_backend:
-        flag.backend = config.BACKEND
-        update(flag)
-
-    print(f"{len(flags_missing_backend)} flags updated")
+from jobrunner.cli.agent import add_backend_to_job
```
inglesp, line 1: Shouldn't this be `add_backend_to_flags`?

inglesp (dismissed):
Approve of the approach but I think there's an important typo that needs a fix!

jobrunner/cli/add_backend_to_flags.py
rebkwok: Oops. Yes, it should!

rebkwok: Fixed

Issue #891, opened 2025-04-23
Controller needs to handle the results of "db maintenance check" tasks and set flags appropriately


evansd:
For now, this needs to be done in the `update_controller()` function:
https://github.com/opensafely-core/job-runner/blob/27338a78a838e9c0abc54f4be72b0b7bf59c1c61/jobrunner/agent/task_api.py#L14-L19

This currently plays the role of what will be the controller's HTTP endpoint, which is the place we'll handle this in future.

Doing this properly depends on having per-backend db-maintenance flags (see issue below) but I suppose for now we could get away with just setting the single global flag.

evansd:
Closed by
 * #965

PR #965, opened 2025-05-08
Automatically schedule DBSTATUS tasks and handle the results

### LINKED ISSUE #890: Controller needs to regularly dispatch a "db maintenance check" task to the TPP backend
I think the basic logic would be:
```
Are we in manual maintenance mode? If so, mark any running maintenance tasks as inactive and exit.

Otherwise, is there a `DBMaintenanceCheck` task on the TPP backend which is either:
 * not completed; or
 * was completed less than N seconds ago?

If not then create a new task, otherwise do nothing.
```

For now, we should set N to however many seconds the current poll interval is set to.

There's an open question as to how the above check gets triggered. Probably the easiest thing for now is to call it as part of the controller run loop. It's easy enough to move it around later if we find somewhere better.

evansd:
This replaces the maintenance mode thread with a sequence of regularly scheduled tasks.

Closes #890 #891


tests/controller/test_main.py
```diff
@@ -763,3 +763,58 @@ def test_handle_transient_error(patched_handle_job, db, monkeypatch, exc, error_
...
+    assert len(tasks) == 1
+
+    # Set the finished time further back in the past
+    task = database.find_one(Task, type=TaskType.DBSTATUS)
+    task.finished_at = task.finished_at - 10000
+    database.update(task)
+
+    # After ticking time forward a significant amount, running the loop should now
+    # create a new active task
+    freezer.tick(delta=10000)
```
rebkwok, line 810: Do we need to do both of these (manually setting the finished time back and ticking forward)? Wouldn't just ticking forward by an amount that's more than the poll interval be sufficient?

rebkwok (approved):

tests/controller/test_main.py
evansd: Oh, sorry. That's pointless code left from an earlier iteration. Thanks for spotting!

rebkwok (approved):

PR #966, opened 2025-05-08
RUNJOB controller test factory refactor

bloodearnest:
Extracted this from my spike, as it seems a worthy change.

### DIFF
--- a/tests/agent/stubs.py
+++ b/tests/agent/stubs.py
@@ -3,7 +3,11 @@ from collections import defaultdict
 
 from jobrunner.job_executor import ExecutorState, JobDefinition, JobStatus
 from jobrunner.schema import AgentTask
-from tests.factories import canceljob_db_task_factory, runjob_db_task_factory
+from tests.factories import (
+    canceljob_db_task_factory,
+    job_factory,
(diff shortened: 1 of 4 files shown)


jobrunner/controller/main.py
```diff
@@ -533,7 +533,7 @@ def update_job(job):
     update(job, exclude_fields=["cancelled"])
 
 
-def create_task_for_job(job):
+def create_task_for_job(job, **kwargs):
```
evansd, line 536: A micro thing, but I'm not sure it makes sense to be changing the prod code here just to make test setup more convenient. Given that this function returns the object and doesn't insert it, can't the factory function modify it afterwards to apply the `kwargs`?

evansd (approved):
:+1: Constructing the job and passing it in looks a lot better to me than fishing it out of the database afterwards.

jobrunner/controller/main.py
bloodearnest: Yeah, it can.

I originally tried using the whole create-then-insert-in-transaction logic, which made that harder. When I went to just using create_task_for_job, then I missed that that could work.

Issue #964, opened 2025-05-08
Tasks HTTP API - authentication


rebkwok:
#963  implements the basic HTTP API but doesn't implement authentication. We'll need backends to provide some form of authentication to the controller endpoints.

Job-server handles authentication in calls from jobrunner in the form of a token, [associated with a backend](https://github.com/opensafely-core/job-server/blob/main/jobserver/api/jobs.py#L251). Backends don't call with a plain text backend parameter.

We aren't planning to use DRF, and we don't have a Backend model to store auth_tokens on the way that job-server does, but we could use a similar system with tokens per backend in environment variables.

Issue #963, opened 2025-05-08
Tasks HTTP API - stage 1


rebkwok:
This involves essentially swapping out the agent's task api methods for http calls to the controller app.
We should plan to merge to a dev branch so we can test deployment (not in scope for this ticket) on the test backend first.

- [x] Get the django app running locally
     - Add some `just` commands for it
    - Update django settings and use env variables to configure for local vs prod run
- [x] Make Task model serializable to json
- [x] deserializable from Task.asdict() to an AgentTask
- [x] Add urls/views that do the equivalent of the two agent task_api functions
      - Get_active_tasks (to return serialized Tasks)
      - Update_controller
- [x] Add tests for the new views
- [x] Add agent config for controller endpoint
- [x] Wire up agent task api methods to use the new api endpoints
      - Handle error responses
      - Retries? Possibly not needed since the agent will automatically retry already
- [x] Update tests to use new http api (we may be able to use django's liveserver in tests that run the agent loop)

Note that we're deliberately avoiding authentication and deployment at this stage

PR #971, opened 2025-05-09
Task api part 1a

rebkwok:
The first part of #963 
- adds some environment variable configuration for the django app
- adds a serializer for Task
- adds the two endpoints that the agent will need to call to get active tasks for a backend and to update a task

Note that we've currently put the backend in the url scheme, and there's no authentication yet. This shouldn't be a problem because there's no way of calling these endpoints with any access to actual data yet. Depending on how we decide to do auth (see #964) we may change the url scheme, but we'll deal with that later.


controller_app/settings.py
```diff
@@ -21,12 +22,16 @@
 # See https://docs.djangoproject.com/en/5.1/howto/deployment/checklist/
 
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = "django-insecure-7x$$d9^(-rb2uj57*d%u#z68fh*@y8s)8zu(*_t-ck%5jmk)ef"
+# TODO add environment variables to dokku app
+# TODO Note: these env vars are labelled with the additional "JOBRUNNER" prefix
```
bloodearnest, line 26: How about CONTROLLER rather than JOBRUNNER? I think that maybe clearer as to ownership, and I wonder if we want to move away from the term job-runner in general?

controller_app/urls.py
```diff
@@ -5,4 +5,5 @@
 
 urlpatterns = [
     path("", views.index),
+    path("<str:backend>/tasks/", views.active_tasks, name="active_tasks"),
```
evansd, line 8: I actually prefer being explicit like this. Using the auth token to determine the backend feels to me like mixing concerns.

jobrunner/models.py
```diff
@@ -436,3 +436,11 @@ class Task:
 
     # ensure this table exists
     migration(4, __tableschema__)
+
+    def asdict(self):
+        data = dataclasses.asdict(self)
+        for key, value in data.items():
+            # Convert Enums to strings for straightforward JSON serialization
+            if isinstance(value, Enum):
+                data[key] = value.value
```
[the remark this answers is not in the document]
bloodearnest: I kinda with we'd done NAME = "NAME" for these (as opposed to NAME = "name"), so value and name were the same.  But I suppose there advantages to having them separate.

controller_app/views.py
```diff
@@ -1,8 +1,42 @@
...
 
 @csrf_exempt
 def index(request):
     response = JsonResponse({"method": request.method})
     return response
+
+
+def active_tasks(request, backend):
+    active_tasks = [task.asdict() for task in get_active_tasks(backend)]
+    return JsonResponse({"tasks": active_tasks})
```
bloodearnest, line 21: I think we should be serializing AgentTask's here, not db Tasks?

The point of AgentTask was a simplified serializable task definition with just the data the agent needs, not the db level stuff?

bloodearnest (approved):

controller_app/views.py
bloodearnest: The conversion of db Task -> AgentTask that happens in the current agent.task_api is an implementation detail of the fake db api, not a deliberate step that needs to happen on the agent side.

controller_app/settings.py
```diff
@@ -21,12 +22,16 @@
...
+# so as not to conlict with Airlock in production (where the env variables are
+# shared). They can be renamed once the app is running outside of the backends
+SECRET_KEY = os.environ["DJANGO_JOBRUNNER_SECRET_KEY"]
```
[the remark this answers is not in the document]
rebkwok: yes - I'll do that in the next stage

controller_app/settings.py
rebkwok: Sure

bloodearnest (approved):

PR #979, opened 2025-05-14
Redact filenames from results sent to controller

rebkwok:
- list job ids as inputs for a job definition rather than the filenames themselves
- use job ids to read past job metadata and retrieve the required input filenames from there
- redact any references to output files or patterns by default when reading job metadata
    - this includes `outputs`, `unmatched_outputs` and `unmatched_patterns`, and also (in the case of unmatched patterns/outputs only), `status_message` and `hint`, which contain references to the filenames or patterns
    - we always read job metadata again when retrieving the job status (which now holds the metadata as job_status.results), so redacting filenames by default means any time the agent calls get_status, it won't have filename data
    - just in case, there's also a new assertion that the status.results doesn't contain the key "outputs" before we update the controller
- writing job metadata is unchanged, so output files, patterns, messages etc are still saved to file and before and will appear in the logs that users can view on Airlock


jobrunner/executors/local.py
```diff
@@ -72,17 +72,34 @@ def get_log_dir(job_id):
     return config.JOB_LOG_DIR / month_dir / container_name(job_id)
 
 
-def read_job_metadata(job_id):
+def read_job_metadata(job_id, redact=True):
```
evansd, line 75: Rather than having a flag here I think it might be both simpler and clearer to factor this as:
```py
def read_job_metadata(job_id):
    metadata = read_job_metadata_unredacted(job_id)
    return redact_job_metadata(metadata)
```
Then code which needs access to the unredacted metadata can use the other function.

jobrunner/executors/local.py
```diff
@@ -72,17 +72,34 @@ def get_log_dir(job_id):
...
+    If there are unmatched outputs or patterns, we also redact the
+    status_message and hint which may contain messages referring to
+    the unmatched filenames or patterns.
+    """
+    for key in ["unmatched_outputs", "unmatched_patterns"]:
+        if metadata[key]:
+            metadata["status_message"] = ""
+            metadata["hint"] = ""
+        del metadata[key]
+    del metadata["outputs"]
```
evansd, line 100: My suggested code above relies on it not being an error to redact an empty dict, but I think that better captures the behaviour we want here. Using `metadata.pop(key,  None)` instead of `del metadata[key]` would do that.

jobrunner/job_executor.py
```diff
@@ -131,9 +131,9 @@ def from_dict(cls, metadata: dict):
         except (TypeError, ValueError):
             exit_code = None
         return cls(
-            outputs=metadata["outputs"],
-            unmatched_patterns=metadata["unmatched_patterns"],
-            unmatched_outputs=metadata["unmatched_outputs"],
+            outputs=metadata.get("outputs"),
+            unmatched_patterns=metadata.get("unmatched_patterns"),
+            unmatched_outputs=metadata.get("unmatched_outputs"),
```
evansd, line 136: This tingles my spider-senses a bit in that it indicates that we have JobResults which both do and don't contain file information and that doesn't seem good. I wonder if it's because we're using the same class in two places where we really want two classes. I'll need to poke around a bit more to understand what's happening here.

evansd (approved):
A surprisingly simple change! And a very clear PR to follow.

I've left a micro style suggestion, and then I'd still like to understand better what's going on with `JobResults`. But I don't think there's anything blocking here.

jobrunner/controller/main.py
```diff
@@ -348,7 +348,7 @@ def job_to_job_definition(job):
         image=full_image,
         args=action_args,
         env=env,
-        inputs=input_files,
```
bloodearnest, line 351: I wonder if we should use a different key, rather than changing the type of inputs? Thinking about tasks that are in-flight when this deploys, and potential breakage.

Or is the plan to cancel them all and get the controller to retry with the new job definition format post deploy?

jobrunner/executors/local.py
bloodearnest: Hmm, we need the unredacted data sometimes though? Like, to know which files to copy?

I think we should only redact at the system boundary - when we send it to the controller?

I think its fine for JobStatus.results to be the unredacted version - the agent loop may need to use the unredacted info, even it doesnt right now?

But agent.main.update_job_task can do the redaction before passing it up to the controller? That way, we *ensure* that the data is redacted. Currently, it's possible make mistakes in future an accidentally send an unredacted results up to the controller?

jobrunner/executors/local.py
```diff
@@ -72,17 +72,34 @@ def get_log_dir(job_id):
...
     if path:
         metadata = json.loads(path.read_text())
         # ensure all metadata has default values, as older json files might be
         # missing some
-        return METADATA_DEFAULTS | metadata
-
+        metadata = METADATA_DEFAULTS | metadata
+        if redact:
+            redact_job_metadata(metadata)
+        return metadata
```
bloodearnest, line 84: See comment above - what about doing this in agent.main.update_job_task?

jobrunner/job_executor.py
bloodearnest: I think that we should get rid of JobResults and just use what we need from Task.agent_results directly.

It's only still there because some the existing APIs used it, and we wanted to do minimal changes.  I think it should go at some point soon.

tests/test_local_executor.py
```diff
@@ -138,8 +138,21 @@ def test_read_metadata_path(job_definition):
 def test_prepare_success(
     docker_cleanup, job_definition, test_repo, tmp_work_dir, freezer
 ):
-    job_definition.inputs = ["output/input.csv"]
     populate_workspace(job_definition.workspace, "output/input.csv")
+    # this job requires input files from a previous job
```
bloodearnest, line 142: I wonder if we should update populate_workspace to require a job  to files mapping, and do this logic of creating past metadata.json for each job?

I think we'll need it a lot - its already used twice here.

jobrunner/controller/main.py
evansd: Yeah, a new key might be cleaner and more explicit here. Deployment is going to be a bit fiddly whichever way we cut it. But we don't have loads of activity at the moment so we may be able to get away without having to worry about in flight tasks.

rebkwok: I'll make a new key, but I think if we wanted to handle in flight tasks we'd need to check in `prepare() ` for either `inputs` or `input_job_ids` (or whatever) and do the file-copying differently depending on which the task definition has. I'm not sure it's worth it - with not much activity going on, we could either wait until there are no jobs running before deploying, or run a prepare-for-reboot to cancel ongoing ones

rebkwok:
This has been updated to:
- uses `input_job_ids` as the job definition key instead of reusing `inputs`
- redact in update_job_task
- also redacts `level4_excluded_files`
- send just True/False values for `has_unmatched_patterns` and `has_level4_excluded_files`, because the controller needs to use those in order to determine job success
- add a new `TaskResults` class which represents the redacted results that the agent sends back, and which the controller stores in the db for the job. So we now only use `JobResults` in the executor (punting on whether to remove it altogether for another time)
- create the dependent test job metadata in `populate_workspace`

bloodearnest (dismissed):
Nice, surprisingly simple in the end.

evansd (approved):
Yep, agreed. All looks good and easy to follow but probably worth getting the naming right here as Simon says.

jobrunner/schema.py
```diff
@@ -44,3 +44,70 @@ def asdict(self):
             if isinstance(value, Enum):
                 data[key] = value.value
         return data
+
+
+@dataclass
+class TaskResults:
```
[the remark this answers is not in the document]
evansd: Oh, I didn't actually realise CANCELJOB sent these. The controller ignores them completely at the moment. But `JobTaskResults` doesn't seem too bad as a name to me.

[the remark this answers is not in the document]
bloodearnest: Good point, JobTaskResults is good :+1:

evansd (approved):

PR #983, opened 2025-05-14
Adjust error telemetry

### LINKED ISSUE #980: Telemetry: ensure that we are getting allets on platform errors, even if the job is not failed.
Currently, the alert are based on INTERNAL_ERROR, which is a job StatusCode

https://ui.honeycomb.io/bennett-institute-for-applied-data-science/environments/production/datasets/jobrunner/triggers/32TyFDooh2p

We are no longer failing jobs with INTERNAL_ERROR status codes when handling platform errors.

So we should figure out how to a) emit platform error information in telemetry and b) switch our alerts to using it.

bloodearnest:
Adjust the way we track agent/controller exceptions in telemetry

 - flip the logic from tracking transient to tracking fatal, assuming transient is the default.
 - always set the attribute, to either true or false
 - consolidate tracing functions in controller, which fixes a telemetry bug

Fixes #980

### DIFF
--- a/jobrunner/agent/main.py
+++ b/jobrunner/agent/main.py
@@ -10,7 +10,6 @@ from jobrunner.config import agent as config
 from jobrunner.config import common as common_config
 from jobrunner.executors import get_executor_api
 from jobrunner.job_executor import ExecutorAPI, ExecutorState, JobDefinition, JobStatus
-from jobrunner.lib.database import is_database_locked_error
 from jobrunner.lib.docker import docker, get_network_config_args
 from jobrunner.lib.log_utils import configure_logging, set_log_context
 from jobrunner.schema import TaskType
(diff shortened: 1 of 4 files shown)


rebkwok (approved):
If the agent encounters a fatal error it'll send the exception info up to and then the controller will receive the results with the error and raise it again, and also send the fatal_job_error info up - is that right? I assume that is what we want

bloodearnest:
> If the agent encounters a fatal error it'll send the exception info up to and then the controller will receive the results with the error and raise it again, and also send the fatal_job_error info up - is that right? I assume that is what we want

That's a good point, I hadn't considered the double reporting.

BUt I think that is probably what we want, as we want to see the error in both sets of telemetry, I think.

It means that just by looking at controller telemetry we can see the agent error.

Issue #901, opened 2025-04-24
Calculate VERSION for tracing


rebkwok:
VERSION used to be identified using git tags, however, that hasn't worked since we changed the deployment to use docker,. It's used to set the `service.version` attribute on the otel provider (in tracing.py).

It was removed in [6bdce7415](https://github.com/opensafely-core/job-runner/pull/896/commits/6bdce7415b0680cd8a61f3ef75b83f30dbf50f88) and set to just `VERSION = os.environ.get("VERSION", "")` in [20fb0fcd](https://github.com/opensafely-core/job-runner/pull/896/commits/20fb0fcda4c9f0776f640097d7596be10d0b17d9).

We should be able to identify the version from the docker image sha somehow and set it to that.

madwort:
this is now working on TPP https://ui.honeycomb.io/bennett-institute-for-applied-data-science/environments/production/datasets/jobrunner/board-query/BvJk5LN1Lrc/result/qDwuP6GrCqQ?vs=hideCompare&cstype_0=stacked

PR #985, opened 2025-05-14
Make the git rev available to the running app

madwort:
Although the docker image is tagged with the git revision, this is only accessible to the docker host, not to the app inside the running container. 

fixes https://github.com/opensafely-core/job-runner/issues/901


jobrunner/config/common.py
```diff
@@ -10,7 +10,18 @@
 BACKENDS = os.environ.get("BACKENDS", "test,tpp,emis").strip().split(",")
 
 # Used for tracing in both agent and controller
-VERSION = os.environ.get("VERSION", "")
+# This refers to a file created in the docker image by Dockerfile
+JOBRUNNER_VERSION_FILE_PATH = os.environ.get(
+    "JOBRUNNER_VERSION_FILE_PATH", "/app/metadata/version.txt"
+)
```
evansd, line 16: It would be slightly neater to do this as:
```py
JOBRUNNER_VERSION_FILE_PATH = Path(
    os.environ.get(
        "JOBRUNNER_VERSION_FILE_PATH", "/app/metadata/version.txt"
    )
)
```
Then you can call `.exists()` on it instead of using `os.path.exists` and you don't need to wrap it again below.

evansd (approved):

jobrunner/config/common.py
madwort: nice, just added

evansd (approved):

PR #990, opened 2025-05-15
Use job-server tokens for task HTTP API auth

### LINKED ISSUE #964: Tasks HTTP API - authentication

rebkwok:
Fixes #964

### DIFF
--- a/controller_app/views.py
+++ b/controller_app/views.py
@@ -1,15 +1,43 @@
 import json
 import logging
+from functools import wraps
 
 from django.http import JsonResponse
 from django.views.decorators.csrf import csrf_exempt
 from django.views.decorators.http import require_POST
 
+from jobrunner.config import common as common_config
+from jobrunner.config import controller as config
 from jobrunner.controller.task_api import get_active_tasks, handle_task_update
 from jobrunner.schema import AgentTask
 
 
 log = logging.getLogger(__name__)
+
+
+def require_backend_authentication(view_fn):
+    """Ensure a valid authentication token was received"""
+
+    @wraps(view_fn)
+    def wrapped_view(request, backend):
+        if backend not in common_config.BACKENDS:
+            return JsonResponse(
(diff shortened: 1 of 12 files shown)


controller_app/views.py
```diff
@@ -1,39 +1,67 @@
...
+                {"error": "Not found", "details": f"Backend '{backend}' not found"},
+                status=404,
+            )
+        token = request.headers.get("Authorization")
+        if not token:
+            error = "No token provided"
+        elif token != config.JOB_SERVER_TOKENS[backend]:
+            error = f"Invalid token for backend '{backend}'"
+
+        if error:
```
madwort, line 35: nit: move error closer to where it's used

```suggestion
        if backend not in common_config.BACKENDS:
            return JsonResponse(
                {"error": "Not found", "details": f"Backend '{backend}' not found"},
                status=404,
            )
        token = request.headers.get("Authorization")

        error = None
        if not token:
            error = "No token provided"
        elif token != config.JOB_SERVER_TOKENS[backend]:
            error = f"Invalid token for backend '{backend}'"

        if error:
```

rebkwok: will update, but won't accept the suggestion otherwise you won't be able to re-review

rebkwok: Done - although now I remember that this is going into `dev` so accepting the suggestion would have been fine :shrug:

madwort (approved):

PR #994, opened 2025-05-16
Create separate agent/controller service entrypoints

evansd:
Both the agent and the controller still have ancillary processes they need to run. We could in theory deploy these as further background processes using `run-one-constantly` but we'd need to re-implement some of the logging and error handling machinery and so for now it seems to better to stick with the multi-threaded service pattern.


jobrunner/agent/service.py
```diff
@@ -0,0 +1,43 @@
...
+log = logging.getLogger(__name__)
+
+
+start_thread = ThreadWrapper(log)
+
+
+def main():
+    """Run the sync and agent loops in threads, then run the controller loop
+
+    Also runs stats thread and maybe maintenance thread.
```
rebkwok, line 25: We don't have a maintenance thread anymore, do we?

jobrunner/agent/service.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+log = logging.getLogger(__name__)
+
+
+start_thread = ThreadWrapper(log)
+
+
+def main():
+    """Run the sync and agent loops in threads, then run the controller loop
```
rebkwok, line 23: This is just running the agent and metrics loops now

jobrunner/controller/service.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+log = logging.getLogger(__name__)
+
+
+start_thread = ThreadWrapper(log)
+
+
+def main():
+    """Run the sync and agent loops in threads, then run the controller loop
```
rebkwok, line 25: Run the sync and tick loops in threads, then controller loop?

rebkwok (approved):
Just some comments on docstrings

PR #1001, opened 2025-05-20
Fix trace attribute types

### LINKED ISSUE #998: some telemetry attributes are lists values
Some trace span attributes are being stored or read as list values, rather than strings:

Image

rebkwok:
Fixes #998


bloodearnest (approved):
Hah, I went looking for the cause of ths, expecting it was likely a trailing comma, and I *still didn't see it*!

Thanks for fixing

PR #1002, opened 2025-05-20
Add telemetry for the django app

### LINKED ISSUE #989: Add telemetry for django app

rebkwok:
Adds the auto-instrumentation and setup in gunicorn.conf.py and manage.py in the same way as Airlock.
Just traces backend and task id from the views for now.

This is going into main, so it should get applied to the dokku4 app, but won't actually trace anything there yet because it doesn't have `OTEL_EXPORTER_OTLP_HEADERS` set.

Fixes #989

### DIFF
--- a/controller_app/views.py
+++ b/controller_app/views.py
@@ -3,12 +3,18 @@ import logging
 from django.http import HttpResponse, JsonResponse
 from django.views.decorators.csrf import csrf_exempt
 from django.views.decorators.http import require_POST
+from opentelemetry import trace
 
 from jobrunner.controller.task_api import get_active_tasks, handle_task_update
 from jobrunner.schema import AgentTask
 
 
 log = logging.getLogger(__name__)
+
+
+def trace_attributes(**attrs):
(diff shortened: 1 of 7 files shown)


controller_app/views.py
```diff
@@ -11,13 +12,19 @@
 log = logging.getLogger(__name__)
 
 
+def trace_attributes(**attrs):
+    span = trace.get_current_span()
+    span.set_attributes(attrs)
```
bloodearnest, line 17: Worth using tracing.set_span_attributes to handle weird value types passed by accident?

evansd (dismissed):
Looks good. I'm assuming most of the setup code is borrow from one of our existing Django apps?

evansd:
:facepalm: Oh sorry, you already said it came from Airlock.

controller_app/views.py
rebkwok: yes - I thought about it and then didn't because I mistakenly remembered it doing job-related things

evansd (approved):

PR #1006, opened 2025-05-20
Store fine-grained agent stage in the controller

### LINKED ISSUE #986: Telemetry: store fine grained agent state in controller
As discussed in RAP team meeting, thi:
 - is simple to do
 - gives us parity with previous metrics
 - is actually useful when looking at job traces.

rebkwok:
This was slightly more complex than I initially thought, because I didn't want to set jobs to `PREPARING` when the task is created, because we want `status_code` to reflect the agent stage, and not change to to `PREPARING` until the agent says it is. But `set_code()` still needs to receive a new code when the task is created, so I've added a new `INITIATED` state.

Fixes #986

### DIFF
--- a/jobrunner/controller/main.py
+++ b/jobrunner/controller/main.py
@@ -249,9 +249,21 @@ def handle_job(job, mode=None, paused=None):
                 save_results(job, results)
                 # TODO: Delete obsolete files
         else:
-            # Record the fact that we've visited the job (and not forgotten about it)
-            # even though nothing has changed
-            refresh_job_timestamps(job)
+            # A task exists for this job already and it hasn't completed yet
+            # The current task stage may be None if the agent hasn't sent back any
+            # update yet, otherwise it should be in one of the running status codes which
+            # mirror ExecutorState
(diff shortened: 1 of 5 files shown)


jobrunner/controller/main.py
```diff
@@ -248,6 +248,22 @@ def handle_job(job, mode=None, paused=None):
                 results = JobTaskResults.from_dict(task.agent_results)
                 save_results(job, results)
                 # TODO: Delete obsolete files
+        elif task.agent_stage is not None and task.agent_stage != job.status_code:
```
evansd, line 251: Do we need an extra branch here? What would happen if we ran `set_code(job, StatusCode.from_value ...)` for every not yet completed job and got rid of `refresh_job_timestamps()` altogether?

evansd (dismissed):

jobrunner/controller/main.py
rebkwok: good point, will update

evansd (approved):

PR #1010, opened 2025-05-21
Send state change timestamps from agent to controller

### LINKED ISSUE #1009: re-add state timings update_controller
We now track the fine grained executor states in the controller for metrics purposes

However, we use the timestamps the controller receives the change, not the timestamp the change happened, which means that we're probably up to 30s out. preparing/finalizing are probably most impacted by this inaccuracy, as they are often <30s.

We originally included the timestamps of the change in the data sent up to the agent, but we removed it. If we re-added it, the controller can use the provided timestamp as the time of the state change, which will make our metrics more accurate.

rebkwok:
Fixes #1009 

The state change timestamp is computed and stored already on JobStatus, so in `update_job_task` we now send it to `update_controller`.  (It can also be None, for non-job-related tasks - i.e. DBSTATUS)

For running jobs, the controller uses the task's agent_timestamp_ns (if available) instead of the current time to set the job's status_code_updated_at timestamp and for tracing.

### DIFF
--- a/jobrunner/controller/main.py
+++ b/jobrunner/controller/main.py
@@ -246,7 +246,7 @@ def handle_job(job, mode=None, paused=None):
                 raise PlatformError(job_error)
             else:
                 results = JobTaskResults.from_dict(task.agent_results)
-                save_results(job, results)
+                save_results(job, results, task.agent_timestamp_ns)
                 # TODO: Delete obsolete files
         else:
             # A task exists for this job already and it hasn't completed yet
@@ -263,10 +263,11 @@ def handle_job(job, mode=None, paused=None):
                 job,
                 StatusCode.from_value(task.agent_stage, default=job.status_code),
                 job.status_message,
+                task.agent_timestamp_ns,
             )
 
 
-def save_results(job, results):
+def save_results(job, results, timestamp_ns):
     """Extract the results of the execution and update the job accordingly."""
     message = None
     error = False
@@ -296,7 +297,9 @@ def save_results(job, results):
         if results.has_level4_excluded_files:
             message += ", but some file marked as moderately_sensitive were excluded. See job log for details."
 
-    set_code(job, code, message, error=error, results=results)
+    set_code(
+        job, code, message, error=error, results=results, timestamp_ns=timestamp_ns
+    )
 
 
 def job_to_job_definition(job):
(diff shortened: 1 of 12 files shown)


rebkwok:
Reminder to self: this will require running `migrate` once deployed

evansd (approved):
This looks like it was pretty fiddly to do, but was very easy to follow thanks to the clear commit sequence.

jobrunner/controller/main.py
```diff
@@ -414,7 +414,9 @@ def set_code(
         ]:
             # we've started running
             job.state = State.RUNNING
-            job.started_at = timestamp_s
+            # Only update started_at if it hasn't already been set
+            if not job.started_at:
```
bloodearnest, line 418: IIRC we needed to do this reset every time we started executing, because we used to use job.started_at as the timestamp to calculate aggregate metrics.  We had some issues where job retries wouldn't reset it (e.g. db maintenance mode), which would throw all the maths off, as it would be based on the 1st jobs start time, not the retried start time.

However, we no longer use job.started_at for the metrics calculations, because we can't, we now use the container's created timestamp, which is a better solution anyway.

But I mentioned because the above change means we'll always have the started_at time of the job be the first time it started. Any retries will *not* reset it.

That maybe what we want, depends on how we (and job-server) use started_at. But its one of those things that is a bit messy, so mentioning it in case there are potential things to think about.

rebkwok: We reset started_at to None if the status code has changed to WAITING_ON_REBOOT or WAITING_DB_MAINTENANCE, so the started_at will get reset when the task is picked  up again.

rebkwok: Are there other ways to retry?  Other than manually setting the job in the database back to a non-running status. It doesn't seem accurate to change the started_at time when a job initiates, prepares and then starts executing.
If retrying always a job always means we have a new task, we could just set started_at when the status changes to INITIATED, because that's when the task is created.

bloodearnest: > We reset started_at to None if the status code has changed to WAITING_ON_REBOOT or WAITING_DB_MAINTENANCE, so the started_at will get reset when the task is picked up again.

ah, right, yes, then the always setting it was a red herring then.

> Are there other ways to retry? Other than manually setting the job in the database back to a non-running status.

retry_job was another way, I'm don't think it survived the split though?

> It doesn't seem accurate to change the started_at time when a job initiates, prepares and then starts executing.
If retrying always a job always means we have a new task, we could just set started_at when the status changes to INITIATED, because that's when the task is created.

Yeah, its a bit unclear to me what start at is supposed to track exactly. It's sent to job-server though, and stored in the db, which is used in the dashboard metrics, so its a bit messy. I think this is one of the questions we should probably punt on, and fix in step 3.

rebkwok: Yup, retry_job is gone

bloodearnest (approved):

PR #1012, opened 2025-05-21
Don't re-raise Exceptions when jobs return errors

### LINKED ISSUE #999: Stop re-raising agent errors in controller
Currently, we have one location for error handling in controller loop.

It will handle all errors raised by the main controller logic.

However, currently, we re-raise agent errors as Platform errors, so the same exception get handled twice.

Instead of handling agent errors in the main controller error handler, we should handle them as part of normal logic (e.g. decide if its fatal, emit elemetry, etc), but not re-raise them. Then the top level error handling is only dealing with exceptions rising from the controller code itself, not the agent.

Some discussion: 
https://bennettoxford.slack.com/archives/C069YDR4NCA/p1747383544143289

madwort:
* fixes #999

### DIFF
--- a/jobrunner/controller/main.py
+++ b/jobrunner/controller/main.py
@@ -118,8 +118,8 @@ def handle_single_job(job):
         try:
             handle_job(job, mode, paused)
         except Exception as exc:
-            if is_fatal_job_error(exc):
-                span.set_attribute("fatal_job_error", True)
+            span.set_attribute("fatal_job_error", is_fatal_controller_error(exc))
+            if is_fatal_controller_error(exc):
                 mark_job_as_failed(
                     job,
                     StatusCode.INTERNAL_ERROR,
@@ -129,8 +129,6 @@ def handle_single_job(job):
                     "investigating.",
                     error=exc,
                 )
-            else:
-                span.set_attribute("fatal_job_error", False)
             # Do not clean up, as we may want to debug
             #
             # Raising will kill the main loop, by design. The service manager
@@ -143,7 +141,8 @@ def handle_single_job(job):
             span.set_attribute("final_code", job.status_code.name)
 
 
-def is_fatal_job_error(exc: Exception) -> bool:
+def is_fatal_controller_error(exc: Exception) -> bool:
+    """Returns whether an Exception thrown by the controller should be fatal to the job."""
     # To faciliate the migration to the split agent/controller world we don't currently
     # consider _any_ errors as hard failures. But we will do so later and we want to
     # ensure that these code paths are adequately tested so we provide a simple
--- a/jobrunner/controller/main.py
+++ b/jobrunner/controller/main.py
@@ -150,6 +150,13 @@ def is_fatal_controller_error(exc: Exception) -> bool:
     return "test_hard_failure" in str(exc)
 
 
+def is_fatal_job_error(exc: Exception) -> bool:
+    """Returns whether an Exception thrown in the agent & returned to the controller should be fatal to the job."""
+    # An example might be: if there is version skew between the agent & the controller
+    # and the agent reports an exception due to an API change.
+    return "test_job_failure" in str(exc)
+
+
 def handle_job(job, mode=None, paused=None):
     """Handle an active job.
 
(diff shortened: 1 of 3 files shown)


jobrunner/controller/main.py
```diff
@@ -242,8 +235,12 @@ def handle_job(job, mode=None, paused=None):
         assert task is not None
         if task.agent_complete:
             if job_error := task.agent_results["error"]:
-                # Handled elsewhere
-                raise PlatformError(job_error)
+                mark_job_as_failed(
```
bloodearnest, line 238: We should only error the job if is_fatal_error is true?

I think we need to move that check here?

Otherwise, *any* agent error will fail the job, which is what we are trying to avoid with is_fatal_error?

rebkwok: The agent doesn't set error=True unless it encounters an is_fatal_error, does it?  It only calls `mark_task_as_error` for fatal errors, and afaict that's the only thing that calls finalize with error=False.

bloodearnest: So, the agent's is_fatal_task_error is to decide whether this *task* can be retried on the next loop, or needs completing returning to the controller.

The controllers `is_fatal_job_error` decides whether this *job* can be retried on the next loop (i.e. a new task created for it), or whether the job needs marking as failed and returning to job-server.

Currently, we are relying on the raising of PlatformError, and for this to be handled in handle_single_job, and set appropriately.

In this change, we are no longer raising PlatformError in this case.  But we still want to use is_fatal_job_error to decided whether to fail the job, or to retry with a new task on the next loop.

evansd (commented):
I think the original ticket is more involved that it might seem at first glance; it's not a quick fix.

The crucial line is:
> decide if its fatal

As in, we need logic which takes task errors and decides whether the job needs failing or the task needs rescheduling. And I don't think we're yet clear on what the appropriate conditions are here.

I guess our current approach is to always try to keep going unless we know for certain that the job is fatally borked, so maybe an initial implementation is just:
```py
def is_fatal_error(task_error):
    return False
```

We still need the logic to schedule a new task though.

jobrunner/controller/main.py
madwort: > The controllers is_fatal_job_error decides whether this job can be retried on the next loop (i.e. a new task created for it), or whether the job needs marking as failed and returned to job-server.

yes, from our reading of the code & tests it does this so that it can restart the controller itself & retry in the event of an internal error in the controller (e.g. controller database "database locked"). Which still happens, because this PR retains the `is_fatal_job_error` test in `manage_single_job`. I think the agent is also resilient in the same way.

However, if we have a resilient agent which retries in the event of internal errors (& potentially other non-fatal errors in researcher code / the agent environment) & a resilient controller, then it seems to me that there's no situation where the controller would want to retry a job by creating a new task with exactly the same specification for a task where the agent returns an error - actually it's kind of a layering violation? 

I realise that we can just put in hooks at every possible layer - and maybe we should - it just seemed unnecessary. It's slightly difficult to reason about this because we're not working with concrete scenarios, we're working with synthetic theoreticals in tests.

bloodearnest: So, one case where restarting a job with new task by the controller would be useful when we are deploying something that changes the JobDefintion format, as we did recently.  In that case, restrying with new task with the new format is a nice property.

Currently we retry all jobs and all tasks. That is a deliberate choice we made to make these big deployments more robust in the case of transient errors in deployment, of which we've had a few, and avoid cascading failures of jobs that we'd have to manually resubmit.

We have discussed in the team meeting that we need to revist this once we're more stable and deployed in dokku4, and actually decide and implement what errors should be counted as fatal or not.

Once we have that, the control flow around what we're discussing between the two becomes more useful.

The agent makes local decisions about tasks, the controller about jobs. I see it as better layering, as opposed to a violation?

madwort: > So, one case where restarting a job with new task by the controller would be useful when we are deploying something that changes the JobDefintion format, as we did recently. In that case, restrying with new task with the new format is a nice property.

Ok, so the scenario is - a new controller sends a new format task to an old agent, the task fails because it's bad format. The new controller retries with a new task & this time gets the new agent & it works. Makes sense.

> The agent makes local decisions about tasks, the controller about jobs. I see it as better layering, as opposed to a violation?

The way I was looking at this was - the agent has a task, the agent does its level best to run the task but it can't & returns an error, but the controller respawns an identical task & sends it back to the agent. What does the controller know about the task that the agent doesn't? If we can have version skew between the controller & agent, though, that could do it.

bloodearnest: Right, that's what is_fatal_job_error needs to decide, in future.

Certain classes of task failure may be retryable, others may not be.  It will depend somewhat on what is_fatal_task_error ends up doing.

Examples of agent errors, and possibly ways of handling them

 - image doesn't exist on agent. This is probably fatal to the task, and probably also the job
 - we run out of disk space. This is probably fatal to the task, but not the job, as we we'd expect something to fix manually
 - something unknown: maybe we retry once or twice, then its fatal to the job.
 - memory limit hit
    - if it hit the per job limit, I think we fail the task and the job
    - if it hit a global limit, we maybe fail the task, but retry the job.
 
There's bunch of combinations that could occur, and AFAICS we need control points for that logic in both loops.

bloodearnest: We're not doing *any* of this right now, which makes it harder to reason about, but I expect we will do some of it soonish.

The is_fatal_*_error logic is there as a plug point when we do want to.

madwort: > Examples of agent errors, and possibly ways of handling them
> 
>     * we run out of disk space. This is probably fatal to the task, but not the job, as we we'd expect something to fix manually

but in this case I don't think the answer is to retry the job immediately by creating a new task, it's to -waves hands- pause the job awaiting a manual unpause when action has been taken - which is pretty close to failing the job & reschedule it.  

>     * memory limit hit
>       * if it hit a global limit, we maybe fail the task, but retry the job.

I think the agent should retry the task in this case? at least, the agent should know more about this situation than the controller?

> We're not doing any of this right now, which makes it harder to reason about, but I expect we will do some of it soonish

we agree!

anyway, I'm going to add the plug point.

PR #1014, opened 2025-05-22
Tweak main controller code style

madwort:
When pairing with Providence, we were finding it a bit fiddly with `handle_jobs`, `handle_single_job` & `handle_job`. This squeezes some stuff in either direction out of `handle_single_job` to reduce it to a telemetry-only wrapper (which could now be a decorator, I think). 

* Also reduce unnecessary indentation

### DIFF
--- a/jobrunner/controller/main.py
+++ b/jobrunner/controller/main.py
@@ -125,12 +125,11 @@ def handle_jobs():
 
 def handle_job_telemetry_wrapper(job):
     """A wrapper for handle_job which adds OpenTelemetry"""
-    attrs = {
-        "initial_state": job.state.name,
-        "initial_code": job.status_code.name,
-    }
-
     with tracer.start_as_current_span("LOOP_JOB") as span:
+        attrs = {
+            "initial_state": job.state.name,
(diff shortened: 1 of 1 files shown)


bloodearnest (commented):
So, I specifically unwrapped the telemetry function in #983 a few days ago, as it IME leads to confusing flow. I added the telemetry wrapper functions originally, and have regretted the design ever since.

Having error handling and error telemetry split across two exception handlers is not nice to read, and I think makes it hard to maintain, and introduce inconsistencies.  

Whilst the names handle_{job,jobs,single_job} may well be less than ideal, I think a better solution to that is renaming them, not adding an additional function into the mix?

bloodearnest:
Ah, ok, I missed that.

But it still splits the exception handling across handle_jobs and handle_jobs_telemetry_wrapper, right? That's the part that concerns me, as that's what didn't work about the previous arrangement.

But perhaps this doesn't matter once you merge in #1012, as that error handling will not have any logic in it?

madwort:
closing to bring back later

PR #1015, opened 2025-05-23
Make canceljob_db_task_factory take an optional job

rebkwok:
Makes canceljob_db_task_factory consistent with runjob_db_task_factory, which allows us to more easily make multiple tasks for the same job, and also simplifies some tests that need to make an inactive cancel task.


bloodearnest (approved):
Nice change, it very much worth aggresively refactoring and maintaining our test factories IMO

PR #1016, opened 2025-05-23
Add attributes to agent task for telemetry

### LINKED ISSUE #987: Telemetry: add opaque attribute dict to AgentTask
Currently the agent is not sent information like `project` or `user`, as it doesn't need them.

However, its useful to be able to tag telemetry that the agent generates with such metadata.

We should add an opaque dict to the AgentTask that contains a bunch of key value pairs the the controller populates with whatever we decide is needed.

The agent can then automatically add these key-pairs as attributes on any spans it generates that related to the that task, w/o caring about the contents. This is not just limited to RUNJOB or CANCELJOB, but could be used for any other tasks.

This is easy in the agent loop, where we have ready access to the appropriate AgentTask, but a bit harder in the metrics thread, as we don't currently have access to AgentTask, just running containers, since we do persist AgentTasks in a db or similar, but we still want to tag the METRICS spans we emit.

We'll probably have to do something like:
a) call get_active_tasks() there too, just to get the metadata for a given job container
b) annotate the containers with the metrics dict, so we can pull it from the container metadata we already fetch.

rebkwok:
Fixes #987 

The ticket mentions adding an attribute dict to AgentTask, which the controller populates with any key-value pairs. Since the main ones that we want to add at the moment are only available on a SavedJobRequest, I've added the attribute dict to the Task model so it will be stored in the db.  Otherwise it will need to query SavedJobRequest for every task, every time the agent and metrics loops call get_active_tasks().

### DIFF
--- a/jobrunner/controller/main.py
+++ b/jobrunner/controller/main.py
@@ -32,7 +32,11 @@ from jobrunner.lib.database import (
 )
 from jobrunner.lib.log_utils import configure_logging, set_log_context
 from jobrunner.models import Job, State, StatusCode, Task, TaskType
-from jobrunner.queries import calculate_workspace_state, get_flag_value
+from jobrunner.queries import (
+    calculate_workspace_state,
+    get_flag_value,
+    get_saved_job_request,
+)
 from jobrunner.schema import JobTaskResults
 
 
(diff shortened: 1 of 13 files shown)


jobrunner/controller/main.py
```diff
@@ -542,6 +546,16 @@ def update_job(job):
     update(job, exclude_fields=["cancelled"])
 
 
+def get_attributes_for_job_task(job):
+    job_request = get_saved_job_request(job)
+
+    return {
+        "user": job_request.get("created_by", "unknown"),
```
Providence-o, line 553: Do we intend to define `created_by`, `project` and `orgs` as part of the JobRequest class or are they defined somewhere else?

Providence-o (approved):

PR #1017, opened 2025-05-23
Make tests error if they need the db fixture

rebkwok:
Tests that need a database use the `db` fixture, which patches the DATABASE_FILE config to a temporary filepath. If you write a test and forget to add the db fixture, but you have a local db, the test will pass, but it'll use your (real) local db. This means that it (a) messes with your local environment and (b) makes the test fail in CI only, where there's no local db for it to use.

This adds an autouse fixture that always sets the DATABASE_FILE config variable to a temporary path. If a test attempts to use the database without the db fixture, it will fail because the db fixture runs the actual database setup, which we only want to do if it's required.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -53,7 +53,6 @@ def set_tmp_workdir_config(monkeypatch, tmp_path):
     monkeypatch.setattr(
         "jobrunner.config.controller.DATABASE_FILE", tmp_path / "db.sqlite"
     )
-
     config_vars = {
         "common": ["GIT_REPO_DIR"],
         "agent": [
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -196,14 +195,22 @@ def test_repo(tmp_work_dir):
     )
 
 
+@pytest.fixture(autouse=True)
+def db_config(monkeypatch, request):
+    """
+    Always patch DATABASE_FILE to a temporary path to avoid mangling
+    a local db file if a test accidentally omits the db fixture.
+    """
+    database_file = f"file:db-{request.node.name}?mode=memory&cache=shared"
```
evansd, line 204: I wonder if it would be clearer to set `DATABASE_FILE` to `None` here and then only explicitly set an actual path inside the `db()` fixture. Otherwise there's a slightly weird implicit dependency between the two fixtures with one setting the path and the other using it.

evansd (dismissed):
This is a very good idea, thanks.

tests/conftest.py
rebkwok: That's actually what I did before, and then I changed it to this instead - partly because I realised that the existing metrics db is an autouse fixture. I didn't want to make the db autouse because it would do the migrations etc on every test.

If you set it to None, it fails at the connection, and I'd added an assertion with a user-friendly error message (since the DATBASE_FILE config should only ever end up being None in a test):
```
def filename_or_get_default(filename=None):
    if filename is None:
        filename = config.DATABASE_FILE
        assert filename is not None, (
            "DATABASE_FILE config must not be None; did you omit the `db` fixture in a test?"
        )
    return filename
```

I quite like the helpful message, so happy to revert to this approach if you also prefer it.

evansd: Oh yeah, I think having that assert is really helpful and will definitely save someone time in future. If you don't mind doing the work revert than I think that would be better. I don't think it's hugely critical though.

rebkwok: Nah, the most time-consuming part will be rewriting the commit message

bloodearnest:
This is a great idea, it has bitten me before too.

PR #1025, opened 2025-05-28
Update DEVELOPERS.md

### LINKED ISSUE #1020: Update DEVELOPERS.md following recent changes
DEVELOPERS.md should describe what developers need to know to work on a component.  It is now quite out of date.  We should update it, and/or remove parts that are no longer relevant.

Consider sharing notes with whoever's working on #1019.

rebkwok:
Fixes #1020


docs/execution_state_diagram.md.md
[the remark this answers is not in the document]
rebkwok: yup, i noticed that but then I was trying to mush retries into it so didn't update, will do now

docs/controller_state_diagram.md
```diff
@@ -0,0 +1,123 @@
...
+
+    }
+
+    state finalized_fork_state <<choice>>
+    AgentStages --> finalized_fork_state
+    finalized_fork_state --> succeeded
+    finalized_fork_state --> failed: fatal error
+    finalized_fork_state --> retry: non-fatal error
+
+    retry --> created
```
madwort, line 63: wondering if it's worth clarifying how this works?
```suggestion
    retry --> created: RUNJOB task deactivated, controller restarted, new RUNJOB task created
```

rebkwok: Is this how it works?  

For a controller error, if it's [fatal we mark the job as failed](https://github.com/opensafely-core/job-runner/blob/main/jobrunner/controller/main.py#L122), which doesn't deactivate the task but means the job won't be picked up again. If it's non-fatal, we don't do anything with the job or task, which I think should mean that the _same_ task gets picked up by the agent again.

For a [job error](https://github.com/opensafely-core/job-runner/blob/main/jobrunner/controller/main.py#L246), the task is already inactive because the error was a final agent state. If it's non-fatal we put it into a pending (retry) state so a new task gets created. But we don't restart the controller.

madwort: yep you're right. we should split that into two states as you suggest, though.

madwort: > I've made it go back to the initiated step

yep agree, I think this is best.

madwort (approved):
I think the agent/controller diagrams really add something!

Issue #967, opened 2025-05-08
Tasks HTTP API - deployment


rebkwok:
#963  implements the basic HTTP API but doesn't handle deployment.

Initially we plan to deploy this inside the backends, running alongside the controller/agent, until we are able to split the controller out.  This will probably involve running the django app in a separate docker container that the agent can call.

We'll need:
- to be able to build and run the controller app using the existing Dockerfile
- a docker-compose service? for running the controller app
- a way for the job-runner container to be able to call the app container (via docker compose networking?)
- a local docker test that runs the two containers and tests with manual add-job
- updates to backend-server's docker-compose to allow the two containers to run together in prod too

evansd:
I think we need three containers here: the agent, the controller update loop, and the controller web service. It may be possible later to do away with the update loop and make the controller entirely event driven, but that's not something we plan to do immediately.

It would probably also be possible to hack the controller loop and web service to run from the same entrypoint, but that seems a bit pointless given that we have to have multiple containers in any case.

bloodearnest:
For similar reasons as when we moved to running the backgroung uploader process in airlock, I am keen to keep the number of controller containers to a single multiprocess container if possible.

But I think same run-one-constantly approach should work here too?

madwort:
> For similar reasons as when we moved to running the backgroung uploader process in airlock, I am keen to keep the number of controller containers to a single multiprocess container if possible.

slightly curious - what were those reasons?

bloodearnest:
I think that a single "atomic" deployment step, and single service to stop/start, view logs etc, is to be strongly preferred to having to deploy/stop/start two containers.

We could have tooling that treats them as a single unit wtr the above operation, but that adds complexity, and the seam between them still exists, and things can go wrong.

inglesp:
The deployment has happened.

PR #1026, opened 2025-05-28
Changes to `Procfile` required for deployment

evansd:
Note that this means switching to running the control loop in a separate container, rather than as a background process in the web container. I believe that the [factors][1] which led us to do this were specific to the docker-compose deployment, and that running as a separate containers is a better fit for Dokku.

[1]: https://github.com/opensafely-core/job-runner/issues/967#issuecomment-2883537349

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,3 @@
 web: /opt/venv/bin/gunicorn --config gunicorn.conf.py controller_app.wsgi
+service: /opt/venv/bin/python -m jobrunner.controller.service
 release: /opt/venv/bin/python -m jobrunner.cli.controller.migrate


rebkwok (approved):
Do we need an app.json and a custom healthcheck to ensure that the migrations are complete before the web container starts up fully?

rebkwok:
https://dokku.com/docs/deployment/zero-downtime-deploys/

Not quite the same checks, but bennettbot runs 3 containers in the same dokku app and has [custom healthchecks](https://github.com/bennettoxford/bennettbot/blob/main/app.json) for 2 of them (the bot healthcheck runs a script that adds a file once it's started up, and the Procfile `release` [removes it](https://github.com/bennettoxford/bennettbot/blob/main/Procfile))

PR #1032, opened 2025-05-29
Run cleanup() for cancelled jobs in the PREPARED stage

rebkwok:
A job in the PREPARED state has no container yet, but it does have a volume (which is how get_status() identifies it as PREPARED. https://github.com/opensafely-core/job-runner/blob/4df192219032451622ce99818402a20abc8e21ca/jobrunner/executors/local.py#L305

We therefore also need to run cleanup after cancelling  a job that is PREPARED; cleanup will try to delete the container, but it ignores errors if there's no container to remove, and it will delete the volume.

### DIFF
--- a/jobrunner/agent/main.py
+++ b/jobrunner/agent/main.py
@@ -159,14 +159,11 @@ def handle_cancel_job_task(task, api):
                     f"unexpected state of job {job.id}: {initial_job_status.state}"
                 )
 
-        # Clean up based on the starting job state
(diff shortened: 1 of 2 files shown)


rebkwok:
> It's idempotent so calling it even if the container and volume don't exist should be safe.

When I looked at the code I thought that it handled non-existent containers but not volumes, but actually it does handle volumes too, so have updated

evansd (approved):

PR #1036, opened 2025-05-30
Add some deployment notes

evansd:


DEPLOY.md
```diff
@@ -0,0 +1,101 @@
...
+# these loops running simultaneously
+dokku checks:disable rap-controller service
+
+# Create an appropriately owned storage directory and mount it
+sudo mkdir -p /var/lib/dokku/data/storage/rap-controller
+sudo chown rap-controller:rap-controller /var/lib/dokku/data/storage/rap-controller
+dokku storage:mount rap-controller /var/lib/dokku/data/storage/rap-controller:/storage
+
+# Run app as correct user, both the deployed services and one-off commands
+dokku docker-options:add rap-controller deploy,run "--user=$(id -u rap-controller):$(id -g rap-controller)"
```
madwort, line 50: nit: this is fine & good. /ideally/ we would also add this user to the docker image, but afaik for us the only current downside is aesthetic:

```
madwort@dokku4:~$ dokku enter rap-controller web
groups: cannot find name for group ID 10002
I have no name!@e4c0461e1d26:/app$ whoami
whoami: cannot find name for user ID 10002
```
(I think we have tidied this up in some other containers but not all)

DEPLOY.md
```diff
@@ -0,0 +1,101 @@
...
+  # These are the default values in the codebase, but it seems clearer
+  # to set them explicitly here
+  JOB_SERVER_ENDPOINT=https://jobs.opensafely.org/api/v2/
+  OTEL_SERVICE_NAME=jobrunner
+)
+```
+
+Finally we can apply the configuration and deploy the app:
+```bash
+dokku config:set rap-controller "${config[@]}" "${secrets[@]}"
```
madwort, line 99: hah! nifty!

DEPLOY.md
```diff
@@ -0,0 +1,101 @@
...
+tooling. This will ensure that Dokku is installed and configured with the
+necessary plugins and that a `rap-controller` user has been created.
+
+First we create the app and do some basic configuration:
+```bash
+dokku apps:create rap-controller
+
+# Set up the domain
+dokku domains:set rap-controller controller.opensafely.org
+dokku letsencrypt:enable rap-controller
```
madwort, line 38: does this need to happen after you've deployed a container for the first time? doesn't really matter as you've said this is illustrative & someone running this should be able to handle that if it came up.

madwort (dismissed):

DEPLOY.md
evansd: Ah yeah, was discussing this earlier with Becky. IMNSHO adding users to the Docker image is the wrong way of dealing with this. Made an issue here to look at later:
https://github.com/opensafely-core/base-docker/issues/39

DEPLOY.md
evansd: It could well be. These instructions obviously aren't actually what I did because I didn't do the initial setup. I can well imagine this doesn't work until you've got an app live, but I can also imagine it _does_ work given that the LetsEncrypt plugin has to take over at least some URLs in order to do the auth dance.

Would be good to confirm one way or the other, but not urgent enough for me to do that now I think.

DEPLOY.md
```diff
@@ -0,0 +1,104 @@
...
+
+# Disable zero-dowtime deploys for service, we don't ever want two of
```
rebkwok, line 40: typo: s/dowtime/downtime

DEPLOY.md
```diff
@@ -0,0 +1,104 @@
...
+# these loops running simultaneously
+dokku checks:disable rap-controller service
+
+# Create an appropriately owned storage directory and mount it
+sudo mkdir -p /var/lib/dokku/data/storage/rap-controller
+sudo chown rap-controller:rap-controller /var/lib/dokku/data/storage/rap-controller
+dokku storage:mount rap-controller /var/lib/dokku/data/storage/rap-controller:/storage
```
rebkwok, line 47: The ansible script in sysadmin will do this [here](https://github.com/bennettoxford/sysadmin/blob/main/infra/dokku_app_users.yaml#L14)

In fact, I'm not sure that the ansible script will work until the dokku app is created, because it runs dokku commands that expect the app

DEPLOY.md
madwort: ok let's discuss in that ticket

DEPLOY.md
madwort: ottmh if you haven't deployed an app at all, then port 80 isn't mapped to anything yet, so dokku attempts to redirect the url but letsencrypt finds a closed port & fails. doesn't need doing now.

DEPLOY.md
evansd: Ah, so I had the same worry initially then I saw it ran with `check_mode: false` and assumed that meant it would just ignore failures. But now I've double-checked and that isn't what it does at all!

So yeah, this is a problem. But I don't understand how this is supposed to work because you can't create the app until the server is set up with Dokku, which is what Ansible is supposed to do. Maybe this has never been run from scratch against a fresh server?

I think updating the command so that it just outputs nothing if the app isn't configured yet would be the right thing to do here.

Thanks for catching this.

rebkwok: I don't think it has ever been run from scratch. I think we switched all apps to run as non root a while ago and it was maybe set up then, but all the apps already existed

rebkwok: There is an ignore-failure ansible option (can't remember exactly what it's called), maybe we just need that?

PR #1046, opened 2025-06-04
Tweak comments

madwort:


jobrunner/agent/tracing.py
```diff
@@ -62,7 +62,7 @@ def set_job_span_metadata(span, job: JobDefinition, **attrs):
 
 
 def trace_job_attributes(job: JobDefinition):
-    """These attributes are added to every span in order to slice and dice by
+    """These attributes are added to every job span in order to slice and dice by
```
rebkwok, line 65: I think it's still the task span, but we add job attributes to it (from a task's job definition); this is called from inside handle_run_job_task and handle_cancel_job_task, which are wrapped by handle_single_task, which sets the LOOP_TASK span

madwort: you're quite right, changing

rebkwok (approved):

PR #1055, opened 2025-06-05
Reduce indentation

madwort:
* This function spans multiple screens & has multiple levels of nesting, so I think this change makes it easier to read.


madwort:
good suggestion!

rebkwok (approved):

PR #1059, opened 2025-06-06
Update cohort-extractor comment to point to ehrQL

madwort:


jobrunner/config/controller.py
```diff
@@ -120,8 +120,8 @@ def parse_job_resource_weights(config_file_template):
 
 # Map known exit codes to user-friendly messages
 DATABASE_EXIT_CODES = {
-    # Custom database-related exit codes return from cohortextractor, see
-    # https://github.com/opensafely-core/cohort-extractor/blob/0a314a909817dbcc48907643e0b6eeff319337db/cohortextractor/cohortextractor.py#L787
+    # Custom database-related exit codes return from ehrQL, see e.g.
+    # https://github.com/opensafely-core/ehrql/blob/main/ehrql/backends/tpp.py#L159-L171
```
evansd, line 124: Sorry for the post-merge comment, I just happened to be looking at this.

I think that when linking to code we almost always want an immutable link to a specific commit (the `y` keyboard shortcut in the Github UI will give you one of these). Lines (and whole files) move about and then it's very awkward to work backwards and figure out what it was supposed to be pointing out. Whereas jumping forwards from a specific commit to the latest version is much easier.

PR #1061, opened 2025-06-10
Fix bug in update timestamps

rebkwok:
A bug in the previous implementation of using set_code with a timestamp meant that we actually never set the timestamp to the agent-provided timestamp for running jobs.  However, we still don't want to _always_ set the timestamp to the agent-provided timestamp, because then we'd never change the updated_at time for a job whose status code hadn't changed.

This fixes the bug with the positional/keyword args, and also updaes set_code() so that we use the agent-provided task timestamp only when a status code has changed. If it hasn't changed, we use the current time instead.

(Note that we could perhaps add a field to the Task to record the time it was updated in `task_api.handle_task_update()`, and use that for running jobs. That would be more accurate in recording the time the _agent_ last updated, but would involve database changes so I'm leaving it for now)

### DIFF
--- a/jobrunner/controller/main.py
+++ b/jobrunner/controller/main.py
@@ -297,7 +297,7 @@ def handle_running_job(job):
             job,
             StatusCode.from_value(task.agent_stage, default=job.status_code),
             job.status_message,
-            task.agent_timestamp_ns,
+            task_timestamp_ns=task.agent_timestamp_ns,
         )
 
 
@@ -332,7 +332,7 @@ def save_results(job, results, timestamp_ns):
             message += ", but some file(s) marked as moderately_sensitive were excluded. See job log for details."
 
     set_code(
-        job, code, message, error=error, results=results, timestamp_ns=timestamp_ns
+        job, code, message, error=error, results=results, task_timestamp_ns=timestamp_ns
     )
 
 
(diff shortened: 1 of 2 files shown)


jobrunner/controller/main.py
```diff
@@ -427,15 +427,22 @@ def set_code(
...
+
+    def get_timestamps(from_task):
+        if from_task and task_timestamp_ns is not None:
+            timestamp_ns = task_timestamp_ns
+            timestamp_s = int(timestamp_ns / 1e9)
+        else:
+            t = time.time()
+            timestamp_s = int(t)
+            timestamp_ns = int(t * 1e9)
+        return timestamp_ns, timestamp_s
```
evansd, line 439: Having read the entire function and thought about the behaviour we want I can see why you've gone for an inner function here, but it wasn't immediately obvious to me and I did find it a bit confusing on first reading.

I wonder if there's another way of factoring this that might make the behaviour more obvious. Would it work to have a clause at the top of the function which creates a current timestamp and then ensures that `task_timestamp_ns` is always set:
```py
current_timestamp_ns = int(time.time() * 1e9)
if task_timestamp_ns is None:
    task_timestamp_ns = current_timestamp_ns
```

And then each of the branches can have either
```py
timestamp_ns = task_timestamp_ns
timestamp_s = int(task_timestamp_ns / 1e9)
```

```py
timestamp_ns = current_timestamp_ns
timestamp_s = int(current_timestamp_ns / 1e9)
```

That feels like it _might_ be clearer to me, but not definitely. Fine to leave as is if you'd prefer.

evansd (approved):
Nice. This was a much less fiddly fix than I feared.

I've made an ignorable micro-comment about factoring `get_timestamps()` differently, but otherwise all looks good.

PR #1075, opened 2025-06-25
Prefix open telemetry attributes

bloodearnest:
Follow otel best practice, and prefix our span attributes.

This also helps filtering on attribute in Honeycomb UI, and improves clarity
and readibility of the attributes.


madwort (approved):
Really good to use namespaces. 

Probably good if [Add temporary backwards compatibility attribute names](https://github.com/opensafely-core/job-runner/pull/1075/commits/f58b1ae799e5642d7b1529b1466fc36534fea65f) is a standalone commit so that it can be easily reverted later? Either just remember not to squash when merging or add that as a separate PR?

bloodearnest:
> Really good to use namespaces.
> 
> Probably good if [Add temporary backwards compatibility attribute names](https://github.com/opensafely-core/job-runner/pull/1075/commits/f58b1ae799e5642d7b1529b1466fc36534fea65f) is a standalone commit so that it can be easily reverted later? Either just remember not to squash when merging or add that as a separate PR?

I specifically wrote it as separate commit so that it would be easy to back out :+1: 

I never squash merge, I thought our policy was not to?!

PR #1076, opened 2025-06-25
Use new tpp-database utils docker image

madwort:


evansd (dismissed):
Will also need to pull this image in the prod environment before deploying.

evansd (approved):

PR #1078, opened 2025-06-27
Remove cohortextractor

madwort:
* mostly removing dummy commands from test projects
* also remove action special case


tests/fixtures/full_project/analysis/dummy_action.py
```diff
@@ -10,7 +10,7 @@ def main(output_file):
     "action" does the bare minimum just so we can have another action in the
     project.yaml.  Once databuilder supports more column types in the dummy
     data then we can remove this in favour of more meaningful actions such as
-    those used in the cohortextractor version of the integration test.
+    those previously used in the cohortextractor version of the integration test.
```
rebkwok, line 13: I think we can remove this entire file, and the action that uses it from the `full_study/project.yaml`.  The full study now does do all the integration test things that the cohort-extract one did (the comment shows its age by the fact that it refers to databuilder, not ehrql...)

madwort: ooh, yeah, I think you're right, have removed it

rebkwok (approved):

PR #1084, opened 2025-07-04
Remove non-functioning port variable

madwort:
* also make explicit the default port 8000 (which is what we are currently deployed on)


bloodearnest (approved):

bloodearnest:
Is it worth adding EXPOSE 8000 to the docker file?

PR #1088, opened 2025-07-10
Add agent cli command for emitting ehrQL telemetry

bloodearnest:
This adds an agent CLI command, which can be executed in the deployed environment with:
```
just ehrql_telemetry DATASET_NAME JOB_SPECIFIER
```

This will read the job's log file, parse out the relevant events, and push them as backdated otel records into the specified Honeycomb dataset. 

`DATASET_NAME` can be chosen pretty much arbitrarily as long as:
 * you avoid using one of our existing service names like `rap-controller`;
 * you use a consistent dataset name when processing all the jobs you want to analyse.

`JOB_SPECIFIER` can either be a job ID or a path to a job's log directory.

For more detail on why we are using this approach see the below document:
https://docs.google.com/document/d/1EXdQNSOUOvbx42JUw952GvCwh87C4sLbYlLkMQqe4EI/

### DIFF
--- a/agent/cli/ehrql_telemetry.py
+++ b/agent/cli/ehrql_telemetry.py
@@ -54,7 +54,7 @@ def main(argv):
 
     # force our name to be used as dataset
     os.environ["OTEL_SERVICE_NAME"] = args.dataset
-    setup_default_tracing()
+    setup_default_tracing("agent")
 
     run(metadata, logs)
 
--- a/agent/cli/ehrql_telemetry.py
+++ b/agent/cli/ehrql_telemetry.py
@@ -96,8 +96,11 @@ def convert_ehrql_logs(logfile_path):  # pragma: nocover
     process = subprocess.run(
         [
             "docker",
-            "run--rm",
-            "ghcr.io/opensafely-core/ehrql:v1/app/scripts/parse_logs.py",
+            "run",
+            "--rm",
+            "--interactive",
+            "ghcr.io/opensafely-core/ehrql:v1",
+            "/app/scripts/parse_logs.py",
         ],
         input=logfile_path.read_text(),
         capture_output=True,
--- a/agent/cli/ehrql_telemetry.py
+++ b/agent/cli/ehrql_telemetry.py
@@ -13,8 +13,15 @@ from common.tracing import setup_default_tracing
 
 
 def main(argv):
-    parser = argparse.ArgumentParser(description="db_telemetry")
-    parser.add_argument("dataset", help="Honeycomb dataset to emit telemetry to")
+    parser = argparse.ArgumentParser(
+        description=(
+            "Emit an ehrql job's query data as telemetry."
+            "Reads the job's log file, and converts the query timing to otel, and sends them out."
+        )
+    )
+    parser.add_argument(
+        "dataset", help="Temporary Honeycomb dataset to emit telemetry to."
+    )
     parser.add_argument(
         "job", help="job to emit telemetry for. Either job id or path to job log dir."
     )
--- a/agent/cli/ehrql_telemetry.py
+++ b/agent/cli/ehrql_telemetry.py
@@ -1,0 +1,134 @@
+import argparse
+import json
+import os
+import subprocess
+import sys
+from datetime import datetime
+from pathlib import Path
+
+from opentelemetry import context, trace
+
+from agent.executors.local import job_metadata_path
+from common.tracing import setup_default_tracing
+
+
+def main(argv):
+    parser = argparse.ArgumentParser(description="db_telemetry")
+    parser.add_argument("dataset", help="Honeycomb dataset to emit telemetry to")
+    parser.add_argument(
+        "job", help="job to emit telemetry for. Either job id or path to job log dir."
+    )
+    args = parser.parse_args(argv)
+
+    if os.path.isdir(args.job):
+        log_dir = Path(args.job)
+    else:
+        log_dir = job_metadata_path(args.job).parent
+
+    metadata_path = log_dir / "metadata.json"
+    logfile_path = log_dir / "logs.txt"
+    metadata = json.loads(metadata_path.read_text())
+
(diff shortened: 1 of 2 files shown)


agent/cli/ehrql_telemetry.py
```diff
@@ -0,0 +1,124 @@
...
+    metadata = json.loads(metadata_path.read_text())
+
+    assert "ehrql" in metadata["container_metadata"]["Config"]["Image"], (
+        "Cowardly refusing to emit telemetry for non-ehrql job"
+    )
+
+    if args.logfile:
+        logfile_path = args.logfile
+    else:
+        logfile_path = metadata_path.parent / "logs.json"
```
evansd, line 42: I feel like we could make this significantly easier to use if we got it to do the conversion itself. I don't think it's much more complicated than:
```py
def parse_ehrql_logs(logs_text):
    process = subprocess.run(
        ["docker", "run" "--rm", "ghcr.io/opensafely-core/ehrql:v1" "/app/scripts/parse_logs.py"],
        input=text,
        encoding="utf-8",
        capture_output=True,
        check=True,
    )
    return process.stdout
```

Obviously I'll need to get my changes to ehrQL merged, and you might want to stub it out in tests, but feels like it could be a big usability win.

evansd (approved):

agent/cli/ehrql_telemetry.py
bloodearnest: done

agent/cli/ehrql_telemetry.py
```diff
@@ -0,0 +1,134 @@
...
+from pathlib import Path
+
+from opentelemetry import context, trace
+
+from agent.executors.local import job_metadata_path
+from common.tracing import setup_default_tracing
+
+
+def main(argv):
+    parser = argparse.ArgumentParser(description="db_telemetry")
```
evansd, line 16: I think it might be worth getting some of the extra detail from the PR description into the help text (or even just the docstring?). The PR is good for breadcrumbs, but I'm just thinking about trying to surface this information more easily the next time someone comes back to this.

evansd (commented):
I don't think `convert_ehrql_logs()` works as it stands (see comment).

Here's an extract from an ehrQL log in case it's useful as basic test case:
```
2025-07-01T16:56:30.687958737Z [info   ] Compiling dataset definition from analysis/dataset_definition/dataset_definition_vax.py
2025-07-01T16:56:46.227330349Z [info   ] Generating dataset
2025-07-01T16:57:00.993809188Z [info   ] Running query 001 / 331
2025-07-01T16:57:00.995883302Z [info   ] SQL:
2025-07-01T16:57:00.995931727Z           SELECT * INTO [#tmp_1] FROM (SELECT patients.patient_id AS patient_id 
2025-07-01T16:57:00.995952045Z           FROM (
2025-07-01T16:57:00.995970602Z                       SELECT
2025-07-01T16:57:00.995988488Z                           Patient_ID as patient_id,
2025-07-01T16:57:00.996005352Z                           DateOfBirth as date_of_birth,
2025-07-01T16:57:00.996023614Z                           CASE
2025-07-01T16:57:00.996043264Z                               WHEN Sex = 'M' THEN 'male' COLLATE Latin1_General_CI_AS
2025-07-01T16:57:00.996061649Z                               WHEN Sex = 'F' THEN 'female' COLLATE Latin1_General_CI_AS
2025-07-01T16:57:00.996210794Z                               WHEN Sex = 'I' THEN 'intersex' COLLATE Latin1_General_CI_AS
2025-07-01T16:57:00.996234454Z                               ELSE 'unknown' COLLATE Latin1_General_CI_AS
2025-07-01T16:57:00.996252317Z                           END AS sex,
2025-07-01T16:57:00.996265184Z                           NULLIF(DateOfDeath, '99991231') AS date_of_death
2025-07-01T16:57:00.996277761Z                       FROM Patient
2025-07-01T16:57:00.996290041Z                   ) AS patients UNION SELECT [PatientsWithTypeOneDissent].[Patient_ID] AS patient_id 
2025-07-01T16:57:00.996302988Z           FROM [PatientsWithTypeOneDissent]) AS anon_1
2025-07-01T16:57:01.010057734Z [info   ] scans logical physical read_ahead lob_logical lob_physical lob_read_ahead table
2025-07-01T16:57:01.010114642Z           1     0       0        0          0           0            0              PatientsWithTypeOneDissent
2025-07-01T16:57:01.010136538Z           1     0       0        0          0           0            0              Patient
2025-07-01T16:57:01.010240821Z [info   ] 0 seconds: exec_cpu_ms=3 exec_elapsed_ms=3 exec_cpu_ratio=1.0 parse_cpu_ms=0 parse_elapsed_ms=0 query_id=query 001 / 331
```

agent/cli/ehrql_telemetry.py
```diff
@@ -0,0 +1,134 @@
...
+    return int(
+        datetime.strptime(ts[0:-4], "%Y-%m-%dT%H:%M:%S.%f").timestamp() * 1_000_000_000
+    )
+
+
+def convert_ehrql_logs(logfile_path):  # pragma: nocover
+    process = subprocess.run(
+        [
+            "docker",
+            "run--rm",
```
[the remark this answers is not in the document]
bloodearnest: Hmm, yeah.

I am not keen a full functional test here. The only interface I want to commit to in tests is the json output. I don't want to embed the raw logs as test data, that feels brittle. And it would be slow.

But, as this shows, there needs to be some coverage of the docker invocation, so I guess its unavoidable.

I explicitly passed a path, as if I don't, all the tests need to provide dummpy logs.txt file to open that doesn't get used, but is required to exist. Deferring the opening of the file to the function simplifies the test setup a bit (I originally passed a string, and changed to this).

agent/cli/ehrql_telemetry.py
bloodearnest: Good idea, will do.

evansd (dismissed):

evansd (approved):

PR #1092, opened 2025-07-10
Telemetry for find_where

madwort:


bloodearnest:
Maybe we can measure manually and add as attribute to current span?

PR #1096, opened 2025-07-16
Provide a single run command

lucyb:
This starts all the services that are needed for local development in a single terminal.

### DIFF
--- a/justfile
+++ b/justfile
@@ -185,3 +185,6 @@ run-agent-service: devenv
 
 run-controller-service: devenv
     $BIN/python -m controller.service
+
+run:
+    { just run-app & just run-agent-service & just run-controller-service; }


madwort (approved):

justfile
```diff
@@ -185,3 +185,6 @@ run-agent-service: devenv
 
 run-controller-service: devenv
     $BIN/python -m controller.service
+
+run:
+    { just run-app & just run-agent-service & just run-controller-service; }
```
bloodearnest, line 190: An issue with this formulation is that it leaves the app and agent running after the controller service exits

How about a bash wrapper that cleans up? I run the webapp last, so it can still response to signals, e.g. reload

```suggestion
    set -e
    
    cleanup() {
        # get any background jobs started by this script and kill them
        jobs -p | xargs -r kill 2>/dev/null || true
    }
    
    # Set trap to cleanup on exit
    trap cleanup EXIT INT TERM
 
    just run-agent-service &
    just run-controller-service &
    just run-app
```

lucyb: Thanks for the alternative suggestion. I'm a little confused though, can you provide the steps you took where you ended up with both the app and the agent still running? When I run it locally (in zsh) and either force an exception to be raised in the controller or send a ctrl-c, all services exit as expected.

PR #1099, opened 2025-07-17
Spike using Django ORM to read DB and reimplement `calculate_workspace_state`

mikerkelly:
This is a proof-of-concept PR. Don't merge it.

Connect the controller webapp Django framework to the default location of its database and use `inspectdb` to generate some models.

In a management command, reimplement `calculate_workspace_state` using
querysets and a partition approach and compare perfomance with the extant
implementation that uses our homebrew ORM and some manipulation in Python.

Very brief initial profiling shows this as approximately 50 times faster on my
machine when tested against around 2000 jobs/action from the
https://github.com/opensafely/disparities-comparison repo.

This may demonstrate that this function can be optimized sufficiently to use
synchronously. Using the Django ORM may be a viable alternative to
optimizing the homebrew ORM code used. The Django ORM approach can be used
alongside code using the homebrew approach, which could enable a gradual
switchover, if that's where we want to arrive.

Clearly the homebrew-ORM-using code could be optimized further (eg it does some work
in Python that could be off-loaded to the database engine with `FIRST_VALUE`
and `PARTITION`, and I think we know the value-type unpacking is slow). Maybe
there's a better queryset approach, also.

My very brief assessment is that the queryset code needs at most three passes
through the rows, O(n), where the extant approach uses a sub-group and sorting
approach that may be O(n log n) or possibly O(n^2 log n).

### DIFF
--- a/controller/queries.py
+++ b/controller/queries.py
@@ -38,7 +38,7 @@ def calculate_workspace_state(backend, workspace):
             ordered_jobs = sorted(jobs, key=attrgetter("created_at"), reverse=True)
             latest_jobs.append(ordered_jobs[0])
 
-    return latest_jobs
+    return (all_jobs, latest_jobs)
 
 
 def group_by(iterable, key):
--- a/controller/webapp/management/commands/compare_calculate_workspace_state.py
+++ b/controller/webapp/management/commands/compare_calculate_workspace_state.py
@@ -1,0 +1,52 @@
+import time
+
+from django.core.management.base import BaseCommand
+from django.db.models import F, Window
+from django.db.models.functions import FirstValue
+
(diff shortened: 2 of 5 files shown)


mikerkelly:
There's an error in the management command timing where the queryset isn't evaluated, so the speedup calculations are wrong/irrelevant. The main slowness in the existing code is parsing every packed list/dict JSON field into Python objects, most of which aren't needed in this context. That can be addressed within the existing ORM.

See https://github.com/opensafely-core/job-runner/pull/1103 for further investigation and [Slack ](https://bennettoxford.slack.com/archives/C093H4M6TE2/p1752853202940619)for related discussions.

PR #1102, opened 2025-07-17
Add initial configuration and plumbing for restricted datasets

bloodearnest:
The initial permissions configuration is taken from the current yaml
config in opensafely cli, with a number of changes made to match the way
t1oo's work in job-server, which is where we intended this config to
live long term.

 - key by project rather than repo
 - use project slug rather than number, as job-runner doesn't know that.
 - use python rather than yaml

Have used the names of the tables as defined in opensafely-cli for now,
we may want to change the string values once we know what they will be.

The plumbing is fairly straightforward, it just sets a job env var with
the values of any permitted tables as a comma separate string for ehrql
to parse.

Lots of design discussion here: https://bennettoxford.slack.com/archives/C069YDR4NCA/p1752683993581859

### DIFF
--- a/controller/main.py
+++ b/controller/main.py
@@ -31,6 +31,7 @@ from controller.lib.database import (
     update_where,
 )
 from controller.models import Job, State, StatusCode, Task, TaskType
+from controller.permissions import datasets
 from controller.queries import (
     calculate_workspace_state,
     get_flag_value,
@@ -339,8 +340,13 @@ def job_to_job_definition(job, task_id):
     if image.startswith("stata-mp"):
         env["STATA_LICENSE"] = str(config.STATA_LICENSE)
 
-    if job.requires_db and job.repo_url in config.REPOS_WITH_EHRQL_EVENT_LEVEL_ACCESS:
-        env["EHRQL_ENABLE_EVENT_LEVEL_QUERIES"] = "True"
+    if job.requires_db:
+        if job.repo_url in config.REPOS_WITH_EHRQL_EVENT_LEVEL_ACCESS:
+            env["EHRQL_ENABLE_EVENT_LEVEL_QUERIES"] = "True"
+        job_request = get_saved_job_request(job)
+        env["EHRQL_PERMITTED_TABLES"] = ",".join(
+            datasets.PERMISSIONS.get(job_request["project"], [])
+        )
 
     # Jobs which are running reusable actions pull their code from the reusable
     # action repo, all other jobs pull their code from the study repo
--- a/controller/permissions/datasets.py
+++ b/controller/permissions/datasets.py
@@ -1,0 +1,72 @@
+PERMISSIONS = {
+    # project 37
+    "factors-associated-with-covid-19-vaccination": ["icnarc"],
+    # no project number
+    "immunosuppresant-medication": ["icnarc"],
+    # project 54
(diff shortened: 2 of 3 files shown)


bloodearnest:
This is the claude assisted script used to match repo to project. It generated yaml, I ended up hand converting that to python.

```
$ cat get.py
#!/usr/bin/env python3
"""
Script to transform repository_permissions.yaml by replacing repo org/path 
with project.slug and project.number information.

This script reads YAML from stdin and outputs the transformed version.
It uses Django ORM to query the job-server database for actual Project data.

Usage in Django shell:
    python manage.py shell < repo_permissions_transformer.py
"""

import yaml
import sys
from typing import Dict, Any, Optional
from dataclasses import dataclass

from jobserver.models import Workspace


@dataclass
class Project:
    """Represents a Project model from job-server"""
    slug: str
    number: int
    name: str = ""


def get_project_from_github_url(github_url: str) -> Optional[Project]:
    """
    Get a Project model from a workspace's GitHub URL.
    
    Args:
        github_url: The GitHub repository URL (e.g., "opensafely/study-repo")
        
    Returns:
        Project object if found, None otherwise
    """
    workspace = Workspace.objects.filter(repo__url__like=f"https://github.com/{github_url}").first()
    if workspace:
        return Project(
            slug=workspace.project.slug,
            number=workspace.project.number,
            name=workspace.project.name
        )
    return None


def load_repository_permissions() -> Dict[str, Any]:
    """Load the repository_permissions.yaml data from hardcoded path."""
    with open('repository_permissions.yaml', 'r') as f:
        return yaml.safe_load(f)


def transform_permissions_data(data: Dict[str, Any]) -> Dict[str, Any]:
    """
    Transform the repository permissions data by replacing repo org/path 
    with project.slug and project.number.
    """
    transformed_data = {}
    
    for repo_path, permissions in data.items():
        project = get_project_from_github_url(repo_path)
        
        if project:
            new_key = f"{project.slug}:  # project.{project.number}"
            transformed_data[new_key] = permissions
        else:
            new_key = f"{repo_path}:  # project.UNKNOWN"
            transformed_data[new_key] = permissions
    
    return transformed_data


def output_yaml(data: Dict[str, Any]):
    """Output the transformed YAML data in the correct format."""
    for key, value in data.items():
        print(key)
        if isinstance(value, dict):
            for sub_key, sub_value in value.items():
                if isinstance(sub_value, list):
                    formatted_list = "[" + ", ".join(f"'{item}'" for item in sub_value) + "]"
                    print(f"  {sub_key}: {formatted_list}")
                else:
                    print(f"  {sub_key}: {sub_value}")
        else:
            print(f"  {value}")


def main():
    """Main function to load, transform, and output the YAML data."""
    data = load_repository_permissions()
    transformed_data = transform_permissions_data(data)
    output_yaml(transformed_data)


# Execute main function when run in Django shell
if __name__ == "__main__":
    main()
else:
    main()
```

rebkwok (dismissed):
As mentioned in the slack thread, finding project by first associated workspace _could_ return the wrong project if any repo is associated with multiple projects. I definitely know of repos that are associated with multiple workspaces, but I don't know about different projects. Is it work a tweak of the script just to get the project for all associated workspaces for the repos and confirm that there's at most one?

rebkwok (approved):

PR #1104, opened 2025-07-18
Check git repo in the agent

### LINKED ISSUE #1077: Validate url path in agent
We validate the github url in the controller code: ~~https://github.com/opensafely-core/job-runner/blob/main/jobrunner/create_or_update_jobs.py#L122~~ https://github.com/opensafely-core/job-runner/blob/main/controller/create_or_update_jobs.py#L118

However, we should probably also validate this in the agent too, in case the controller is compromised.

There may be other security related things that we should re-validate in the agent too, but this is the obvious one.

The threat model here as that a user has compromised both the controller and the proxy that run on dokku4, so a attacker could submit a job from an arbitrary github org, or perhaps anywhere. Adding the check in the agent protects us against that scenario.

bloodearnest:
Fixes #1077, by also doing the github validation checks on the agent, as well as the controller.

Include some cleanup refactorings, and also had to plumb in the the branch information into the job definition the agent receives.  This is a bit messy, as we only have that data on the original job request json.  But did not want to change that whilst the RAP API step 2 work is happening. I expect in step 3 we'll have some space to rework how this data passes through the system.

### DIFF
--- a/tests/controller/test_main.py
+++ b/tests/controller/test_main.py
@@ -942,6 +942,12 @@ def test_job_definition_ehrql_permitted_tables(db, monkeypatch, project, expecte
             "project": project,
             "orgs": ["org1", "org2"],
             "backend": "test",
+            "workspace": {
+                "name": "workspace",
+                "repo": "repo",
+                "commit": "commit",
+                "branch": "main",
+            },
         }
     )
 


common/lib/github_validators.py
```diff
@@ -7,6 +7,18 @@ class GithubValidationError(Exception):
     pass
 
 
+def validate_repo_and_commit(allowed_gitub_orgs, repo_url, commit, branch):
+    if not allowed_gitub_orgs:
```
madwort, line 11: I realise this is preserving existing behaviour, but while we're here I'm just going to note that `allowed_github_orgs` as Falsey / empty list meaning "all github orgs are allowed" is the opposite of what I would have expected

controller/main.py
```diff
@@ -342,9 +342,14 @@ def job_to_job_definition(job, task_id):
     if job.requires_db and job.repo_url in config.REPOS_WITH_EHRQL_EVENT_LEVEL_ACCESS:
         env["EHRQL_ENABLE_EVENT_LEVEL_QUERIES"] = "True"
 
+    original_job_request = get_saved_job_request(job)
```
madwort, line 345: I reckon this might be worth an inline comment for future travellers (e.g. "Use the saved job request as at the time of writing `branch` is not part of the Job model, RAP API phase 2 is in progress, and we may ultimately switch ORM.")

madwort (dismissed):

controller/main.py
```diff
@@ -342,9 +342,18 @@ def job_to_job_definition(job, task_id):
     if job.requires_db and job.repo_url in config.REPOS_WITH_EHRQL_EVENT_LEVEL_ACCESS:
         env["EHRQL_ENABLE_EVENT_LEVEL_QUERIES"] = "True"
 
+    # we need branch information, which is currently only in the original job
+    # request data. We are trying to avoid changing the db schema whilst RAP
+    # API step 2 work is going on, so we just grab the original request. In
+    # future, I expect rewriting the OPM to use django will fix this.
```
madwort, line 348: typo?
```suggestion
    # future, I expect rewriting the ORM to use django will fix this.
```

madwort (dismissed):

madwort (approved):

PR #1110, opened 2025-07-24
Add a backend status endpoint

### LINKED ISSUE #1107: Create Controller `/backend/status/{ID}` endpoint
# Why are we doing this?

Currently, Job Server receives updates on backend status via `Flag` data packed into the HTTP headers of the `/jobs/` and `/jobs-requests/` Job Server endpoints when the Controller accesses them in its `sync` loop.

The RAP API V2 work is inverting the control flow and possibly removing the sync loop. There should be a new Controller endpoint that Job Server or other clients can call synchronously to discover the status of a backend. Clients might want to do this separately to any of the other endpoints.

# What are we doing

Creating a new Controller endpoint called something like `/backend/status/{ID}` that returns approximately the same information about Backend status that Job Server receives currently in its API HTTP headers.

It might get the status for one backend, or for all of them, TBD.

rebkwok:
Fixes #1107 

Adds an endpoint at `<backend name>/backend/status/`, which returns the flags that are currently passed in headers, in the same format, i.e.
```
{
    "flags": {
        <flag_id>: {"v": <flag value>, "ts": <flag timestamp>}
}
```

This isn't exactly the urlpattern stated in the issue (backend/status/<backend-name>), but follows the pattern of existing urls.

I've also moved the existing views to a `task_views` module, and added this one in a `rap_views` module, to keep the agent views separate from the rap API views. I haven't split the `urls.py`.


controller/webapp/views/auth/rap.py
```diff
@@ -0,0 +1,63 @@
...
+from django.http import JsonResponse
+
+from common import config as common_config
+from controller import config
+
+
+def require_client_token_backend_authentication(view_fn):
+    """
+    Ensure a valid client authentication token was received
+    and nsure the token is valid for the specified backend
```
madwort, line 12:
```suggestion
    and ensure the token is valid for the specified backend
```

tests/controller/webapp/test_rap_views.py
```diff
@@ -0,0 +1,177 @@
...
+
+def test_backends_status_view(db, client, monkeypatch, freezer):
+    freezer.move_to(TEST_DATESTR)
+    monkeypatch.setattr(
+        "controller.config.CLIENT_TOKENS",
+        {"test_token": ["test_backend1", "test_backend2"]},
+    )
+    headers = {"Authorization": "test_token"}
+
+    # set flag for unauthorsised backend
```
madwort, line 116:
```suggestion
    # set flag for unauthorised backend
```

madwort (dismissed):
Yeah I think this is good, as discussed on Slack maybe this introduces some tech debt - but that's as intended for the current initiative.

madwort (approved):

PR #1111, opened 2025-07-24
Add `create_jobs` telemetry

mikerkelly:


rebkwok (approved):

controller/create_or_update_jobs.py
```diff
@@ -83,36 +84,52 @@ def create_or_update_jobs(job_request):
...
-    insert_into_database(job_request, new_jobs)
-    return len(new_jobs)
+    tracer = trace.get_tracer("create_jobs")
+    with tracer.start_as_current_span("create_jobs") as span:
+        validate_job_request(job_request)
+        project_file = get_project_file(job_request)
+        span.set_attribute("backend", job_request.backend)
+        span.set_attribute("workspace", job_request.workspace)
+
+        with tracer.start_as_current_span("create_jobs.load_pipeline"):
```
bloodearnest, line 94: For these 4 measurements, I recommend not using new spans, and instead measuring the timing manually and adding as an attribute, for two reasons:

1) these spans are missing useful attributes to filter on. Whilst it is possible to filter on a parent span's attributes, its slower, and is best avoided.

2) each of these spans is an event that counts against our event budget.  Adding attributes to an existing span is free. Best practice seems to be stuffing more attributes into less spans, for both diagnostic and scalability reasons.

Note that this approach to using attributes rather than spans to measure timings only works if you are measuring 1 thing for that span. If you are measuring n, then you may well need to use individual spans.  But just be aware of the fact that emitting spans in large loop can generate a lot of events.

This is actually a fairly recent learning for me, so we don't have tooling to support this easily. But I've been thinking that something like this might work?

```
from contextlib import contextmanager
import time
from opentelemetry import trace

@contextmanager
def time_operation(attribute_name: str, span=None):
    """
    Context manager to time an operation and add it as an attribute to a span.
    
    Args:
        attribute_name: Name of the attribute to set on the span
        span: Optional span to add attribute to. If None, uses current span.
    """
    if span is None:
        span = trace.get_current_span()
    
    start_time = time.perf_counter()
    try:
        yield
    finally:
        end_time = time.perf_counter()
        duration = end_time - start_time
        span.set_attribute(attribute_name, duration)
```

bloodearnest (commented):
This is great idea, and should give some really useful data.

Left a few thoughts on the telemetry specifics.

controller/create_or_update_jobs.py
mikerkelly: I've added something along these lines, see what you think.

controller/create_or_update_jobs.py
```diff
@@ -33,6 +33,7 @@
 
 
 log = logging.getLogger(__name__)
+tracer = trace.get_tracer(__name__)
```
bloodearnest, line 37: Hah, actually using `__name__` is not a bad idea, but I didn't mean to stretch the analogy to loggers that far!

controller/create_or_update_jobs.py
```diff
@@ -91,19 +92,19 @@ def create_jobs(job_request):
         validate_job_request(job_request)
         project_file = get_project_file(job_request)
 
-        with tracer.start_as_current_span("create_jobs.load_pipeline"):
+        with time_for_span("duration_ms.load_pipeline", span):
```
bloodearnest, line 95: Good idea to include the units in the name, that's considered best practice AIUI.

However, the convention for naming for otel attrs would I think would be `load_pipeline.duration_ms`, i.e. operation first, details later. We'd then put any other load_pipline related things underneath it (e.g. `load_pipeline.action_count=n`) to group all load_pipeline things together. It's just an organisational convention, though.

There's lots of examples here: https://opentelemetry.io/docs/specs/semconv/general/attributes/

Also, I don't think that time_for_span is measuring ms, but seconds? At least, I think perf_counter is seconds.

However, I think ms is probably better, so maybe change time_for_span to report ms?

common/tracing.py
```diff
@@ -118,3 +120,24 @@ def set_span_attributes(span, attributes):
         clean_attrs[k] = v
 
     span.set_attributes(clean_attrs)
+
+
+@contextmanager
+def time_for_span(attribute_name: str, span=None):
```
bloodearnest, line 126: nitpick: not sure about the name. 

How about `duration_as_span_attr`? Bit mechanical, but maybe easier to understand?

bloodearnest (approved):
Looks good, a couple of suggestsions, and I think the duration units ms/s thing probably needs fixing?

common/tracing.py
```diff
@@ -118,3 +120,25 @@ def set_span_attributes(span, attributes):
...
+    if span is None:
+        span = trace.get_current_span()
+
+    start_time = time.perf_counter_ns()
+    try:
+        yield
+    finally:
+        end_time = time.perf_counter_ns()
+        # Convert duration to milliseconds
+        duration_ms = (end_time - start_time) / 1e6
```
bloodearnest, line 143: Quick note I think you probably want `* 1e6` here, unless ms stands for megaseconds :rofl:  

Also probably worth wrapping whole thing in an int(...)? Don't need floating point ms, I think

rebkwok: I don't think I do. Duration is in nanoseconds now. Can `int()` it though

rebkwok: I _did_ originally divide by `1e-6` :laughing:

bloodearnest (approved):

Issue #1126, opened 2025-07-31
[spike] Investigate OpenAPI specification for RAP API


rebkwok:
Investigate whether the [OpenAPI spec](https://spec.openapis.org/oas/latest.html) would be a useful tool for describing the RAP API.

rebkwok:
Branch open-api-spike

[OpenAPI spec](https://github.com/opensafely-core/job-runner/blob/open-api-spike/controller/webapp/api_spec/openapi.yaml) for one endpoint only (<backend>/backend/status)
These include descriptions and examples, which will appear in the docs (see below)

Testing with hypothesis, using [schemathesis](https://schemathesis.readthedocs.io/)

Run with:
```
just test-api-spec
```
(also runs the webserver and may be a bit flakey running together; can also run `just run-app` and in a separate terminal, `just schemathesis`)


Autogenerated [docs](https://github.com/opensafely-core/job-runner/blob/open-api-spike/controller/webapp/api_spec/api_docs.html) (with [redoc](https://github.com/Redocly/redoc))

To view docs, run the web server with `just run-app` and go to:
docs: http://localhost:3000/api-docs
spec (as json): http://localhost:3000/api_spec.json

rebkwok:
Autogenerated docs screenshot

<img width="1477" height="993" alt="Image" src="https://github.com/user-attachments/assets/2611dec1-8a53-45d2-aaae-0330ac9ce07a" />

madwort:
for ref. [otel unicode issue](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/3237)

PR #1130, opened 2025-08-01
OpenAPI spike

rebkwok:
See notes in #1126


controller/webapp/api_spec/openapi.yaml
```diff
@@ -0,0 +1,84 @@
...
+          in: path
+          description: name of backend
+          required: true
+          schema:
+            type: string
+            pattern: ^[A-Za-z0-9].+$
+            enum:
+              - test
+              - tpp
+              - emis
```
madwort, line 23: not sure it's wrong, but seems slightly weird to have `pattern` as well as `enum`

rebkwok: Oh this was my (misunderstood) attempt to make it not generate weird unicode. It can go.

madwort (approved):

PR #1142, opened 2025-08-11
Document our use of the OpenAPI specification

rebkwok:


DEVELOPERS.md
```diff
@@ -705,3 +706,8 @@ $dokku4 dokku run rap-controller python manage.py prepare_for_reboot --backend t
 == READY TO REBOOT ==
 Safe to reboot now
 ```
+
+## RAP API
+
+The RAP API is used by clients to communicate with the RAP Controller. We use the
```
mikerkelly, line 712:
```suggestion
The RAP API is used by non-agent clients to communicate with the RAP Controller. We use the
```

controller/webapp/api_spec/RAP_API_DEVELOPERS.md
```diff
@@ -0,0 +1,174 @@
...
+                    id: 1234567
+```
+
+Schemathesis will generate examples that use the values from each of the schema exampkes we provided.
+We can then set up our config in the tests so that e.g. the `test` backend is valid and authorised in the test call, the `tpp` backend is valid but not authorised, and the `emis` backend is invalid.
+
+## Documentation
+
+```
+just generate-api-docs
```
mikerkelly, line 147: Is it worth mentioning that these need to be committed? When should they be regenerated and updated?

Is it worth saying how to serve and view the docs in local dev?

mikerkelly (approved):
These are great, nice. Just a couple of minor comments.

controller/webapp/api_spec/RAP_API_DEVELOPERS.md
rebkwok: I think it probably is worth saying how to generate and view, although when the workflow PR goes in, the build will fail if they need re-generated (and I'll add a note about that too)

mikerkelly (approved):

PR #1151, opened 2025-08-20
Add dummy /rap/create endpoint

rebkwok:
This adds the API spec and schema validation for the create endpoint, but doesn't actually do any creating yet.


controller/webapp/views/validators/dataclasses.py
```diff
@@ -54,3 +54,22 @@ def from_request(cls, body_data: dict):
             job_request_id=job_request_id,
             actions=actions,
         )
+
+
+@dataclass
+class CreateRequest(RequestBody):
+    """
+    Represents a request to create ajob request
```
madwort: > create ajob request

typo

madwort (approved):
Looks good & thorough, the only question is whether it would be easier to clear the naming ambiguity of `job_request` vs `rap` now or leave it for later

controller/webapp/views/validators/dataclasses.py
```diff
@@ -43,29 +43,32 @@ def validate_schema(body_data, schema_name):
...
 
 
 @dataclass
 class CreateRequest(RequestBody):
     """
-    Represents a request to create a job request
+    Represents a request to create a RAP
+
+    Note: a RAP is currently still referred to as a `job_request`
+    in other parts of the code
```
madwort, line 71: :+1:

PR #1154, opened 2025-08-22
Basic implementation of rap/status

### LINKED ISSUE #1140: Create very basic Controller `/rap/status/` endpoint that returns a 200
# Why are we doing this?
RAP API V2 work is planning to replace most or all of the `controller.sync` loop with calls from Job Server to new controller endpoints. One of these is an endpoint to query RAP status (JobRequest/Jobs).

Having stubs of the different endpoints that initially do very little will allow us to make issues to exercise those in Job Server in parallel to the issue to implement the actual required endpoint behaviour. This also separates some of the boilerplate from the detailed implementation.

# What are we doing

Decided on `GET /rap/status` (vs. `GET /rap` - more REST-ish but has issues elsewhere).

Creating a new endpoint similar to the backend status and cancel endpoints that already exist in the controller. It should:

- have an openapi.yaml definition defined matching the current behaviour below.
- have a `RequestBody` dataclass defined that uses the definition for JSON validation as for `CancelRequest`
- have a view in `rap_views` decorated with `@get_backends_for_client_token` and `@validate_request_body`
- the view should return status 200 OK and a stub JSON body for now (maybe returning the name of the authorized backends)
- have a defined URL
- have some basic tests of this behaviour

madwort:
* receives a POST request with a list of rap_ids
* checks if the rap_ids are valid & the backend token is valid
* returns a dummy response
* super-simple
* fixes #1140


controller/webapp/views/rap_views.py
```diff
@@ -179,3 +183,56 @@ def create(request, *, token_backends, request_obj: CreateRequest):
...
+        jobs = find_where(Job, job_request_id=rap_id)
+        for job in jobs:
+            if job.backend not in token_backends:
+                return JsonResponse(
+                    {
+                        "error": "Not allowed",
+                        "details": f"Not allowed for backend '{job.backend}'",
+                    },
+                    status=403,
+                )
```
mikerkelly, line 226: I think the `Job` are all guaranteed to have the same backend within a RAP so is it sufficient to check just one? Although this is probably good belt-and-braces in case that changes in future I suppose.

controller/webapp/views/rap_views.py
```diff
@@ -179,3 +183,56 @@ def create(request, *, token_backends, request_obj: CreateRequest):
...
+    """
+    Get the status of an existing RAP (job request). Although this has no side-effects,
+    use POST rather than GET in order to avoid any complications around request
+    length etc.
+
+    token_backends: a list of backends that the client token (provided in the
+    request's Authorization header) has access to. Added by the
+    get_backends_for_client_token decorator.
+
+    See controller/webapp/api_spec/openapi.yaml for required request body
```
mikerkelly, line 202: The module docstring as a whole could get some breadcrumbs like this so it's clear what the whole thing relates to. Doesn't have to be done here though.

mikerkelly (approved):
Looks good, thanks.

I think this could have fulfilled the Issue requirements without actually checking the RAP ID related Jobs even existed and left that for later, but it's also fine to have done that here. I'm guessing it will persist as a way of ending early before doing the expensive work.

controller/webapp/views/rap_views.py
```diff
@@ -179,3 +183,56 @@ def create(request, *, token_backends, request_obj: CreateRequest):
...
+    """
+
+    # TODO: This aborts as soon as we encounter an error. Should we return statuses for
+    # those rap_ids which are valid/allowed?
+    for rap_id in request_obj.rap_ids:
+        if not exists_where(Job, job_request_id=rap_id):
+            return JsonResponse(
```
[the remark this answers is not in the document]
madwort: yeah, I was just doing the minimum to generate an error case

PR #1157, opened 2025-08-27
Update dotenv-sample client token to match job-server

rebkwok:
In a local environment running job-server and the rap controller together, <BACKEND>_CLIENT_TOKENS must match the value of job-server's RAP_API_TOKEN in order for job-server to be able to authenticate with the controller.

### DIFF
--- a/dotenv-sample
+++ b/dotenv-sample
@@ -61,7 +61,8 @@ TEST_JOB_SERVER_TOKEN=pass
 # and is a comma-separated list of client tokens that are allowed to access information
 # about this backend. These may be duplicated for clients that can request information about
 # multiple backends.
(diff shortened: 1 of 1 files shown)


dotenv-sample
```diff
@@ -61,7 +61,8 @@ TEST_JOB_SERVER_TOKEN=pass
 # and is a comma-separated list of client tokens that are allowed to access information
 # about this backend. These may be duplicated for clients that can request information about
 # multiple backends.
-TEST_CLIENT_TOKENS=token
+# Note: his corresponds to RAP_API_TOKEN in job-server
```
mikerkelly, line 64: typo "his: => "this"

mikerkelly (approved):

PR #1185, opened 2025-09-18
Return run_command as '' instead of None for run_all jobs

madwort:
* The db schema on job-server does not allow a null value for the
  run_command, so by returning the empty string we can make the
  client code a little simpler
* this is also consistent what the previous run loop implementation
  did - for reasons which I had not got to the bottom of
* the api spec does say that this field is a string, so `None` isn't ideal


mikerkelly (approved):
Looks good to me and more in line with API definition.

Maybe an argument for checking against the schemas on what the controller sends as well as receives to catch cases like this.

PR #1197, opened 2025-10-02
Disable the jobs status sync for the test backend

madwort:


tests/test_integration_new_sync.py
```diff
@@ -28,9 +28,9 @@ def set_agent_config(monkeypatch, tmp_work_dir):
     monkeypatch.setattr("agent.config.USING_DUMMY_DATA_BACKEND", True)
     # Note that as we are running ehrql actions in this test, we need to set
     # the backend to a value that ehrql will accept
-    # RAP API stage 2: Use `emis` rather than `test` so that we can disable the
+    # RAP API stage 2: Use `test` rather than `test` so that we can disable the
```
rebkwok, line 32: Probably want to just remove this comment?

rebkwok (approved):
The new-syn version of the test is pleasingly straightforward :)
Just a comment on a comment that doesn't make sense anymore

tests/test_integration_new_sync.py
madwort: hah yes!

rebkwok (approved):

PR #1202, opened 2025-10-08
Log less when cancelling multiple actions.

mikerkelly:
Probably this is Job Server cancelling a whole request.  Let's just log the number of actions in that case.  Technically partial cancellations are possible but Job Server will not do that at present, and this avoids spamming 100s of KB when cancelling large requests.

Add a test for multiple actions to exercise this, though we don't test the logged values to avoid brittle tests.


controller/webapp/views/rap_views.py
```diff
@@ -177,11 +177,20 @@ def cancel(request, *, token_backends, request_obj: CancelRequest):
             status=404,
         )
 
-    log.info(
-        "Cancelling actions for job_request %s: %s",
-        request_obj.rap_id,
-        request_obj.actions,
-    )
+    if len(actions_to_cancel) > 1:
```
rebkwok, line 180: I know they should equate to the same thing, but it feels a bit odd to look at `len(actions_to_cancel)` here but then log `len(request_obj.actions)`.

rebkwok (dismissed):

PR #1217, opened 2025-10-16
Filter active jobs in the controller by backends allowed for the client token

rebkwok:
In the status endpoint, we check for jobs that are active in the controller, but which the client didn't request. We need to also filter those jobs by backends which the client has access to.


tests/controller/webapp/test_rap_views.py
```diff
@@ -847,13 +847,15 @@ def test_status_view_tracing_with_unexpected_rap_ids(db, client, monkeypatch):
     job1 = job_factory(action="action1", backend="test")
     job2 = job_factory(action="action1", backend="test")
     job3 = job_factory(action="action1", backend="test1")
+    job_factory(action="action1", backend="test1")
```
madwort: should be backend `test` - currently is the same as `job3`

```suggestion
    job_factory(action="action1", backend="test")
```

madwort: oh yes, sorry!

madwort (approved):

PR #1219, opened 2025-10-17
Remove openapi-spec-validator dependency

rebkwok:
A dependency conflict is causing pip-compile to downgrade openapi-spec-validator significantly (from 0.7.2 to 0.5.5), to a version where it doesn't work with our api spec.

Since it's not used in any CI tests, only for local development when adding to the API spec, and there are alternatives (the swagger editor and the schemathesis tests), let's just remove it.

More notes on my investigation in [slack thread](https://bennettoxford.slack.com/archives/C069YDR4NCA/p1760516958920349)


madwort (approved):
Really weird error, I think you've spent enough time investigating so agree with removing.

PR #1232, opened 2025-10-23
Update documentation with RAP API implementation

### LINKED ISSUE #1215: Update documentation
Update DEVELOPERS.md to the new state of play.

Including:
- Operating principles - update description of communication between job-server and controller
- Architecture - update controller and controller app sections in particular
- Consider updating the part on running jobs on the test backend; as it stands, this will create jobs that job-server doesn't know about and may cause honeycomb alerts (we should at least warn about it).

Providence-o:
Fixes #1215

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -8,6 +8,11 @@
   - [Testing](#testing)
 - [Operating Principles](#operating-principles)
 - [Architecture](#architecture)
+  - [The RAP Agent](#the-rap-agent)
+  - [The RAP Controller](#the-rap-controller)
+  - [The RAP Controller APIs](#the-rap-controller-apis)
+  - [Configuration](#configuration)
+  - [Job State](#job-state)
 - [Running jobs on the test backend](#running-jobs-on-the-test-backend)
 - [job-runner docker image](#job-runner-docker-image)
 - [Database schema and migrations](#database-schema-and-migrations)
@@ -16,7 +21,6 @@
   - [Turn manual database maintenance mode on/off](#turn-manual-database-maintenance-mode-onoff-on-a-specific-backend)
   - [Pause a backend](#pause-a-backend)
   - [Prepare for reboot](#prepare-for-reboot)
-- [RAP API](#rap-api)
 
 
 ## Local development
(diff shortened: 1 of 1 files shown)


DEVELOPERS.md
```diff
@@ -470,30 +440,41 @@ together as a [service](./agent/service.py).
...
+Jobs are created via the [RAP API](#rap-api) `create` endpoint and the [create_or_update_jobs](./controller/create_or_update_jobs.py) module does the bulk of the work.
 
 Only the Controller has access to its database of Jobs and Tasks.
 
-### The RAP Controller API
+### The RAP Controller APIs
 
 This is a very simple [Django application](./controller.webapp/) that allows external
-applications and users (currently just the Agent) to communicate with the Controller.
+applications and users to communicate with the Controller.
```
rebkwok: Maybe add something like:
Users of the RAP Controller APIs may be Agents inside backends (which consume the [Controller Task API](#controller-task-api)) or clients in the outside world e.g. job server (which consume the [RAP API](#rap-api)).

DEVELOPERS.md
```diff
@@ -470,30 +440,41 @@ together as a [service](./agent/service.py).
...
 - `/<backend>/task/update/`: receives information about a task and updates the controller database
 
+#### RAP API
+
+The RAP API is used by non-agent clients to communicate with the RAP Controller. It has four endpoints which are defined in the [rap api views](./controller/webapp/views/rap_views.py) module. They are authenticated by the tokens received as part of jobserver's request.
+
+- `/backend/status/`: returns information about the status of backends. This endpoint is polled every minute.
+- `/rap/cancel/`: cancels jobs within a specific job request.
+- `/rap/create/`: creates jobs for a job request.
+- `/rap/status/`: returns current status of jobs after the controller database has been updated by the `create_or_update_jobs` module. This endpoint is polled every minute.
```
rebkwok:
```suggestion
- `/rap/status/`: returns current status of jobs in the controller database for the requested RAP IDs. This endpoint is polled by job-server every minute.
```
(The create_or_update_jobs module is used by the /create/ endpoint; the status endpoint will return new jobs that were created by the create endpoint, but only when job-server asks for them. It'll also return new job status as jobs are updated by the agent, so I think the relevant thing is that it'll return the jobs as they currently are in the controller db, regardless of what updated them)

DEVELOPERS.md
rebkwok:
```suggestion
The RAP agent communicates with the RAP controller via an [http API (a Django app)](controller.webapp/), using the [Controller TasksAPI](#controller-task-api)
```

rebkwok (approved):

PR #1258, opened 2025-11-07
Ensure all mutations are in a transaction.

bloodearnest:
To use sqlite concurrently, we *have* to use a transaction for writes,
or else we don't get the good concurrent write semantics we want.
There were some cases where we were not, which likely caused a db lock
in production.

We also use larger, multi-write-statement transactions in some places,
to ensure consistancy, and we don't want get into nested transactions.

So, this change adds an ensure_transaction decorator, and wraps all the
database mutation methods with it.

This ensures we can't accidentally perform a write that is not in
a transaction, but we can still use larger outer transactions to group
related writes together.

See discussion here: https://bennettoxford.slack.com/archives/C069YDR4NCA/p1762466776554599

### DIFF
--- a/controller/lib/database.py
+++ b/controller/lib/database.py
@@ -11,6 +11,7 @@ shouldn't be too large a job.
 
 import contextlib
 import dataclasses
+import functools
 import json
 import logging
 import sqlite3
@@ -31,6 +32,18 @@ MIGRATIONS = {}
 
 
 tracer = trace.get_tracer("db")
+
+
+def ensure_transaction(f):
+    @functools.wraps(f)
+    def wrapper(*args, **kwargs):
+        conn = get_connection()
+        if conn.in_transaction:
+            return f(*args, **kwargs)
+        with transaction():
+            return f(*args, **kwargs)
+
+    return wrapper
 
 
 def databaseclass(cls):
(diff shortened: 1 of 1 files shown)


bloodearnest:
> This is nifty, and solves our problem. I wonder though if it would be better to have it throw an error if it's not in a transaction. Given that all this is exercised by the tests that should force us to explicitly add a transaction wherever we need it. And it might be easier to reason about if its explicit rather than implicit like this.

That is the approach I tried first.

However, there's a lot of adhoc db inserts in test setup that would now need have transactions wrapped around them, which is feels like an unpleasant burden to write and maintain.

I went this route as it DTRT. It does arguably hides potential mistakes when things should be in an outer transaction but aren't.

:thinking: 

We could maybe add more factories that do the transaction stuff for us? For example, there's a lot of calls to insert_task() in the test setup. Maybe if we had a factory.insert_task() method that did the transaction?

evansd:
Ah yeah, I see. I guess the proper fix is, as you say, to handle this all in test factories but that does feel like more work than we'd want to do right now.

I wonder if we should create an issue to come back to it. It might be the kind of tedious work that would suit an LLM.

bloodearnest:
https://github.com/opensafely-core/job-runner/issues/1260

Issue #1257, opened 2025-11-07
Pass NDOO config from RAP API to ehrQL


suzannehamilton:
We are [updating ehrQL to filter out patients who have requested a National Data Opt-Out](https://github.com/opensafely-core/ehrql/issues/2589).

This config may be project-specific. In this case, [our proposal](https://docs.google.com/document/d/1h0e4eyLpBVivdZmztC8T6yFrIk8KNfg4WCz3rmSnulk/edit?tab=t.0#heading=h.h6vpt53fcb7m) is to store a list of projects which can continue to query the existing population, and pass the information to ehrQL so it applies the filtering consistently within a project.

We are proposing to store this list in the job-server codebase. When a user submits a job, job-server will pass the NDOO config to the RAP API along with other information about the job request.

This means the RAP API needs to be updated to receive this information, store it and share it with the RAP agent. The RAP agent should then [pass it to ehrQL](https://github.com/opensafely-core/ehrql/issues/2589).

The agent [already passes dataset and event-level data permissions to ehrQL](https://github.com/opensafely-core/job-runner/blob/7e7bec55a0c4d6e83e9d8e7b450a579a8258a374/controller/main.py#L358-L363). We can extend this to include the NDOO config.

We are planning to [move the dataset permissions config to the job-server repo](https://github.com/opensafely-core/job-server/issues/5163), and we might update the T1OO permissions to fit the same pattern. See the [permissions proposal](https://docs.google.com/document/d/1h0e4eyLpBVivdZmztC8T6yFrIk8KNfg4WCz3rmSnulk/edit?tab=t.0#heading=h.h6vpt53fcb7m) for how we could make this all consistent.

We should not start filtering patients immediately, so we need to add a feature flag or something similar so that we only apply it when we're ready to switch over to the new population.

suzannehamilton:
This is very similar to the GP activation config change in #1256. I've kept them separate in case we need to implement one of them sooner.

suzannehamilton:
@rebkwok Neither of them are in scope for this issue. But we should keep them in mind when designing the API, because we're likely to want all the permissions to be consistent.

rebkwok:
> Regarding the name features, Becky, am I correct in thinking that "access to features - e.g. event-level data" isn't the feature flag, but the project-level permission to access event-level data?

[...]

PR #1270, opened 2025-11-19
Add analysis_scope parameter to OpenAPI spec for rap/create endpoint

rebkwok:
See #1257 for discussion of the naming and shape of this parameter. Names potentially still TBC.


evansd (approved):
Lots more discussion in Slack about this:
https://bennettoxford.slack.com/archives/C080S7W2ZPX/p1763732541039939

PR #1275, opened 2025-11-28
Removing SavedJobRequest: part 1

rebkwok:
In preparation for removing the SavedJobRequest, add the attributes that we use (`project`, `orgs`, `branch`, `created_by`) to the `Job` model, and populate them from the request body posted to the `/rap/create` endpoint.

Add a management command to populate all existing Jobs with the attributes retrieved from their SavedJobRequest.

Doesn't use any of the new `Job` attributes yet; to be done after this one is merged and the management command run.

(I've run the management command on a copy of the prod db to ensure it works on real data)

### DIFF
--- a/controller/webapp/management/commands/backpopulate_from_saved_job_request.py
+++ b/controller/webapp/management/commands/backpopulate_from_saved_job_request.py
@@ -1,0 +1,63 @@
+import logging
+
+from django.core.management.base import BaseCommand
+
+from controller.lib import database
+from controller.models import Job
+from controller.queries import get_saved_job_request
+
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """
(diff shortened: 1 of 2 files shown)


controller/webapp/management/commands/backpopulate_from_saved_job_request.py
```diff
@@ -0,0 +1,63 @@
...
+from controller.models import Job
+from controller.queries import get_saved_job_request
+
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """
+    Development utility for creating and submitting a RAP without having a
```
evansd: I start every management command by copying an existing one too!

I guess it doesn't really matter given this is a one-shot command.

evansd (approved):
Looks good. I appreciate the careful thought that's gone into the sequencing here.

PR #1278, opened 2025-12-02
Pass NDOO permission received via RAP API to ehrql

### LINKED ISSUE #1257: Pass NDOO config from RAP API to ehrQL

rebkwok:
- Adds the new `analysis_scope` field to Job model
- Populates analysis_scope using a combination of data received in the RAP API create request (currently just NDOO permission) and permissions stored in this repo
- Converts analysis scope to the flat list that ehrql expects in its `EHRQL_PERMISSIONS` env variable

Fixes #1257

### DIFF
--- a/controller/tracing.py
+++ b/controller/tracing.py
@@ -267,6 +267,10 @@ def trace_attributes(job, results=None):
         attrs["job.base_revision"] = results.base_revision
         attrs["job.base_created"] = results.base_created
 
+    if job.analysis_scope:
+        for key, value in job.analysis_scope.items():
+            attrs[f"job.analysis_scope.{key}"] = ",".join(value)
+
     return attrs
 
 
--- a/tests/controller/test_tracing.py
+++ b/tests/controller/test_tracing.py
@@ -64,6 +64,46 @@ def test_trace_attributes(db):
     }
 
 
+def test_trace_attributes_analysis_scope(db):
+    job = job_factory(
+        workspace="workspace",
+        action="generate_dataset",
+        commit="commit",
+        action_repo_url="action_repo",
+        action_commit="commit",
+        status_message="message",
+        backend="test",
+        project="project",
+        orgs=["org1", "org2"],
+        user="testuser",
+        requires_db=True,
+        analysis_scope={"foo": ["bar1", "bar2"]},
+    )
+
+    attrs = tracing.trace_attributes(job, results=None)
+
+    assert attrs == {
+        "job.backend": "test",
+        "job.id": job.id,
+        "job.request": job.rap_id,
+        "job.workspace": "workspace",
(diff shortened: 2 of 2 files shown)


controller/main.py
```diff
@@ -350,12 +351,17 @@ def job_to_job_definition(job, task_id, image_sha=None):
...
-        env["EHRQL_PERMISSIONS"] = json.dumps(permissions)
+        # Build the full analysis scope dict to ensure that we have nay permissions that
+        # are stored within this repo (this should be unnecessary once any jobs created before
+        # controller.create_or_update_jobs.create_jobs() did this have completed)
+        analysis_scope = build_analysis_scope(
+            job.analysis_scope, job.project, job.repo_url
+        )
+        env["EHRQL_PERMISSIONS"] = json.dumps(
+            list(chain.from_iterable(analysis_scope.values()))
+        )
```
evansd: I guess at some point we could update ehrQL to accept the structured object directly here, but I don't think there's any immediate need to do so.

rebkwok: yes, I assumed we might want to in future, but for now we don't need to change ehrql if we just turn it into what it currently expects

PR #1284, opened 2025-12-04
Add breadcrumbs linking token names to env vars

evansd:
Given an arbitrary Github token that's about to expire we want to be able to search across all repos to find where it is used.


madwort (approved):

common/config.py
```diff
@@ -39,6 +39,14 @@
 GITHUB_PROXY_DOMAIN = os.environ.get(
     "GITHUB_PROXY_DOMAIN", "github-proxy.opensafely.org"
 )
+# This is a token belonging the `opensafely-readonly` Github user:
```
madwort: "belonging to"

madwort (approved):

PR #1394, opened 2026-04-15
Cache resource usage for each backend

### LINKED ISSUE #1387: Expensive n+1 query for large numbers of pending jobs.
In the handle_pending_job loop, get_reason_not_started is called for every single pending job. This calls `find_where()` for all RUNNING jobs.

When we have a lot of PENDING jobs, (and there for also a full set of RUNNING jobs), this is called constantly on every pending loop for every job. Because this involve deserializing all ~15 running jobs into python every time, its taking up a lot of cpu.

e.g. we just saw 6500+ PENDING jobs, and 93% of our cpu time is in find_where, and pegged at 100% cpu.

Here's a py-spy flamegraph that shows this.

Image

Potential fixes

1) make the query super quick, and not do the full ORM deserialization. We don't really have the tooling for this currently, but it could work.

2) cache the set of running jobs once at the top of the loop, and reuse it for all pending jobs calculations this iteration of the loop. That might mean that jobs towards the end of the loop get a slightly stale view of running jobs, but if they are at the end of the loop (later jobs), they are not likely to be scheduled this iteration anyway.

3) ???

Detailed discussion of his here: https://bennettoxford.slack.com/archives/C069YDR4NCA/p1775139519310739

evansd:
As described in #1387, the existing code recalculates the current resource usage each time it considers whether to start a job. This involves retrieving some large-ish objects from the database and so when we have a very large number of pending jobs this adds up to a significant amount of work.

However the resource usage only changes when the set of running jobs changes, so we can cache the results so long as we invalidate this cache whenever we start or stop a job. Fortunately there's a common code path we can hook into here to achieve this.

For good measure we also reset the cache at the beginning of each loop over the active jobs. This means that if, somehow, there are changes to job state which have happened out-of-band then we'll correct our view of the world as soon we run the next loop.

I've tested this locally on a job queue container 12 running jobs and 2,000 pending ones. Previously each full loop over the queue took 120 seconds. With this patch it takes 850ms.

Closes #1387

### DIFF
--- a/controller/main.py
+++ b/controller/main.py
@@ -59,6 +59,9 @@ def main(exit_callback=lambda _: False):
 
 
 def handle_jobs():
+    # Not strictly necessary but minimises the potential for weird issues
+    invalidate_resource_usage_cache()
+
     log.debug("Querying database for active jobs")
     active_jobs = find_where(Job, state__in=[State.PENDING, State.RUNNING])
     log.debug("Done query")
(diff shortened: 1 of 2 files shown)


rebkwok (dismissed):
All of this will be hit by existing tests, but do we want to explicitly assert that the cache is invalidated when we expect it to be?

rebkwok (approved):

PR #1408, opened 2026-04-24
Temporary check for migrations at deploy time

rebkwok:
Write a file to the same location as the db file during the migrate command, so we can check whether our Procfile release command has access to mounted storage during dokku deploys.


evansd (approved):
🤞

PR #1418, opened 2026-05-07
Upgrade schemathesis and fix tests

rebkwok:
We were passing auth headers in the tests by adding them on the pytest `case` fixture itself. This seems to have worked in the past, but with a recent upgrade, it looks like it means that the endpoint is called with the headers set directly on the `case` (so the response was correctly 200), even when schemathesis generated test case metadata with no headers, and expected to get a 401.  It also confusingly meant that the example `curl` command that it reported to use to replicate the test case _did_ include authorization headers.

In fact we should be passing the headers to the `case.call()` (which is sort of how it's described in the docs, for call_and_validate() https://schemathesis.readthedocs.io/en/stable/tutorials/pytest/

We also don't need separate tests for no/bad auth, because schemathesis will generate those as test cases anyway (if it didn't, we'd know because there would be no 401 cases for any endpoints)

### DIFF
--- a/requirements.uvmirror.txt
+++ b/requirements.uvmirror.txt
@@ -215,7 +215,7 @@ ruamel-yaml-clib==0.2.8
 ruff==0.15.12
 ruyaml==0.91.0
     # via opensafely-pipeline
-schemathesis==4.15.2
+schemathesis==4.16.1
 setuptools==82.0.1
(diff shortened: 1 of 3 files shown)


alarthast (approved):
LGTM 😄 I found the commit message really easy to understand.

PR #1422, opened 2026-05-12
Fix intermittent daemon threads error and other test flakiness

rebkwok:
Remove the `_catch_shutdown_errors` threading excepthook. It tried to suppress errors with a try/catch during shutdown, and to print the error in case stdout was available. However, the threading errors weren't being suppressed properly, and possibly the print was actually triggering them. In any case, we've been seeing them quite frequently, and not with the except print statement.

The issue appears to actually be with the agent.task_api's module-level requests.Session. We now have a session-scoped fixture that sets `Connection: close` on `agent.task_api`'s Session. This ensures the live server's daemon request-handler threads exit cleanly after each response rather than persisting into interpreter shutdown. It also adds a retry, for the cases where the Connection close wasn't registered by django in the tests, which resulted in an intermittent RemoteDisconnected error.

Also adds a regression check in fixture teardown that fails loudly if any `process_request_thread` daemon threads are still alive after the session ends.

I've tested this by running the docker tests repeatedly locally. Without this fix, they hit the threading error within the first few runs. With it, they ran 50x without hitting it. However, they did highlight another intermittent error in the few tests that make actual calls to ghcr.io, so there's also a retry added there; it retries a max of 3 times with a 1sec delay, so although it's a change in production code, it shouldn't have a huge impact. With this, my repeated docker tests ran successfully 50/50 times.

### DIFF
--- a/agent/lib/docker.py
+++ b/agent/lib/docker.py
@@ -7,6 +7,7 @@ import logging
 import os
 import socket
 import subprocess
+import time
 import urllib.parse
 
 from agent import config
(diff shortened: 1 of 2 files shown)


evansd (approved):
Oh amazing! Thanks Becky (and I guess thanks Claude as well). These intermittent failures have been _really_ frustrating me and I'm so pleased we've finally tracked it down.

Given that the module level session seems to be responsible for the complexity here I wonder if there's another way of factoring things that gets rid of that. But that's a question for another day.

I think the Docker retry behaviour is useful to have in prod. Obviously we have a full task-level retry that should ensure we handle network flakiness but it's much neater (and doesn't cause spurious alerts) if we can retry at this level.

rebkwok:
Yeah, I did consider getting rid of the module level session once I realised that Claude was (eventually) right about that being the issue. But I remember Simon having valid reasons for it being module level (presumably for performance, we [could, but don't, use it to persist headers](https://requests.readthedocs.io/en/latest/user/advanced/#session-objects)), and afaik it only causes issues in tests, and specifically because we use the live_server fixture. Maybe there's a different refactor that allows it to be persisted across requests but doesn't just define it at the module level. I might spent half an hour investigating that, but if it's not easy, I'll just go with what we've got so far.

evansd:
Oh, I think we should definitely merge this as it's such an improvement over the status quo and the complexity is only the test config.

bloodearnest:
Huh, how interesting. I've been using module level Session instances for decades w/o encountering this issue before.

The benefit of persistant sessions is that we can have a persistent connection pool, and don't have to constantly make new TLS connections every time we make a HTTP request.

evansd:
Oh yeah, I think persistent sessions are definitely important. I was just wondering about managing that explicitly rather than using a module global. But as you say, that's a pretty standard pattern and I've not seen it cause this problem before.

rebkwok:
It's specifically the interaction with the django live_server in tests that causes the problem. Otherwise the module level session seems to be fine.

PR #1426, opened 2026-05-13
Refactor task api session and fix test flakiness

rebkwok:
Remove the _catch_shutdown_errors threading excepthook. It tried to suppress errors with a try/catch during shutdown, and to print the error in case stdout was available. However, the threading errors weren't being suppressed properly, and possibly the print was actually triggering them. In any case, we've been seeing them quite frequently, and not with the except print statement.

The issue appears to actually be with the agent.task_api's module-level requests.Session, which was interacting badly with the live server's daemon request-handler. 

This PR refactors the task_api into a TaskAPI client which is used as a context manager to ensure that the session is closed on exit. The agent loops run inside the context manager, so the Session is reused across requests, as we want. In tests, we have a fixture that uses the client as a context manager too, and means that each test closes the session properly. 

I've tested this by running the docker tests repeatedly locally. Without this fix, they hit the threading error within the first few runs. With it, they ran 50x without hitting it. However, they did highlight another intermittent error in the few tests that make actual calls to ghcr.io, so there's also a retry added there; it retries a max of 3 times with a 1sec delay, so although it's a change in production code, it shouldn't have a huge impact. With this, my repeated docker tests ran successfully 50/50 times.

Note: there's an [alternative PR ](https://github.com/opensafely-core/job-runner/pull/1422)that leaves the main code as it is and handles the test errors in a fixture. That works, but the fixture workaround is complex. Although this solution involves more code change, I think it's better.

### DIFF
--- a/agent/main.py
+++ b/agent/main.py
@@ -30,17 +30,18 @@ def main(exit_callback=lambda _: False):  # pragma: no cover
     log.info("agent.main loop started")
     api = get_executor_api()
 
-    while True:
-        active_tasks = handle_tasks(api)
+    with task_api.TaskAPI() as task_api_client:
+        while True:
+            active_tasks = handle_tasks(api, task_api_client)
 
-        if exit_callback(active_tasks):
-            break
+            if exit_callback(active_tasks):
+                break
 
-        time.sleep(common_config.JOB_LOOP_INTERVAL)
+            time.sleep(common_config.JOB_LOOP_INTERVAL)
 
 
-def handle_tasks(api: ExecutorAPI | None):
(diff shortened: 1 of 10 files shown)


evansd (commented):
Directionally, this feels correct to me; as in, I think having an explicitly constructed client instance is probably the way to go. But I'm not convinced that just threading it through all the existing functions (which already have an executor API instance threaded through them) is the right way to factor this. If we're going to do something like this then I want to step back and make sure that the end result is something simple and coherent on its own terms.

In that sense, although the other PR is more of a hack it worries me less because it's contained in the test code.

PR #1431, opened 2026-05-19
Fix misleading prepare for reboot status

alarthast:
There is a niche corner case where a dev runs the `prepare_for_reboot` status report before the `prepare_for_reboot` command itself, in which case the report misleadingly displays "Safe to reboot now".

Even though this is an unlikely scenario, we probably don't want to display the "Safe to reboot" message in any circumstances other than when it is truly safe. When the backend is not paused, prompt the dev to run `prepare_for_reboot` instead.

Maybe we could combine the command and the status check by adding a `while` loop that runs the status check until it's safe to reboot, but that seems like a task for another day and I have gone with the simplest change for now.

### DIFF
--- a/controller/cli/prepare_for_reboot.py
+++ b/controller/cli/prepare_for_reboot.py
@@ -14,7 +14,6 @@ from controller.queries import get_flag_value
 
 
 def main(backend, status=False, skip_confirm=False):
-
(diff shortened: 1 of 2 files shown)


rebkwok (approved):

alarthast:
Thanks @rebkwok! Forgot to post this in Slack

PR #1435, opened 2026-05-21
Suppress opensafely-pipeline ProjectWarnings at runtime

rebkwok:
Add a warnings filter that silences noisy-but-expected ProjectWarnings pipeline library. They're only intended as warnings to users and are irrelevent in the controller/agent and just fill logs up with noise.

common.config is imported early by both controller and agent, so seems like a suitable place for it.

Also adds a regression test (which I confirm fails pre-ignore-filter)

### DIFF
--- a/common/config.py
+++ b/common/config.py
@@ -1,5 +1,9 @@
 import os
+import warnings
 from pathlib import Path
+
+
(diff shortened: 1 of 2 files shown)


evansd (approved):
Thanks!

PR #1437, opened 2026-05-22
Make local repo support an explict setting

evansd:
We want to support local git repos in tests and in development but not in production. There's currently no exploitable pathway here given our deployment configuration, but it's best to be defensive.

### DIFF
--- a/common/config.py
+++ b/common/config.py
@@ -63,3 +63,5 @@ ALLOWED_GITHUB_ORGS = (
     .strip()
     .split(",")
 )
+
+ALLOW_LOCAL_GIT_REPOS = os.environ.get("ALLOW_LOCAL_GIT_REPOS", "").lower() == "true"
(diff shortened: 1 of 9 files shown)


common/config.py
```diff
@@ -63,3 +63,5 @@
     .strip()
     .split(",")
 )
+
+ALLOW_LOCAL_GIT_REPOS = os.environ.get("ALLOW_LOCAL_GIT_REPOS") == "true"
```
rebkwok: Maybe add a `.lower()` so it'll match True as well; other things in the `dotenv-sample` use `True`; when I tried this out locally, I initially just added another True var to my .env

rebkwok (commented):
This works fine, but I think it needs some developer docs updates.

The instructions for [running a job locally](https://github.com/opensafely-core/job-runner/blob/main/DEVELOPERS.md#running-jobs-locally) work still, but only if you've added `ALLOW_LOCAL_GIT_REPOS` to your local `.env` - it could do with a note in the docs there as a reminder.

It'd probably also make sense to add it to the `dotenv-sample`, so it's set for local dev

evansd:
Thanks @rebkwok, all very good suggestions which I've now implemented.

rebkwok (approved):

PR #1447, opened 2026-06-09
Fix GitHub Actions vulnerabilities

alarthast:
Based on automated changes from running `zizmor --fix=all .`.

Part of https://github.com/opensafely-core/security/issues/46, https://github.com/opensafely-core/security/issues/47

Verified that all the complaints upon running `zizmor` are not listed in the tickets.

### DIFF
--- a/.github/workflows/build_publish_and_deploy.yaml
+++ b/.github/workflows/build_publish_and_deploy.yaml
@@ -30,7 +30,9 @@ jobs:
     name: Build, publish and deploy docker image
     steps:
     - name: Checkout
-      uses: actions/checkout@v6
+      uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
+      with:
+        persist-credentials: false
     - name: Set up Just
       uses: extractions/setup-just@53165ef7e734c5c07cb06b3c8e7b647c5aa16db3
     - name: Build image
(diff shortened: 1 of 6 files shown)


.github/workflows/build_publish_and_deploy.yaml
```diff
@@ -42,7 +44,7 @@ jobs:
 
     - name: Publish docker image
       run: |
-        echo ${{ secrets.GITHUB_TOKEN }} | docker login "$REGISTRY" -u ${{ github.actor }} --password-stdin
+        echo ${{ secrets.GITHUB_TOKEN }} | docker login "$REGISTRY" -u "${GITHUB_ACTOR}" --password-stdin
```
rebkwok, line 47: Does zizmor have any clarification about this one? I can't find anything that says that using the GITHUB_ACTOR env variable is more secure that using github.actor - just that they refer to the same thing. Unless you're using `pull_request_target`, which we're not - but maybe it's just a guard against that?

rebkwok (approved):

.github/workflows/build_publish_and_deploy.yaml
alarthast: > maybe it's just a guard against that?

Yes, if I am not mistaken, it guards against things like `pull_request_target` / `issue.title` etc. by just flagging all (?) template expressions in `run:` blocks, under their template injection audit rule: https://docs.zizmor.sh/audits/#template-injection

alarthast: Ahh, I think there are severity levels, this was probably flagged as a lower threat one (as opposed to `issue.title` etc.) but it auto-fixed it anyway.

> This is because zizmor considers all template expansions in code contexts to be code smells, and attempting to selectively permit them is more error-prone than forbidding them in a blanket fashion.

PR #1454, opened 2026-06-30
Use job's creation time to determine log directory partition

evansd:
The previous behaviour was to use the current time at whatever point the logs and metadata were being written to determine the partition directory to use. Where a job took multiple attempts (e.g. because previous attempts were interrupted for db maintenance or reboots), and where these attempts spanned a month boundary, the job could end up with multiple log directories.

Additionally, the code which retrieved the metadata for a given job had some unexpected behaviour: for jobs completed in the current month it would retrieve the metadata directly, but for older jobs it would do a search and then assert that only one result was found. This meant that the system worked fine until someone ran a job which depended on the results of another job which had run more than a month ago and had multiple attempts which spanned the month boundary. At this point it would fail the assertion and refuse to start the job, which is exactly what happened at the [beginning of this week](https://bennettoxford.slack.com/archives/C069YDR4NCA/p1782752075360859).

The absolute simplest fix here would be to remove the assertion, treat multiple matches as normal, and always use the lexically greatest match (i.e. the most recent). However, I think it would be preferable to maintain the invariant that each job has exactly one corresponding metadata file. This means that as well as deploying this fix we'll need to manually clean up the tiny handful of duplicated metadata files (see Slack [for details](https://bennettoxford.slack.com/archives/C069YDR4NCA/p1782841874248189?thread_ts=1782752075.360859&cid=C069YDR4NCA)) to avoid future problems.

That's slightly more work but not loads, and it avoids doubling down on a hack.

## How we got here and where we should go next

When we first started storing a persistent archive of job logs and metadata we decided to partition this by month in order to keep the directory sizes manageable. And while these files were – from the system's point of view – written once and then never read, it was easiest to do this by using the current date to determine the log partition.

However, during the work to refactor the original system into separate Agent and Controller components we ended up with a need for the Agent to retrieve historical job metadata. This was driven by a last minute change of requirements that meant we couldn't send output filenames out of the secure environment up to the Controller as originally planned. Because completing the Agent/Controller split was a blocker on beginning planned work on developing the RAP API a conscious decision was made to take on technical debt here. The intention was to pay this down during a third phase of Agent/Controller work, after the RAP API was up and running.

The correct approach here is to treat job/task metadata as a first class component of the Agent's state and design the storage model around that. It would make sense to do this alongside another long-discussed change which is to store output files from individual tasks separately, rather than in a workspace directory shared between tasks.

### DIFF
--- a/agent/executors/local.py
+++ b/agent/executors/local.py
@@ -73,10 +73,13 @@ def get_medium_privacy_workspace(workspace):
         return None
 
 
-def get_log_dir(job_id):
+def get_log_dir(job_definition):
+    created_at = datetime.datetime.fromtimestamp(
+        job_definition.created_at, datetime.UTC
+    )
     # Split log directory up by month to make things slightly more manageable
-    month_dir = datetime.date.today().strftime("%Y-%m")
-    return config.JOB_LOG_DIR / month_dir / container_name(job_id)
+    month_dir = created_at.strftime("%Y-%m")
+    return config.JOB_LOG_DIR / month_dir / container_name(job_definition.id)
 
 
 def read_job_metadata(job_id):
@@ -103,7 +106,7 @@ def read_job_task_metadata(job_definition):
 
 
 def write_job_metadata(job_definition, job_metadata):
-    metadata_path = get_log_dir(job_definition.id) / METADATA_FILE
+    metadata_path = get_log_dir(job_definition) / METADATA_FILE
     metadata_path.parent.mkdir(exist_ok=True, parents=True)
     metadata_path.write_text(json.dumps(job_metadata, indent=2))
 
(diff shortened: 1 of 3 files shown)


agent/executors/local.py
```diff
@@ -103,25 +106,24 @@ def read_job_task_metadata(job_definition):
...
 
 def job_metadata_path(job_id):
     """Return the expected path for the metadata for a job.
 
-    Due to writing to a directory path that includes the month at the time the
-    job was completed. We now need to be able to look up the metadata of a job
-    that may have completed in a previous month, so we use a glob to find it.
+    Due to writing to a directory path that includes the month at the time the job was
+    created we can't directly determine the path from the job ID and instead have to
+    search for it.
```
alarthast: I wonder if we should include a link to this PR here?

Someone reading this comment for the first time might wonder why we can't just feed the `created_at` timestamp into this function and avoid the search.

alarthast (dismissed):
This looks good to me. As I understand it, this change assumes that that `job_definition.created_at` timestamp would not be modified - I have convinced myself that it won't by doing a quick search in the codebase. In either case, if we somehow get duplicates, the assertion would catch it.

agent/executors/local.py
evansd: Oh yeah, I should expand on this in the comment; thanks. The problem is that we don't always have access to the `created_at` timestamp. Sometimes all we have is a list of job IDs. Specifically here:
https://github.com/opensafely-core/job-runner/blob/5bbedc6ab068390590ae0cb35132446f00c934b9/agent/executors/local.py#L404-L409

alarthast: Ahh, I see, thank you!

alarthast (approved):

